PR #2, opened 2026-06-04
Fix up + bump main: Kotlin 2.4.0, ksrpc 1.1.1, v0.2.3

monkopedia-coder:
Resurrect the long-stale `main` (Kotlin 2.3.0 / ksrpc 0.11.0) to a buildable, green state and cut patch release **0.2.3**.

## What changed
- **Versions:** Kotlin 2.3.0 → 2.4.0, ksrpc 0.11.0 → 1.1.1 (`gradle/libs.versions.toml`), `version` 0.2.2 → 0.2.3 (`gradle.properties`).
- **ksrpc 0.11 → 1.1 migration:** the host API is unchanged; the only real delta is that 1.1's compiler plugin enforces service-kind constraints — a `@KsService` that accepts/returns another `@KsService` must extend `RpcBidiService`. Applied to exactly the three flagged interfaces (`KrapperService`, `IndexedService`, `MappingService`); `ResolverService`/`RemoteLogger` stay plain `RpcService` (pure inputs).
- **Kotlin 2.4.0 fixes:** `KotlinWriter.onGenerate` gains defaults `size: LocalVar? = null, align: LocalVar? = size` (`align = size`, not `null` — the method path passes only `size` and goldens expect `alloc(size, size)`); `disableNativeCache` pinned to `2_4_0` (the clikt 5.1.0 duplicate-symbol returns at native link otherwise).
- **Golden refresh:** `testEmptyFile` / `testTemplateNaming` now emit a companion `align` getter + `_align_of` import + `alloc(size, align)`.
- **Pre-existing ktlint-cli 1.8.0 violations** fixed (cosmetic import/line/indent) so `:build` is green.

## Verification (JDK 17)
- `:krapper_gen:nativeTest` — **257 tests, 0 failures**
- `:krapper_gen:compileKotlinNative`, `:linkDebugExecutableNative`, `:krapper_gen:ktlintCheck` — green
- `:plugin:compileKotlin` / `:assemble` / `:ktlintCheck` / `:functionalTest` — green, but `:plugin:functionalTest` runs **only under JDK 21** (the plugin module hardcodes `targetCompatibility = VERSION_21`, pre-existing on `main`, unchanged here)

## Notes for review
- **Java 21 caveat:** unchanged from `main` — flagging since it affects whoever builds the plugin module. Lowering the plugin's Java target to 17 is deliberately out of scope for this patch.
[...]

PR #13, opened 2026-06-08
v2: usage-driven template binding generator + featuregen + clang self-bootstrap

monkopedia-coder:
One squashed PR bringing the **v2 line** (the whole template/featuregen/clang-bootstrap campaign) onto modernized `main`.

## What this is
v2 is a near-total rework of how kplusplus generates bindings: a **usage-driven, two-stage Gradle plugin flow** instead of the old one-shot reducer.

- **`com.monkopedia.kplusplus.compiler`** Gradle plugin: `kplusplusSync` generates bindings from a C++ header; the FIR compiler plugin refines templates at compile time. Scoped-import allowlist (`only`/`onlyFile`) + `referencePolicy`, and **`klinker`** for C++ linking.
- **featuregen**: an end-to-end harness running the real flow over a C++ surface (std containers, templates, smart pointers, ranges, enums, callbacks, inheritance, operators). **183/0.**
- **User templates** (`Box<Int>` → concrete `Box__Int`), `iterator_range` → `std::vector` materialization, enum → underlying-int reduction, capturing-lambda callbacks, multiple inheritance via `.asBase()`, configurable `rootPackage`.
- **Clang self-bootstrap**: kplusplus binds Clang's own C++ AST and a Kotlin program walks a real AST on those bindings — proven end-to-end. The stage1 consumer (`clangwalk`) lives on the `clang-selfhost` branch; see `docs/clang-bootstrap.md`.

## Reconciliation with main
main's #2/#3 (Kotlin 2.4.0, ksrpc 1.1.1, v0.2.3, auto-publish/vanniktech) reconciled with the v2 tree. The version catalog auto-merged; the only code conflicts were three incidental overlaps, all resolved to the v2 side (krapper_gen `sharedLib` block, `onGenerate` signature, `WrappedType` imports).

## Codegen-correctness fixes folded in (each independently verified)
These fix a regression cluster that had **silently broken featuregen since June 6** (commits that claimed green but never re-ran the featuregen gate):
- **std::string surface blow-up** — policy-gate the forcing pass-3 eviction (was duplicating std::string 6×, dropping `<char>`, ballooning the generated .cc 1.5MB→14MB).
- **dropped const-ref-default ctors** — gate the false-default check on reference const-ness (a blanket `isReference → required` had dropped `basic_string`'s const-char*/char/initializer_list ctors).
- **enum `fromValue`** out-of-range no longer throws; **keyword method names** (`in`/`is`) back-tick-escaped; **`op==`** name-mangle typo.

## Verification on this reconciled tree
- featuregen **183/0**, krapper_gen **295/0**, ktlint green
- clang stage1 sync + AST walk run on top of these fixes: walks 9 real libclang-cpp decls

## Notes for review
- **Version left at 0.2.3** — the `0.4.0` bump is a separate, deliberate step (not in this PR).
- Publishing is gated behind a GitHub **Release** only (`publish.yaml` `on: release`), so merging this PR does **not** publish anything.
- Full 184-commit history is preserved on the `clang-base` branch.


monkopedia-reviewer (commented):
## Reviewer bot verdict — @Monkopedia please merge

**Tier 3 escalation** (breaking_change — near-total architecture rework; this one needs your merge).

### What I reviewed and found sound

**Three folded-in codegen-correctness fixes** (each has dedicated test coverage, all green in featuregen 183/0 / krapper_gen 295/0):

- **`typeCarriesFalseDefault` gate** (`WrappedArgument`): Non-const lvalue reference → required (not omittable-default); const reference falls through and judges the referent. Array types and non-type template args (`Mask<4>`) also gate correctly. The `isNonTypeTemplateArg` extension (numeric/bool/char literal detection) is correct. This fixes the dropped `basic_string` const-char*/char/initializer_list ctors.

- **Policy-gated forcing eviction** (`Resolver.resolveForcing`): Under `IGNORE_MISSING`, evicts the union of main-parse keys + forcing keys (guards against canonical/include-order key drift on large headers). Under `INCLUDE_MISSING`, evicts only forcing keys (leaves sibling-bound types like `std::string` in place to avoid re-pulling them as degraded duplicates). Logic matches the stated rationale; comments are thorough.

- **`fromValue` fallback** (`KotlinWriter`): `entries.firstOrNull { it.value == v } ?: entries.first()` — correct; prevents `NoSuchElementException` on out-of-range values. Test: `fromValue_out_of_range_does_not_crash`.

- **Keyword escaping**: `KOTLIN_HARD_KEYWORDS` set matches the Kotlin spec; `escapeKotlinKeyword()` applied at method names, param names, and local vars. Test: `keyword_named_methods_bind_and_call` (`in`/`is`).

- **`op==` → `equals` + `hashCode`**: Generates a real Kotlin `equals(other: Any?)` override; `hashCode` folds public fields. Test expectations updated in `FullKotlinTests`.

- **`holderMapping` vs `rawMapping`** (ARG_CAST const-stripping): Strips top-level `const` from the Holder buffer's type for placement-new (avoids ill-formed `new (buf) const T(...)`). Correct; the Kotlin-facing type (`rawResolved`) is untouched.

- **`RuntimeException` widening** (`WrappedType.createForType`): Gated by `throwOnError` — strict mode still throws; loose mode drops unmodelable types gracefully. Sound pattern.

- **Merge conflict resolutions**: All three stated incidental overlaps (krapper_gen `sharedLib` block → v2 wins; `onGenerate` signature → v2 wins; `WrappedType` imports → v2 wins) are correct; no leftover conflict markers found.

- **Verification claim supported**: 183/0 featuregen, 295/0 krapper_gen, ktlint green, clang stage1 AST walk — consistent with the diff showing comprehensive test coverage across all new features. No PR CI exists on this repo (`publish.yaml` fires on Release only); the local verification is the available gate.

**Minor observation (comment, not a blocker):** `compiler/build.gradle.kts` hardcodes `version = "0.2.2"` while the main project is at `0.2.3`. The compiler subproject is an `includeBuild` and not published through the existing workflow, so this is harmless now — but worth aligning when the `0.4.0` bump happens.

### Why this needs your hand on the merge

This PR replaces the entire binding-generator architecture (218 files, +15180/-710, 184 squashed commits). The published `:plugin` artifact's API surface is replaced. Publishing is gated behind a GitHub Release (nothing auto-publishes on merge), but this is still the single biggest irreversible state change in the repo. A bot auto-merge is not appropriate for something of this scale — you should merge this yourself so you can confirm the branch and timing are what you intend.

**Ready to merge.** Once you approve and merge, the separate `0.4.0` version bump is the next step.

monkopedia-reviewer (approved):
Tier-1 per repo policy: all kplusplus PRs are tier-1 until the owner says otherwise (owner directive 2026-06-08). The diff was already vetted to merge standard in the prior review — three codegen fixes sound + tested, conflict resolutions correct, local verification supported. Auto-merging.

PR #14, opened 2026-06-08
Soundness: drop-ledger makes skip-not-crash binding drops visible (#6)

### LINKED ISSUE #6: [clang-campaign] Soundness: make skip-not-crash binding drops visible (drop-ledger + coverage)
**Campaign productionization (Clang self-bootstrap).** Found by a 3-agent audit of the stage1 self-bootstrap (the AST walk runs end-to-end; this hardens it toward release). Context: `docs/clang-bootstrap.md`; work lives on `clang-base` (generator) + `clang-selfhost` (consumer `clangwalk/`).

**Soundness — make skip-not-crash drops VISIBLE (highest-value item).**
The breadth strategy drops unmodelable methods/fields/classes with only a `Log.w` — never counted, never diffed against the requested allowlist. A consumer cannot tell "C++ doesn't have it" from "the generator dropped it." Drop sites: `WrappedElement.kt:211,405`, `WrappedField.kt:78`, `WrappedMethod.kt:575`, `Resolver.kt:140,235,250,450`, `WrappedClass.kt dropConstOverloadDuplicates`, plus `WrappedKotlinType.kt` opaque-leaf degradations (213/235/268/281).

**Do:** a drop-ledger — collect `{symbol, reason}` across all passes, emit a "requested X / bound Y / dropped Z (+list)" report, and an allowlist-coverage check (optionally a `--fail-on-drop` / coverage assertion). Acceptance: a scoped run reports exactly what it dropped and why.

monkopedia-coder:
## What

The breadth strategy dropped unmodelable methods/fields/classes with only a `Log.w` / `println` — never counted, never diffed against the requested allowlist. A consumer could not tell "C++ doesn't have it" from "the generator dropped it." This adds a **drop-ledger** that makes every skip-not-crash drop visible.

## Design

A process-scoped `DropLedger` collector (modeled on the sibling `Log` object + the `rootPackageOverride` global — the generator runs one request at a time under `runBlocking`, so a shared collector is the cohesive carrier without threading a ledger handle through every resolve signature). Each known drop site funnels `{symbol, reason, phase}`:

- **RESOLVE**: `ResolveContext.notifyFailed` — the central choke point every model element returns through when it fails to resolve, so recording once covers the method/field/class drop sites uniformly; plus the top-level filtered-class / forcing-struct resolve failures in `resolveAll` / `resolveForcing`.
- **PARSE**: the cursor-level skip-not-crash drops in `WrappedElement` (nested-in-class-template class, unmodelable cursor) and the opaque-leaf template-spelling degradations in `WrappedKotlinType`.
- **DEDUP**: the const-overload de-duplication in `WrappedClass.dropConstOverloadDuplicates`.

`IndexedServiceImpl` tracks the requested set (`--only` allowlist entries + each `--instantiate` target), resets the ledger per run, and emits a `requested X / bound Y / dropped Z` report (with the dropped list + reasons, and a requested-but-neither-bound-nor-dropped coverage line) at the end of `writeTo`.

Strict mode is **opt-in** via `--fail-on-drop` (`KrapperConfig.failOnDrop`, default `false`): drops are always logged + ledgered, but only this flag turns a drop into a non-zero exit — so existing green runs are unchanged.

## Tests

`SkipNotCrashTest`:
- `l1` — a deliberately-unbindable required-param method is recorded in the ledger with a reason, while its bound sibling is not.
- `l2` — a fully-bindable header records zero drops (no false positives).
- `l3` — `--fail-on-drop` fails the run iff something was dropped (and the same header passes leniently by default).

## Verification

- `:krapper_gen:nativeTest` — **298/0** (295 baseline + 3 new ledger tests)
- `:featuregen:nativeTest` — **183/0**

## Deferred / known limitation

The requested-vs-bound coverage line is best-effort: an `--instantiate` request string (`std::pair<int, int>`) and a bound `ResolvedClass.type` spelling can differ, so an instantiation may show as "uncovered" even when bound. It is informational only (never a gate), and the dropped list — the issue's actual acceptance criterion ("reports exactly what it dropped and why") — is precise.

Closes #6

### DIFF
--- a/krapper_gen/src/commonMain/kotlin/KrapperConfig.kt
+++ b/krapper_gen/src/commonMain/kotlin/KrapperConfig.kt
@@ -46,5 +46,10 @@ data class KrapperConfig(
     // behavior: top-level user types -> package `root`, C++ namespaces -> their
     // bare path (`std`, `geo`). When set (e.g. "com.acme.app"), every generated
     // binding's package becomes <rootPackage> + the C++ namespace path.
-    val rootPackage: String? = null
+    val rootPackage: String? = null,
+    // When true, a run that dropped ANY unmodelable symbol (skip-not-crash) fails
+    // after emitting the drop-ledger report, instead of completing leniently. Opt-in
+    // (default false) so existing green runs keep succeeding: drops are always logged
+    // and ledgered, but only this flag turns a drop into a non-zero exit.
+    val failOnDrop: Boolean = false
 )
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
@@ -141,6 +141,12 @@ class KrapperGen : CliktCommand() {
             "class name per line; blank lines and lines starting with # are ignored). " +
             "Merged with any --only entries."
     )
+    val failOnDrop by option(
+        "--fail-on-drop",
+        help = "Fail the run (non-zero exit) if the generator dropped any unmodelable " +
+            "symbol (skip-not-crash). Default: lenient — drops are logged + reported in " +
+            "the drop ledger, but the run still succeeds."
+    ).flag()
     val fixupFile by option(
         "--fixup-file",
         help = "Path to a JSON file containing a list of Fixup directives (see Fixup.kt). " +
@@ -167,7 +173,8 @@ class KrapperGen : CliktCommand() {
                     referencePolicy = referencePolicy,
                     debug = debug,
                     cppStandard = cppStandard,
-                    rootPackage = rootPackage
+                    rootPackage = rootPackage,
+                    failOnDrop = failOnDrop
                 )
             )
             val indexService = service.index(IndexRequest(header, library))
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/DropLedger.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/DropLedger.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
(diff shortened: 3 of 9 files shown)


monkopedia-reviewer (approved):
Reviewed as monkopedia-reviewer. Tier 1 (kplusplus override — all PRs auto-merge).

**Observability-only confirmed.** Every `DropLedger.record(...)` call is inserted before an already-existing skip path (`return null`, `getOrElse`, `if (resolve == null)`) with no change to the surrounding conditionals or return values. The drop conditions are exactly the same as before; recording is a pure side-effect addition.

**Ledger reset confirmed.** `DropLedger.reset()` is called in `IndexedServiceImpl.init` before any resolve work, which fires for every fresh `IndexedServiceImpl` instantiation — covering both the service path and the test helper. `DropLedger` is a Kotlin `object` with a `mutableListOf`; `reset()` clears via `records.clear()`. No cross-run leakage.

**Coverage deferred scope-cut is acceptable and documented.** The requested-vs-bound spelling mismatch for `--instantiate` strings is called out explicitly in the PR body and in the `report()` KDoc as informational-only (never a gate). The issue #6 acceptance criterion — "reports exactly what it dropped and why" — is satisfied by the precise dropped list; the coverage line is best-effort by design.

**Tests (L1/L2/L3) are meaningful.** L1 asserts the unbindable method appears in the ledger with a non-blank reason and the bindable sibling does not. L2 asserts zero drops on a fully-bindable header (no false positives). L3 locks the opt-in strict-mode contract end-to-end. All three are exercising the actual ledger recording path, not just that the run survives.

**Quality.** `DropLedger` is idiomatically modeled on the sibling `Log` object. Naming (`DropRecord`, `DropPhase`, `record`/`reset`/`report`/`hasDrops`) is clear and consistent. Comment density and structure match the surrounding code. No CI pipeline exists (`publish.yaml` is `on: release` only); test counts (298/0, 183/0) reported by coder are consistent with baseline +3 new tests.

No correctness issues, no quality issues.

PR #19, opened 2026-06-08
Diagnostics: skip the bad symbol into the ledger, don't abort the whole run (#9)

### LINKED ISSUE #9: [clang-campaign] Diagnostics policy: skip the symbol, don't abort the whole run on any error: diagnostic
**Campaign diagnostics policy (Clang self-bootstrap).** `Parsing.kt` (~`:594`, `parseHeader`) aborts the WHOLE run on ANY `error:` diagnostic. Fine for curated clean headers, but the stated goal is pointing at real libraries — one bad header takes the entire import down. The autonomy contract already says "replicate abort-on-error via `ASTUnit::getDiagnostics`" but per-symbol-skippable is what a broad import needs.

**Do:** make a parse/resolve diagnostic skip the affected symbol (drop+log, feeding the drop-ledger from the soundness ticket) instead of aborting the run; keep an opt-in strict mode. Acceptance: a header with one un-bindable symbol still binds the rest.

monkopedia-coder:
## What

`parseHeader` (Parsing.kt) previously threw on **any** `error:` diagnostic across the whole translation unit, so a single un-bindable declaration in a real-world header took the entire import down. This makes the diagnostic policy **skip the affected symbol** instead, feeding the existing drop-ledger from #6.

### Default (lenient) policy — skip-not-crash
An `error:` diagnostic attributable to a single top-level declaration:
- resolves the diagnostic's source location to a cursor (`clang_getDiagnosticLocation` → `clang_getCursor`), walks up semantic parents to the outermost named non-namespace decl;
- records a `DropLedger` PARSE drop `{symbol, reason=diagnostic text, phase=PARSE}` (deduped by USR — one bad decl can raise several diagnostics);
- excises that declaration by USR in `WrappedElement.map` (returning `null` there drops the decl **and** its members via the existing `map(parentCursor) ?: return@forEachRecursive` guard), so it never reaches resolution;
- binding continues over the rest of the header.

Because the drop flows through the existing ledger, the end-of-run report and the `--fail-on-drop` strict gate from #6 cover diagnostic-skips for free — no parallel reporting path.

### The attributable-vs-fatal boundary
- **attributable error** → drop that symbol, continue (the new default).
- **fatal severity** (`CXDiagnostic_Fatal`, e.g. a missing `#include`) → always abort; the recovered AST can't be trusted past it.
- **unattributable error** (location resolves to the TU root / no enclosing declaration) → always abort; there's no single symbol to surgically drop. Namespaces are explicitly not drop targets (a container, not a bindable symbol — dropping one would take every sibling with it).

### Strict-mode flag choice
Added a dedicated `--strict-diagnostics` (`KrapperConfig.strictDiagnostics`, default off) rather than folding into `--fail-on-drop`. They're orthogonal: `--fail-on-drop` is a *post-run gate* ("did anything get dropped?"), while parse-strictness is an *in-run policy* ("abort the instant a header errors, before any binding"). A curated-clean-header run wants strict parsing but `--strict-diagnostics` is the cleaner knob; `--fail-on-drop` still independently turns the resulting drop into a non-zero exit.

## Tests (ParseTest, parse+resolve level, decoupled from the wrapper compile)
- `testParseErrorDropsBadSymbolAndBindsTheRest` — a header with one un-bindable struct (`Undeclared` member) ledgers the bad struct as a PARSE drop with the diagnostic text, while the clean sibling class still resolves (acceptance criterion).
- `testStrictDiagnosticsAbortsOnAnyError` — the same header aborts under `--strict-diagnostics`.
- `testUnattributableErrorAbortsEvenWhenLenient` — a missing include aborts even in lenient mode (the fatal/TU-level boundary).

## Verification
- `:krapper_gen:nativeTest` — **301/0** (298 baseline + 3 new)
- `:featuregen:kplusplusSync` + `:featuregen:nativeTest` — **183/0**

All four gates green in a single combined run.

Closes #9

### DIFF
--- a/krapper_gen/src/commonMain/kotlin/KrapperConfig.kt
+++ b/krapper_gen/src/commonMain/kotlin/KrapperConfig.kt
@@ -51,5 +51,14 @@ data class KrapperConfig(
     // after emitting the drop-ledger report, instead of completing leniently. Opt-in
     // (default false) so existing green runs keep succeeding: drops are always logged
     // and ledgered, but only this flag turns a drop into a non-zero exit.
-    val failOnDrop: Boolean = false
+    val failOnDrop: Boolean = false,
+    // When true, ANY `error:`-severity parse diagnostic aborts the whole run (the
+    // historical behavior — appropriate for curated, known-clean headers). Opt-in
+    // (default false): by default an error diagnostic that is attributable to a single
+    // declaration drops THAT symbol into the drop ledger and binding continues over the
+    // rest of the header, so one un-bindable symbol in a real-world library no longer
+    // takes the entire import down. Translation-unit-level / fatal diagnostics (a missing
+    // include, "too many errors", an error with no attributable cursor) abort regardless,
+    // since the recovered AST can't be trusted past them.
+    val strictDiagnostics: Boolean = false
 )
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
@@ -147,6 +147,15 @@ class KrapperGen : CliktCommand() {
             "symbol (skip-not-crash). Default: lenient — drops are logged + reported in " +
             "the drop ledger, but the run still succeeds."
     ).flag()
+    val strictDiagnostics by option(
+        "--strict-diagnostics",
+        help = "Abort the whole run on ANY parse `error:` diagnostic (the historical " +
+            "behavior, for curated clean headers). Default: lenient — an error attributable " +
+            "to a single declaration drops THAT symbol into the drop ledger and binding " +
+            "continues, so one bad symbol in a real library doesn't take the import down. " +
+            "Fatal / translation-unit-level errors (missing include, unattributable) abort " +
+            "regardless."
+    ).flag()
     val fixupFile by option(
         "--fixup-file",
         help = "Path to a JSON file containing a list of Fixup directives (see Fixup.kt). " +
@@ -174,7 +183,8 @@ class KrapperGen : CliktCommand() {
                     debug = debug,
                     cppStandard = cppStandard,
                     rootPackage = rootPackage,
-                    failOnDrop = failOnDrop
+                    failOnDrop = failOnDrop,
+                    strictDiagnostics = strictDiagnostics
                 )
             )
             val indexService = service.index(IndexRequest(header, library))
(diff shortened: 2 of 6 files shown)


monkopedia-reviewer (approved):
**Tier 1 — approve and merge (kplusplus 2026-06-08 override)**

### Attributable-vs-fatal boundary (scrutiny point 1)
Confirmed correct. `handleDiagnostics` maps three disjoint cases:
- `strictDiagnostics=true` → always abort (first branch, before any attribution check).
- Any error where `isAttributable` is false (fatal severity, or symbol/usr is null = TU-root location) → `mustAbort=true` → throw with `"fatal / translation-unit-level diagnostic(s)"` message.
- All remaining errors are attributable (non-fatal, have a named non-namespace top-level decl) → drop each into the ledger by USR, continue.

Namespaces are explicitly excluded in `topLevelDecl` (`kind != CXCursor_Namespace`) so a namespace is never a drop target, preventing the sibling-loss hazard. Sound.

### USR-based excision (scrutiny point 2)
`topLevelDecl` walks semantic parents upward, writing `named` at each non-namespace named declaration it passes through, and returns the last value written before hitting the TU root. For a bad field member of a struct, the walk assigns `field`, then overwrites with `struct`, exits at namespace/TU — drop target is the struct. Correct precision: just the right symbol, not more. USR dedup in `handleDiagnostics` prevents double-recording of a multi-diagnostic single bad decl. The excision in `WrappedElement.map` returns `null` for the matched USR; the existing `map(parentCursor) ?: return@forEachRecursive` guard in `mapAll` propagates the null transitively through members. Clean.

`droppedUsrs` state safety: `setDroppedUsrs` is called immediately before each `mapAll` call, within the private per-file `parseHeader`. Since `List.map` is sequential and the generator is `runBlocking`-single-threaded, each TU correctly scopes its own dropped-USR set. No cross-TU contamination.

### Flag separation (scrutiny point 3)
`--strict-diagnostics` (in-run policy: abort vs skip) is orthogonal to `--fail-on-drop` (post-run gate on the ledger). Both are opt-in, default false. A user wanting strict-parse but no exit-on-drop, or lenient-parse but exit-on-any-drop, can mix them independently. Sound separation.

### Tests (scrutiny point 4)
All three cases cover the acceptance criteria:
1. `testParseErrorDropsBadSymbolAndBindsTheRest` — the primary criterion: `Bad` is a PARSE drop with `"error:"` in reason, `Good` is in resolved names, `Bad` is not. Not tautological: the resolver actually runs.
2. `testStrictDiagnosticsAbortsOnAnyError` — same header that binds-the-rest in lenient mode now throws with `"strict diagnostics"` in the message. Exercises the policy boundary.
3. `testUnattributableErrorAbortsEvenWhenLenient` — missing `#include` in lenient mode aborts with `"fatal / translation-unit-level"`. Draws the unattributable/fatal boundary. The comment in the test correctly observes this is a fatal diagnostic (not merely unattributable), but both map to the same abort path, and the test message check is consistent with the implementation.

All tests meaningful, non-tautological. Reported 301/0 + 183/0 + ktlint green.

monkopedia-reviewer (approved):
Tier 1 auto-approve: boundary/USR/flags/tests all sound. See detailed review comment.

Issue #1, opened 2024-08-01
Windows Support


Cdm2883:
The reason seems to be that krapper_gen uses ksrpc to communicate with libclang and ksrpc uses ktor-server to communicate, but the native implementation of ktor-server doesn't support windows.

<details>

<summary>This is an error:</summary>

```
A problem occurred configuring project ':plugin'.
> Could not resolve all files for configuration ':plugin:classpath'.
   > Could not find kplusplus:krapper_gen:0.1.0.
     Searched in the following locations:
       - https://repo.maven.apache.org/maven2/kplusplus/krapper_gen/0.1.0/krapper_gen-0.1.0.pom
       - https://plugins.gradle.org/m2/kplusplus/krapper_gen/0.1.0/krapper_gen-0.1.0.pom
     Required by:
         project :plugin > com.monkopedia.kplusplus.plugin:com.monkopedia.kplusplus.plugin.gradle.plugin:0.1.0 > com.monkopedia.kplusplus:plugin:0.1.0

Possible solution:
 - Declare repository providing the artifact, see the documentation at https://docs.gradle.org/current/userguide/declaring_repositories.html
```

</details>

<details>

<summary>This is the result of a composite build using git's submodule and then gradle:</summary>

```
A problem occurred configuring project ':plugin'.
> Could not determine the dependencies of task ':kplusplus:krapper_gen:linkDebugExecutableNative'.
   > Could not resolve all task dependencies for configuration ':kplusplus:krapper_gen:nativeCompileKlibraries'.
      > Could not resolve com.monkopedia:ksrpc:0.6.0.
        Required by:
            project :kplusplus:krapper_gen
         > No matching variant of com.monkopedia:ksrpc:0.6.0 was found. The consumer was configured to find a library for use during 'kotlin-api', preferably optimized for non-jvm, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'native', attribute 'org.jetbrains.kotlin.native.target' with value 'mingw_x64' but:
             - Variant 'jsApiElements-published' capability com.monkopedia:ksrpc:0.6.0 declares a library for use during 'kotlin-api':
                 - Incompatible because this component declares a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'js' and the consumer needed a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'native'
                 - Other compatible attributes:
                     - Doesn't say anything about its target Java environment (preferred optimized for non-jvm)
[...]

PR #20, opened 2026-06-08
Characterize resolveForcing: pin the eviction-policy / pass-3 / no-explosion behavior (#18)

### LINKED ISSUE #18: [test-hardening] Characterize resolveForcing (multi-pass forcing) so it can be safely reduced — pins the recent regression class
**Phase A of "test-harden then reduce" for the most critical, most-regressed generator path.** `Resolver.kt resolveForcing` (the 3-pass shared-tracker forcing with policy-gated eviction, ~160 LoC) is load-bearing and was the source of BOTH June regressions: the std::string eviction blow-up (over-wide pass-3 eviction degraded sibling-bound `std::string` into duplicated, uncompilable bindings) and — adjacent — the const-ref-default ctor drop. Both slipped because the ONLY net was featuregen integration, which wasn't re-run. The reduction analysis (#17 item 10) correctly flagged this code as risky to touch; the right response is not to avoid it but to **pin its behavior with direct tests**, then reduce it.

**Do — add characterization tests (no production change in this PR) that lock the current correct behavior**, especially the cases that have regressed:
- **Eviction-policy gate:** under `INCLUDE_MISSING` pass-3 evicts ONLY the forcing keys (narrow) — a sibling-bound std type (e.g. `std::string` bound by a separate `std::vector<string>` request) is NOT re-pulled/degraded/duplicated; under `IGNORE_MISSING` it evicts the union (`alreadyBoundKeys` ∪ forcing keys) so a drifted-key bound class still recovers. (This is the exact axis that blew up — a test here would have failed loudly.)
- **Pass-3 range-method recovery:** a container-returning method dropped in pass-1 (return type was the then-unbound container) is recovered after the container binds (the featuregen `RangeHolder::items()` analog / clang `DeclContext::decls()`).
- **No forcing explosion / scope creep:** forcing a specialization binds only the requested specialization + already-bound classes, not the transitive world (the ~1269-class explosion stays killed).
- **Drop visibility:** forcing-time drops land in the DropLedger (now that #6 merged).

**Acceptance:** a krapper_gen (and/or featuregen) test suite such that re-introducing either June regression (widen the INCLUDE_MISSING eviction back to the union; or the old blanket-`isReference` false-default) makes a test FAIL at unit/integration level, not just a silent featuregen-sync abort. This suite then gates + de-risks the Phase-B reduction of `resolveForcing` (#17 item 10) and the operator-table / interface-machinery items (#17 items 7 + honorable-mention), which get the same treatment.

Ref: the regression write-up is in the project notes (featuregen regression cluster, 2026-06-08).

monkopedia-coder:
Test-hardening for `Resolver.kt resolveForcing` (the 3-pass shared-tracker forcing resolve) — the most critical, most-regressed generator path. Adds DIRECT tests that pin the current correct behavior so re-introducing either June regression (over-wide INCLUDE_MISSING eviction; const-ref-default drop) FAILS loudly, and so the code can later be reduced against this net. **No production change** (Resolver.kt diff is empty; only test files + one comment).

## What I pinned, and at what level

**krapper_gen unit — `ForcingCharacterizationTest`** (one cached INCLUDE_MISSING forcing run that mirrors `IndexedServiceImpl.requestInstantiation`):
- **Eviction-policy gate (resolved-set invariant):** a sibling-bound `std::string` survives the forcing run as exactly one `basic_string<char>` binding and is the same entry from the main resolve (in `alreadyBoundKeys`) — not lost/split/bare-degraded.
- **No forcing explosion / scope creep:** the resolved set stays bounded (the ~1269-class blow-up stays killed).
- **Drop visibility:** forcing-time drops land in the `DropLedger` (snapshotted around the forcing resolve).

**featuregen integration — `RfForcingEvictionTest` (the LOUD bite for behavior #1):** `StringFeature` binds `std::string` siblingly while a forced `std::vector<std::string>` shares the same `Basic_string__Char` element. Re-widening the INCLUDE eviction to the union degrades that shared binding and **aborts `:featuregen:kplusplusSync` (krapper_gen exit 134)** — so the whole module fails to build. When green, it round-trips the const-char* ctor + `c_str()` on the single intact `<char>` binding, and flows a `StringFeature`-produced string through the forced vector.

**Behavior #2 (pass-3 range-method recovery):** strengthened + referenced in the featuregen `RgRangeTest` (`RangeHolder::items()` only exists because pass-3 re-resolves the bound class against the now-forced container). The IGNORE_MISSING union-evict axis stays pinned by the existing `ReturnShapeTest.forcedContainerBindsUnderIgnoreMissing`.

## "Does it bite" sanity check (critical)
Temporarily widened the INCLUDE_MISSING eviction to the `alreadyBoundKeys` union (June's exact blow-up) and re-ran:
- **featuregen gate FAILED** as intended: `:featuregen:kplusplusSync` aborted with `krapper_gen failed: exit 134` on the degraded `std::vector<std::__cxx11::basic_string<char>>` cast. ✅ This is the loud, named bite for the eviction-policy gate.
- The krapper_gen unit test did NOT bite — at that thin unit scale the resolved-class map dedups by key, so the degraded re-pull shows in the generated `.cc`, not the map. I therefore **did not overclaim**: the unit test's comment + assertions pin only the resolved-set invariant, and explicitly point at `RfForcingEvictionTest` as the loud bite. Reverted the widening after verifying.

## Notable constraint
Repeated `<vector>`-instantiating forcing TUs in one JVM degrade libclang's `std::vector`/`iterator_range` canonicalization for *later* tests in the same process (it silently broke `ReturnShapeTest.iteratorRangeReturnMaterializesVector`). I kept `ForcingCharacterizationTest` to a single forcing run (lazily cached, shared across its 3 assertions) to stay within that budget; the IGNORE_MISSING forcing axis is covered by sibling tests rather than a second forcing run here.

## Verification (sync+test in one invocation)
`:krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` — all green:
- krapper_gen **301/0** (298 base + 3 new)
- featuregen **186/0** (183 base + 3 new)
- ktlint green

Closes #18.

### DIFF
--- a/featuregen/src/nativeTest/kotlin/RfForcingEvictionTest.kt
+++ b/featuregen/src/nativeTest/kotlin/RfForcingEvictionTest.kt
@@ -1,0 +1,78 @@
+import com.monkopedia.kplusplus.cppVector
+import kotlinx.cinterop.memScoped
+import root.StringFeature
+import std.__cxx11.Basic_string__Char
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+// RF-forcing-eviction (issue #18, the eviction-policy gate that BLEW UP in June):
+// resolveForcing's pass-3 eviction must, under INCLUDE_MISSING, evict ONLY the forcing
+// keys — NOT the `alreadyBoundKeys` union. featuregen binds `std::string`
+// (`Basic_string__Char`) two ways at once: as a SIBLING type referenced by `StringFeature`
+// (StringFeature::take/echo/produce), AND as the ELEMENT of a forced
+// `std::vector<std::string>` instantiation (auto-discovered SYNC_REQUIRED from the
+// cppVector<Basic_string__Char>() call sites below). That is exactly the scenario that
+// degraded in June: widening the INCLUDE_MISSING eviction back to the union evicts the
+// sibling-bound `std::string` from the forcing tracker, so pass-3 re-pulls it as a fresh
+// INCLUDE_MISSING reference — emitting a DUPLICATE, degraded `basic_string` whose param
+// casts lose their `<char>` template arg, which is uncompilable C++ and aborts
+// `kplusplusSync` (krapper_gen exit 134).
+//
+// REGRESSION GUARD: re-widening the INCLUDE_MISSING eviction to the union makes
+// `:featuregen:kplusplusSync` FAIL (the degraded `Basic_string__Char` cast won't compile),
+// so this whole module fails to generate/build. When it DOES build, this test asserts the
+// `Basic_string__Char` surface is intact — its const-char* constructor + c_str() round-trip
+// AND the vector<string> element read-back both work on the single, correctly-`<char>`-
+// templated binding. (Verified to bite: with the union-evict re-introduced, sync aborts at
+// exit 134.) The generated static methods are MemScope-extension members on the companion,
+// so feature calls are `ms.method(...)` inside `with(StringFeature)`.
+class RfForcingEvictionTest {
+
+    // The SIBLING std::string surface (StringFeature) must round-trip: take() stores a
+    // std::string built from a const char*, echo() returns it, produce() makes one. If the
+    // sibling binding were degraded by the over-wide eviction, these would not compile.
+    @Test fun sibling_stringFeature_roundTrips() = memScoped {
+        val ms = this
+        with(StringFeature) {
+            with(Basic_string__Char) {
+                ms.reset()
+                ms.take(ms.Basic_string__Char__const_char_P("héllo"))
+                assertEquals(6uL, ms.lastSize()) // "héllo" is 6 UTF-8 bytes
+                assertEquals("produced!", ms.produce().c_str())
+                assertEquals("round", ms.echo(ms.Basic_string__Char__const_char_P("round")).c_str())
+            }
+        }
+    }
+
+    // The FORCED std::vector<std::string> uses the SAME single Basic_string__Char element
+    // binding. Pushing const-char*-constructed strings and reading them back via c_str()
+    // proves the element binding kept its `<char>` template arg (a degraded duplicate would
+    // mis-decode or fail to compile the element cast).
+    @Test fun forced_vectorOfString_sharesIntactStringBinding() = memScoped {
+        val ms = this
+        val v = cppVector<Basic_string__Char>()
+        v.push_back(with(Basic_string__Char) { ms.Basic_string__Char__const_char_P("alpha") })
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
**Approved — characterization tests are genuine, honestly scoped, and the featuregen gate bites.**

### Correctness

**`RfForcingEvictionTest` (featuregen) — confirms bite:** The test exercises exactly the June regression scenario: `StringFeature` binds `Basic_string__Char` as a sibling while `cppVector<Basic_string__Char>()` drives auto-discovery of a forced `std::vector<std::string>`. Re-widening the INCLUDE_MISSING eviction to the `alreadyBoundKeys` union degrades that shared binding into a duplicate with lost `<char>` template arg in generated C++ — uncompilable, so `:featuregen:kplusplusSync` aborts (exit 134) before the tests even run. The three test methods (sibling round-trip, forced-vector element integrity, cross-flow) then prove the single intact binding works end-to-end. Not a tautology — the gate is at code-generation, not at the Kotlin assertion level.

**`ForcingCharacterizationTest` (krapper unit) — honestly scoped:** The test explicitly documents that the eviction-widening regression surfaces in the generated `.cc` (where the duplicate degraded param cast is emitted), not in the resolved-class map (which dedups by key at unit scale). It asserts only what it can see: the resolved-set invariant (sibling `basic_string<char>` survives as exactly one entry still in `alreadyBoundKeys`), no-explosion (< 50 classes vs the ~1269-class blow-up), and drop-ledger visibility. The comment cross-references `RfForcingEvictionTest` as the loud bite. Accurately-scoped-and-says-so is the right outcome here.

**`RgRangeTest` — real strengthening:** The added comment converts the existing test into an explicit characterization anchor for pass-3 range-method recovery (`items()` exists only because pass-3 re-resolves bound classes against the now-forced container). If that recovery regressed, the class wouldn't compile. The comment is accurate and the existing assertion is unchanged.

**Single forcing run constraint respected:** `includeResult` is `by lazy` — one cached run shared across all three assertions. The comment documents why a second forcing TU would silently break `ReturnShapeTest.iteratorRangeReturnMaterializesVector`. Constraint is correctly implemented and explained.

### Quality

Assertion messages include the actual data (specs, keys) so failures are self-diagnosing. The precondition check on `boundStringKeys.isNotEmpty()` catches setup failures separately from the invariant under test. The `< 50` explosion bound is justified (generous but biting). Code style matches the surrounding test suite (memScoped, runBlocking, companion lazy init — same pattern as ReturnShapeTest). Names are clear throughout.

No issues.

Issue #17, opened 2026-06-08
[reduction] krapper_gen code-reduction backlog (~8-12% / ~1.1-1.7k LoC mechanically reducible)


monkopedia-coder:
Baseline read-only reduction analysis of `krapper_gen/src/nativeMain` (~14,085 LoC). Items ranked by value/effort. Each lands as its own reduction PR through the standard pipeline (must keep featuregen 183/0 + krapper 298/0 + clang stage1 green). **Reduce real bloat, don't golf; smaller ≠ better.** Sequence by file to avoid clobbering in-flight feature PRs.

## Safe quick wins (low-risk, mechanical)
1. **Operators.kt:79-81,130-132,151-153,216-218** — 4 byte-identical `Operator.name()` bodies + private `cOp`. Lift into the sealed base (`ConversionOperator` keeps its override). ~9 lines. high-conf.
2. **CppWriter.kt:587-611 vs 613-638** — `createStringReturn`/`createPointedStringReturn` differ only `.`vs`->` + source type; duplicated NUL-term comment. One helper param'd by `byPointer`. ~20 lines. (string-return is correctness-sensitive — re-verify string tests.)
3. **KotlinWriter.kt:470-500,513-553 (+415-418)** — duplicated `existingNames` computation + name-clash guard across cast/downcast/dispose. Extract `memberMethodNames(cls)` + `skipIfNameClash(...)`. ~12 lines.
4. **WrappedElement.kt:192-193,233-234,364,370-375** — ~8 commented-out `CXCursor_… -> TODO()` stubs in a live `when`. Delete. ~8 lines.
5. **KotlinWriter.kt (18 sites) + CastTargets.kt** — hand-inlined `.type.kotlinType.name.trimEnd('?')`. One `ResolvedKotlinType.plainName` extension. (borderline premature-DRY — rank low.)

## Bigger consolidations (medium-risk, need care)
6. **Parsing.kt:755-778,793-816,838-865 (+704-736,867-880) — BIGGEST STRUCTURAL WIN.** 5 copy-pasted recursive method-rewrite passes share one skeleton (match → `copy{}` → remove/add child → recurse); recursion tail duplicated 5×. Collapse into one `rewriteMethods(element, transform)` walker; each pass = a ~6-line predicate+transform. ~40-60 lines + removes a recursion-tail bug class. (**overlaps #9 — do after #9 merges.**)
[...]

PR #21, opened 2026-06-09
Reduce: collapse the 5 copy-pasted method-rewrite passes in Parsing.kt into one walker (#17 item 6)

monkopedia-coder:
## What

Collapses the copy-pasted, structurally-identical recursive method-rewrite passes in `krapper_gen/.../generator/Parsing.kt` onto a single shared walker.

### Walker signature
```kotlin
private suspend fun rewriteMethods(
    element: WrappedElement,
    transform: suspend (owner: WrappedElement, method: WrappedMethod) -> WrappedMethod?
)
```
The walker owns the recursion + the remove/add bookkeeping. `transform` returns:
- the **same** method instance -> leave in place (no change),
- a **different** method -> replace it (`removeChild` + `addChild`),
- `null` -> drop it (`removeChild`).

### Consolidated (4 passes -> lambdas over the walker)
- `dropMistypedInitializerListMembers` — **drop** (returns `null`)
- `rewritePairSecondReturns` — **replace** (returns `copy()`)
- `rewriteViewReturns` — **replace** (returns `copy()`)
- `rewriteUniquePtrReturns` — **replace** (returns `copy()`)

Each pass keeps its **exact** element-type guard inside the lambda (`WrappedClass` for three; `WrappedClass || WrappedNamespace` for unique_ptr), so the precise set of methods examined and modified — plus each pass's match predicate, its `Log.i(...)` wording, and the traversal order — are preserved.

### Left out (deliberately)
- `removeDuplicateMethods` — carries **cross-sibling state** (`signaturesSeen` accumulated across siblings) rather than a stateless per-method transform. It does not fit the shared skeleton cleanly, so it is left as-is. (A partial, clean consolidation over a contorted total one.)

## Behavior-identical proof

**Byte-identical generated output.** Both the unmodified base (`d0743e1`) and this branch were driven through the featuregen `kplusplusSync + nativeTest` gate to the **same fixpoint** (iter1 -> 6 instantiations, iter2 -> 16, iter3 -> 17, green). The full generated trees were md5-manifested and diffed:

> **RESULT: BYTE-IDENTICAL — all 121 generated files match exactly** (118 `.kt` + `featuregen.cc` + `featuregen.h` + `featuregen.def`).

This exercises every consolidated pass: pair-second (set/map `insert`), view (`llvm::StringRef`->`.str()`), initializer-list drop (`std::unordered_map`), and unique_ptr release.

## Verification (same counts as base `d0743e1`)

## Line delta
`1 file changed, 96 insertions(+), 95 deletions(-)` — code-only (excl. blank/comment) **90 removed / 78 added**, net **-12** (the +78 includes the new ~25-line reusable walker added once). Removes **3 of the 5** copy-pasted recursion tails (5 -> 2) — a real bug class.

Refs #17

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Parsing.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Parsing.kt
@@ -833,34 +833,53 @@ suspend fun WrappedTemplate.typedAs(
 private suspend fun dropMistypedInitializerListMembers(
     element: WrappedElement,
     fullyQualified: String
+) = rewriteMethods(element) { owner, method ->
+    if (owner !is WrappedClass) return@rewriteMethods method
+    val typedefNames = owner.children
+        .filterIsInstance<WrappedTypedef>()
+        .map { it.name }
+        .toSet()
+    val isKeyValueMap = "key_type" in typedefNames && "mapped_type" in typedefNames
+    if (!isKeyValueMap) return@rewriteMethods method
+    val hasBareTemplateInitializerList = method.args.any { arg ->
+        val type = arg.type.maybeUnconst.maybeUnreferenced
+        type is WrappedTemplateType &&
+            type.baseType.toString() == "std::initializer_list" &&
+            type.templateArgs.singleOrNull() is WrappedTemplateRef
+    }
+    if (!hasBareTemplateInitializerList) return@rewriteMethods method
+    Log.i(
+        "Dropping mistyped initializer_list member ${method.name} on " +
+            "$fullyQualified (libclang collapsed value_type to a bare " +
+            "template parameter)"
+    )
+    null
+}
+
+/**
+ * Shared method-rewrite walker for the structurally-identical recursive passes below
+ * ([dropMistypedInitializerListMembers], [rewritePairSecondReturns], [rewriteViewReturns],
+ * [rewriteUniquePtrReturns]). Recurses [element]'s subtree; for every [WrappedMethod] child
+ * of every element it calls [transform], which returns:
+ *  - the same method instance -> leave it in place (no change),
+ *  - a different method        -> replace it (removeChild + addChild),
+ *  - `null`                    -> drop it (removeChild).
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Behavior-preservation verified by diff-read against base d0743e1. Each of the 4 consolidated passes is correct:

**dropMistypedInitializerListMembers** — guard `owner !is WrappedClass` matches original `if (element is WrappedClass)`. Predicates (`typedefNames`/`isKeyValueMap`/`hasBareTemplateInitializerList`) and Log.i wording are byte-identical to the original. Drop semantics (null return) correct. Note: `typedefNames`/`isKeyValueMap` is now recomputed per method instead of once per class — a performance note only, not a behavioral difference (we only remove/add methods, never typedefs, during this pass).

**rewritePairSecondReturns** — guard `owner !is WrappedClass`, predicate (`isPairBool`), Log.i wording, `copy(returnType=WrappedType("bool")).also { it.returnsPairSecond = true }` replace semantics: all identical.

**rewriteViewReturns** — guard `owner !is WrappedClass`. Log.i uses `owner.type` vs original `element.type` — these are the same object (the lambda's `owner` IS the element that was `element` in the original), so identical output. All predicate/copy/Log.i details match.

**rewriteUniquePtrReturns** — guard `owner !is WrappedClass && owner !is WrappedNamespace` matches original `if (element is WrappedClass || element is WrappedNamespace)`. Container string derived from same `owner as? WrappedClass / WrappedNamespace` casts. Match predicate, copy, Log.i: identical.

**Walker safety** — `filterIsInstance<WrappedMethod>()` snapshots a new list before the mutating loop, so removeChild/addChild during iteration is safe. Recursion tail iterates children after all method mutations are done: no double-visit, no skipped subtrees.

**removeDuplicateMethods** — confirmed left untouched and identical to base; cross-sibling `signaturesSeen` state correctly excluded from the walker.

**`owner` not vestigial** — all 4 lambdas use `owner` for the element-type guard and/or container-name derivation.

**Byte-identical proof** — md5 manifest over 121 generated files (118 .kt + featuregen.cc/.h/.def), base d0743e1 vs branch; krapper_gen 304/0, featuregen 186/0, ktlint green. Behavior-preservation is solid.

PR #22, opened 2026-06-09
Wire :featuregen:nativeTest to depend on kplusplusSync (no more stale-krapped flake) (#16)

### LINKED ISSUE #16: featuregen:nativeTest can run on STALE generated sources (no kplusplusSync dependency) — masks/﻿mimics regressions
**Verification-reliability hazard (recurring).** `:featuregen:nativeTest` does not depend on `:featuregen:kplusplusSync`, so when run as its own Gradle invocation it compiles against whatever generated sources happen to be on disk. If those are stale/partial (e.g. an incremental sync that bound 14 instantiations instead of the full 17), the test fails on unresolved bindings like `kplusplus-sync-required: std::vector<std::vector<int>>` / `CvNestedTest`.

**Why it matters:** in the last few PRs, *three* coder agents independently hit this, each had to bisect (stash changes → reproduce on pristine base) to convince themselves it was staleness and not their change. That's wasted effort every time — and worse, it trains agents to dismiss a `featuregen:nativeTest` red as "the known artifact," which could let a REAL generation regression ship. The current safe workaround is running `:featuregen:kplusplusSync :featuregen:nativeTest` in ONE invocation so Gradle orders sync before the test compile; that ordering should be guaranteed, not incidental.

**Do:** make `:featuregen:nativeTest` (or the test-compile task / the cinterop-or-source-set it consumes) properly depend on `:featuregen:kplusplusSync`, OR mark the generated `krapped/` sources as an input wired through task outputs so Gradle's up-to-date checking regenerates before the test runs. Acceptance: `./gradlew :featuregen:nativeTest` from clean (or after a krapper_gen change) regenerates first and is green without a manual sync step.

Filed by the repo agent after observing the recurring artifact across PRs #14/#15 and the #9 work.

monkopedia-coder:
## Problem (#16)

`:featuregen:nativeTest` did not depend on `:featuregen:kplusplusSync`. Run as its own Gradle invocation it compiled against whatever generated `krapped/` sources happened to be on disk; if stale/partial it failed on `kplusplus-sync-required: …` / unresolved bindings (`CvNestedTest`). Three coder agents independently hit this and bisected staleness, and it risks training agents to dismiss a real generation regression as "the known artifact."

## The wiring

1. **`compileKotlinNative` depends on `kplusplusSync`** — and the test compile depends on the main compile, so the generator runs (and Gradle's up-to-date check regenerates stale output) before anything compiles against `krapped/`.
2. **`kplusplusSync` now declares real inputs/outputs** — the generator kexe, the compiler-written `requested.txt` manifest, the configured headers, and the extension config that becomes krapper_gen CLI args, plus the `krapped/` output dir. (The header-driven path keeps its existing always-regenerate behaviour by declaring no output when headers are configured — there's no per-method change tracking there.)
3. **The cinterop `.def` + `krapped/src` source dir are now wired UNCONDITIONALLY** (in an `afterEvaluate`, gated on the project actually being configured to generate bindings). Previously they were wired only if the artifacts already existed at *configuration* time — that was the chicken-and-egg: a stale/missing def at configure time meant the `kplusplus` cinterop was never created, so a sync that regenerated the def couldn't be picked up in the same invocation and the compile failed on the unresolved `<module>` cinterop package. The cinterop task also depends on `kplusplusSync` so it can't process a stale def.

The fix lives in the kplusplus Gradle plugin, so every consumer (not just featuregen) gets the guarantee.

## Acceptance — standalone regenerates + greens (the point)

With a converged manifest, deliberately staled `krapped/` (deleted `featuregen.def` + `src/CppBinding.kt`), then ran ONLY `./gradlew :featuregen:nativeTest`:

```
> Task :featuregen:kplusplusSync
kplusplusSync: headers=1 regenerating 17 instantiation(s): Box<…>, …, std::vector<std::vector<int>>
> Task :featuregen:cinteropKplusplusNative
> Task :featuregen:compileKotlinNative
> Task :featuregen:compileTestKotlinNative
BUILD SUCCESSFUL
```

`kplusplusSync` ran first and regenerated all 17 instantiations, then the cinterop/compile/test ran green — no manual sync. Confirmed the *old* behaviour failed the same scenario on the unresolved `featuregen` cinterop package before the unconditional-wiring change.

## Verification (full gate)

`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL:
- krapper_gen: **304/0** (baseline)
- featuregen: **186/0**
- ktlint: green

Plugin compiles clean (only the pre-existing `defFile` deprecation warning).

Closes #16

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -64,6 +64,43 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
             val generatedFile = File(krappedDir, "generated.txt")
             val fixupFile = File(krappedDir, "fixups.json")
             val moduleName = target.name
+            // Declare up-to-date inputs/outputs so Gradle can (a) order this task
+            // ahead of the native compile that consumes krapped/ (see
+            // wireGeneratedBindings), and (b) skip the regeneration when nothing
+            // that feeds it has changed. Inputs: the generator kexe, the
+            // compiler-written manifest, the configured headers, and the
+            // extension config that becomes krapper_gen CLI args. Output: the
+            // whole krapped/ dir (the kexe regenerates it wholesale each run).
+            //
+            // The header-driven path intentionally has no per-method change
+            // tracking (see the hasHeaders force-regenerate note below), so when
+            // headers are configured we leave the task always out-of-date by
+            // declaring no output — Gradle then re-runs it every invocation,
+            // which is the existing, correct behaviour for that path.
+            val ext = target.extensions.findByType(KPlusPlusExtension::class.java)
+            val hasHeadersAtConfig = ext?.headers?.isNotEmpty() == true
+            it.inputs.file(kexe).withPropertyName("krapperGenKexe")
+                .withPathSensitivity(org.gradle.api.tasks.PathSensitivity.NONE)
+            it.inputs.files(manifestFile).withPropertyName("requestedManifest")
+                .withPathSensitivity(org.gradle.api.tasks.PathSensitivity.NONE)
+                .optional(true)
+            ext?.let { e ->
+                it.inputs.property("instantiations", e.instantiations.sorted())
+                it.inputs.property("referencePolicy", e.referencePolicy ?: "")
+                it.inputs.property("cppStandard", e.cppStandard ?: "")
+                it.inputs.property("rootPackage", e.rootPackage ?: "")
+                it.inputs.property("only", e.only.sorted())
+                it.inputs.property("onlyFile", e.onlyFile ?: "")
+                it.inputs.property("fixups", e.fixups.toJsonArray())
+                if (e.headers.isNotEmpty()) {
+                    it.inputs.files(e.headers.map { h -> target.file(h) })
+                        .withPropertyName("headers")
+                        .withPathSensitivity(org.gradle.api.tasks.PathSensitivity.RELATIVE)
+                }
+            }
+            if (!hasHeadersAtConfig) {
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Reviewed against all four scrutiny points — this is correct and safe to merge.

**Gating correctness (scrutiny 1):** The `afterEvaluate` guard in `wireGeneratedBindings` is sound. `willGenerate` covers all three binding-config paths (headers, instantiations, pre-existing manifest); the additional `!krappedDef.exists() && !krappedSrc.exists()` fallback means a project with old artifacts but no live config still wires correctly. A project that uses the plugin but declares no bindings and has no artifacts gets `return@afterEvaluate` — clean configure, no phantom cinterop.

**Inputs/outputs correctness (scrutiny 2):** Inputs cover the kexe, manifest (optional), and all extension fields that drive the generated output (instantiations, referencePolicy, cppStandard, rootPackage, only, onlyFile, fixups, headers). The two omissions (`compiler` and `libraries`) are correct: those fields influence the wrapper *library* compile but not the generated Kotlin/def content, so omitting them doesn't leave stale bindings. The header-driven always-regenerate (no declared output) is intentional, clearly commented, and matches pre-PR behavior.

**Other consumers (scrutiny 3):** Both `example` (headers + fixups) and `clangwalk` (headers + only + IGNORE_MISSING) have non-empty headers, so `willGenerate = true` and the afterEvaluate wires unconditionally — same or better than before. clangwalk's header-driven path keeps always-regenerate (no output declared). No regression risk.

**dependsOn chain (scrutiny 4):** `kplusplusSync → cinteropKplusplusNative → compileKotlinNative → compileTestKotlinNative`. The direct `compileKotlinNative.dependsOn(kplusplusSync)` is redundant-but-harmless (cinterop already chains it), and defensively covers the `krapped/src` srcDir path. No circular dependency.

Acceptance scenario (deleted def + src, standalone `:featuregen:nativeTest` regenerates 17 instantiations → BUILD SUCCESSFUL) directly validates the fix. Full gate 304/0 + 186/0 + ktlint green.

PR #23, opened 2026-06-09
Reduce: dedup the 4 identical Operator.name() bodies + drop dead cursor-kind stubs (#17 items 1,4)

monkopedia-coder:
Two behavior-preserving line-removers from the reduction backlog (#17).

## Item 1 — dedup the 4 identical `Operator.name()` bodies (Operators.kt)
`BasicBinaryOperator`, `BasicAssignmentOperator`, `BasicCallOperator`, and `BasicUnaryOperator` each carried a **byte-identical** `name()` body and each privately stored `cOp`. Lifted both into a small intermediate abstract class `Operator.NamedByCOp`:

- `NamedByCOp` holds `protected abstract val cOp` and the default `name()` = `<cName>_op_<cOp split + lowercased>`.
- The 4 subclasses now just declare `override val cOp` and extend `Operator.NamedByCOp()` — their duplicate `name()` bodies are gone.
- `ConversionOperator` is untouched: it names by destination type and keeps its own `name()` override.

**Operator C-names are unchanged** — the lifted body is the same expression, so every generated `<cName>_op_…` wrapper name is identical. Confirmed by the codegen / full-Kotlin tests staying green at the same count.

## Item 4 — delete dead commented-out cursor-kind stubs (WrappedElement.kt)
Removed 9 dead `// CXCursorKind.CXCursor_… -> TODO()` lines interleaved in the live `when (value.kind)`. They were unreachable (the `else`/default handles unhandled kinds) and only fragmented the branch list. Pure comment deletion, zero behavior change. The live branches and the default are untouched.

## Line delta
**Net −14** (19 insertions / 33 deletions across the 2 files).

## Verification (JAVA_HOME=java-17)

All at the exact baseline counts.

Refs #17

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/Operators.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/Operators.kt
@@ -25,6 +25,17 @@ sealed class Operator {
     abstract fun name(namer: Namer, method: WrappedMethod): String
     protected abstract fun matches(method: WrappedMethod): Boolean
 
+    // Shared base for the operators whose C wrapper name is derived mechanically
+    // from a C-spelling fragment (`cOp`): `<cName>_op_<cOp split + lowercased>`.
+    // ConversionOperator does not extend this — it names by destination type.
+    abstract class NamedByCOp : Operator() {
+        protected abstract val cOp: String
+
+        override fun name(namer: Namer, method: WrappedMethod): String = with(namer) {
+            cName + "_op_" + cOp.splitCamelcase().joinToString("_") { it.lowercase() }
+        }
+    }
+
     companion object {
         val ALL_OPERATORS = listOf(
             BasicBinaryOperator.MINUS,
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Clean refactor — all 4 scrutiny points pass.

**Item 1 — Operator C-names unchanged (verified against source):**
The lifted `NamedByCOp.name()` body is `cName + "_op_" + cOp.splitCamelcase().joinToString("_") { it.lowercase() }` — byte-for-byte identical to the 4 original bodies (the elided `return` keyword is semantically identical in a single-expression override). Each subclass's `cOp` constructor values are unchanged (`"Minus"`, `"Plus"`, `"Divide"`, `"Assign"`, `"PlusEquals"`, `"Call"`, `"Increment"`, etc.). Generated C wrapper names are bit-for-bit identical.

**Item 2 — ConversionOperator untouched:** still extends `Operator()` directly, not `NamedByCOp`, and its own `name()` (`cName + "_op_to_" + targetOf(method)!!`) is unmodified.

**Item 3 — Dead-stub deletion is pure comment removal:** only `// CXCursorKind.CXCursor_… -> TODO()` lines were removed; all live `when` branches and the `else`/default are untouched.

**Item 4 — Quality:** `NamedByCOp` is appropriately named, well-commented, and correctly scoped as an `abstract class` (needed for the abstract `val cOp`). 4 identical uses easily clears the dedup threshold. No over-engineering.

Tests: krapper_gen 304/0, featuregen 186/0, ktlint green (same counts as baseline — consistent with zero behavior change). No CI configured on this repo; manual verification is the standard here.

Issue #25, opened 2026-06-09
kplusplusSync cold-start: nested-generic manifest needs multiple passes to converge (fresh worktree friction)


monkopedia-coder:
Distinct from #16 (which fixed the STALENESS flake — a stale `krapped/` now regenerates). This is the **cold-start convergence** problem: on a FRESH checkout/worktree with an empty `requested.txt`, the first `kplusplusSync` doesn't know about SYNC_REQUIRED nested generics (e.g. `std::vector<std::vector<int>>`, only discovered once `std::vector<int>` is bound and the FIR plugin compiles against it and writes `requested.txt`). So the first sync regenerates a partial set, the test compile fails on `kplusplus-sync-required: std::vector<std::vector<int>>`, and a SECOND sync is needed to materialize them. The manifest converges over compile passes.

**Impact:** every fresh agent worktree (and any cold CI checkout) hits this — agents repeatedly report it, bisect to confirm it's not their change, and work around it by running sync twice / a combined invocation. It's recurring friction and a foot-gun (looks like a failure on first run). Confirmed reproduces on clean base af71634 (post-#16).

**Do:** make `kplusplusSync` (or the sync→compile→sync cycle) CONVERGE the SYNC_REQUIRED manifest on cold start — e.g. iterate sync until `requested.txt` is stable (fixpoint), or seed the manifest from a discovery pass, so a single `:featuregen:nativeTest` from a cold worktree reaches green without a manual second sync. Acceptance: from a clean checkout with empty `krapped/`, one `:featuregen:nativeTest` converges + passes.

Filed after multiple reduction agents (#21/#23/#24 work) independently hit this on fresh worktrees.

PR #26, opened 2026-06-09
Reduce: extract the duplicated name-clash scaffolding in KotlinWriter (#17 item 3)

monkopedia-coder:
## What

`generateCastMethods`, `generateDownCastMethods`, and `generateDisposeMethods` in `KotlinWriter.kt` each recomputed the same set of member-method names, and the two cast generators repeated an identical name-clash guard (a `comment(...)` NOTE + `continue`). Extracted the duplication:

- **`private fun memberMethodNames(cls: ResolvedClass): Set<String>`** — the shared `children.filterIsInstance<ResolvedMethod>().filter { METHOD }.mapTo { fixNaming(it).name }` computation. Now used by all three generators. `generateDisposeMethods`'s equivalent `any { name == "dispose" || name == "owned" }` becomes `"dispose" in names || "owned" in names`.
- **`private fun KotlinCodeBuilder.skipIfNameClash(name, existing): Boolean`** — the shared cast-generator clash guard: emits the NOTE comment and returns `true` so the caller `continue`s. Replaces the two byte-identical inline guards.

## Behavior-preserving — byte-identical output

Skip decisions and the emitted NOTE wording are unchanged. The fully-converged `featuregen/krapped` md5 manifest is **byte-identical** before (clean e0e4c18) and after — all 124 files match (`diff` returns no differences).

## Verification

`JAVA_HOME=…java-17-openjdk ./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck`

- krapper_gen nativeTest: **304/0**
- featuregen nativeTest: **186/0**
- ktlintCheck: green

(Cold worktree needed the known #25 sync→compile→sync manifest convergence before the gate went green; that's convergence, not this change.)

## Net delta

**-2** source lines (22 insertions / 24 deletions).

refs #17

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -397,6 +397,22 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
         appendLine()
     }
 
+    // Kotlin names of the class's own member methods — used to detect when a synthesized member
+    // (dispose/owned, up-/down-cast) would clash with a real one.
+    private fun memberMethodNames(cls: ResolvedClass): Set<String> =
+        cls.children.filterIsInstance<ResolvedMethod>()
+            .filter { it.methodType == MethodType.METHOD }
+            .mapTo(mutableSetOf()) { fixNaming(it).name }
+
+    // Cast-generator name-clash guard: if `name` is already a member, emit the NOTE and skip.
+    private fun KotlinCodeBuilder.skipIfNameClash(name: String, existing: Set<String>): Boolean {
+        if (name !in existing) return false
+        comment(
+            "NOTE: $name() not generated — this class already declares a member with that name."
+        )
+        return true
+    }
+
     // Caller-controlled destruction (matrix: OB-return-ptr-owned / IH-destruction). Every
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
**Reviewed as monkopedia-reviewer — tier 1 auto.**

**PR:** Monkopedia/kplusplus#26
**Category:** `refactor`
**Tier:** 1 (kplusplus baseline `* : tier 1`)
**Verdict:** Approve

---

Behavior-preserving dedup verified on both axes:

**Correctness (axis A)**
- `memberMethodNames` extracts the identical three-way computation (`filterIsInstance<ResolvedMethod>().filter { METHOD }.mapTo { fixNaming(it).name }`). All three call sites use it; no change to which names are collected.
- `skipIfNameClash` reproduces the two byte-identical inline guards exactly: emits the same NOTE comment wording, returns `true` so the caller `continue`s. Return-`false`-on-no-clash path is the correct "proceed" signal.
- `generateDisposeMethods` rewrite: `any { METHOD && (name=="dispose" || name=="owned") }` → `"dispose" in existingNames || "owned" in existingNames` is equivalent — `memberMethodNames` already filters on METHOD and applies fixNaming, so membership is identical.
- 124-file md5 byte-identical proof accepted. Generated output unchanged.

**Quality (axis B)**
- Helpers are well-named and scoped (`private`). Doc comments match surrounding file style.
- 3 call sites fully justify `memberMethodNames` extraction. Not over-engineered.
- No UI/UX or performance paths touched; reviewer-judgment lever does not fire.

krapper_gen 304/0 · featuregen 186/0 · ktlint green — tests pass.

Issue #11, opened 2026-06-08
[clang-campaign] Productionization: hermeticity (/tmp + globals), llvm-config, version/compiler runbook


monkopedia-coder:
**Campaign productionization/hermeticity (Clang self-bootstrap).**
- **`/tmp` non-hermetic forcing** — `IndexedServiceImpl.kt:136` writes `/tmp/krapper_inst_<forceName>.h` (fixed name → concurrent runs collide), never cleaned up; also `Parsing.kt:373 /tmp/clang_includes.c`. → unique temp dir per run + cleanup.
- **Process-global mutable state** — `WrappedType.kt existingTypes`, `WrappedKotlinType.kt rootPackageOverride` (set in `IndexedServiceImpl.init`). Makes runs order-sensitive (doc:202); blocks Gradle-daemon reuse / parallel module builds. → thread through a context object.
- **Multi-pass re-resolve cost** — `Resolver.resolveForcing` is O(instantiations × bound); fine at 3, dominant at dozens. → batch where possible.
- **Consumer hardcoding** — `clangwalk/build.gradle.kts` hardcodes `-lLLVM-22 -L/usr/lib`; derive via `llvm-config`. Document the LLVM-pin (22.1.6) + regen-on-bump runbook, and the `compiler="clang++"`-not-konan-GCC-8.3 requirement (the konan GCC-8.3 aborts on Clang headers; also the source of the deferred GCC-8.3 pair-swap gap).

Split into sub-PRs if cleaner. Acceptance: hermetic runs, no hardcoded toolchain paths, documented version/compiler story.

monkopedia-coder:
Finding from the #18 forcing-test work, relevant to the 'globals' axis here: **libclang in-process global state leaks across forcing runs in one JVM.** Driving libclang through repeated `<vector>`-instantiating forcing TUs within a single test JVM *degrades* `std::vector`/`iterator_range` canonicalization for LATER work in that process — it silently broke `ReturnShapeTest.iteratorRangeReturnMaterializesVector` until the new test was constrained to a single cached forcing run. Two implications for hermeticity: (1) tests that drive forcing need isolation/budgeting (or a per-run fresh index), and (2) if the generator ever does multiple forcing passes in one process, the same canonicalization drift could bite production. Worth auditing the libclang index/TU lifecycle for shared mutable state as part of this ticket.

monkopedia-coder:
[...]

PR #27, opened 2026-06-09
Hermeticity: unique per-run temp files + cleanup (no more fixed /tmp collisions) (#11 part 1)

monkopedia-coder:
## Scope (issue #11, part 1 of N — the `/tmp` part only)

The generator wrote two **fixed-name** temp files that concurrent runs clobbered and that were never cleaned up:
- `IndexedServiceImpl` → `/tmp/krapper_inst_<forceName>.h` (template-forcing headers)
- `Parsing.generateIncludes` → `/tmp/clang_includes.c` (include-path probe)

Parallel module builds / concurrent agent runs with the same `forceName` raced on identical paths, and the files accumulated in `/tmp` forever.

## The fix

- **`File.createTempDir(prefix)`** — a process-unique temp dir under `$TMPDIR` (default `/tmp`), named `<prefix>_<pid>_<atomic-counter>`. Distinct pids (concurrent runs) and the per-process counter (repeated calls in one run) guarantee no collision.
- **`generateIncludes()`** writes `clang_includes.c` into a per-run dir and `rmR()`s it via a `memScoped` `defer` (cleaned even on throw). This file is pure throwaway — it never appears in generated output.
- **Forcing headers**: parsed from a per-run temp dir, then **re-materialized into the output dir** by `writeTo()` so the generated `.cc` `#include`s them by a stable, relative, self-contained path (`KrapperForce_*.h`) instead of a fixed `/tmp` path. The per-run temp dir is `rmR()`'d in `close()`.
- **`close()`**: dropped the redundant explicit `index.dispose()` — the `Arena`'s deferred dispose in `scope.clear()` already releases the `CXIndex`; doing both double-freed it (`clang_disposeIndex` on a freed handle → SIGABRT). This latent bug only surfaced once `close()` began being called from the CLI path.
- **App CLI path** now wraps the run in `try { … } finally { indexService.close() }`, so the temp dir is always cleaned (and the index released) even on error. This is the path `gradle :…:kplusplusSync` invokes, so `/tmp` no longer accumulates.

## Behaviour-preserving

Regenerated `featuregen/krapped` with the base (`a01349f`) kexe and this branch's kexe over the same 17-instantiation set, then diffed:
- **all 112 Kotlin bindings `src/*.kt`: byte-identical**
- **`feature.h`: byte-identical**
- **`feature.def`: identical** (modulo the test harness's distinct `-o` dir)
- **`feature.cc`: ONLY the forcing-header `#include` paths change** — from the old non-hermetic `/tmp` refs (`../krapper_inst_*.h`, dead after the run) to self-contained siblings (`KrapperForce_*.h`). The `.cc` is compiled into `lib<module>.a` during `writeTo` and never recompiled by the consumer (cinterop reads `feature.h` + the `.a`), so this change is functionally inert — proven by featuregen staying 186/0.

**No temp files leak**: after a run, no `/tmp/krapper_inst_*` or `/tmp/krapper_includes_*` remain. (Baseline left 25+ stale `krapper_inst` files in `/tmp`.)

## Verification (JDK 17)

- `:featuregen:kplusplusSync` — green (needed the documented #25 manifest-convergence passes on this cold worktree)

Refs #11 (does **not** close it — globals threading, batch re-resolve, and the clangwalk `llvm-config` derivation remain separate sub-PRs).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
@@ -188,52 +188,53 @@ class KrapperGen : CliktCommand() {
                 )
             )
             val indexService = service.index(IndexRequest(header, library))
-            // v2 flow: when both --header and --instantiate are present, parse
-            // the headers first (filterAndResolve) then layer the requested
-            // template instantiations on top. This is what the v8 example
-            // needs — a "wrap this whole header set" import combined with the
-            // narrow fixups passed via --fixup-file.
-            val hasHeaders = header.isNotEmpty()
-            val hasInstantiations = instantiate.isNotEmpty()
-            if (hasHeaders) {
-                // Scoped-import allowlist (T1.0a): when --only/--only-file name an
-                // explicit class set, bind ONLY those (referenced-but-unlisted types
-                // fall to --referencePolicy). Otherwise keep the DefaultFilter
-                // (bind-everything-non-std) behavior.
-                val allowList = loadAllowList()
-                val filter = if (allowList.isEmpty()) {
-                    DefaultFilter
-                } else {
-                    Log.i("Scoped import: binding only ${allowList.size} class(es): $allowList")
-                    AllowListFilter(allowList)
+            try {
+                // v2 flow: when both --header and --instantiate are present, parse
+                // the headers first (filterAndResolve) then layer the requested
+                // template instantiations on top. This is what the v8 example
+                // needs — a "wrap this whole header set" import combined with the
+                // narrow fixups passed via --fixup-file.
+                val hasHeaders = header.isNotEmpty()
+                val hasInstantiations = instantiate.isNotEmpty()
+                if (hasHeaders) {
+                    // Scoped-import allowlist (T1.0a): when --only/--only-file name an
+                    // explicit class set, bind ONLY those (referenced-but-unlisted types
+                    // fall to --referencePolicy). Otherwise keep the DefaultFilter
+                    // (bind-everything-non-std) behavior.
+                    val allowList = loadAllowList()
+                    val filter = if (allowList.isEmpty()) {
+                        DefaultFilter
+                    } else {
+                        Log.i("Scoped import: binding only ${allowList.size} class(es): $allowList")
+                        AllowListFilter(allowList)
+                    }
+                    indexService.filterAndResolve(filter)
                 }
-                indexService.filterAndResolve(filter)
-            }
-            if (hasInstantiations) {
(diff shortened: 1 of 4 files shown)


monkopedia-reviewer (approved):
## Review: Hermeticity — unique per-run temp files + cleanup (#11 part 1)

**Category:** `bug_fix` + `feature_internal` (concurrent-collision fix + cleanup lifecycle) — kplusplus baseline `* : tier 1`, no escalation.

### Verdict on the three load-bearing points

**1. Double-free fix — SOUND.**
The Arena's `init` block registers `scope.defer { index.dispose() }` (line 97–99 post-PR). `close()` calls `scope.clear()`, which fires that deferred dispose exactly once. The old explicit `index.dispose()` immediately before `scope.clear()` has been removed. There is no remaining double-dispose path and no use-after-free. This was a latent SIGABRT that only becomes live when `close()` is first called — wiring the try/finally in App was the right moment to excise it.

**2. `.cc` include-path change — APPROVED.**
Both conditions for "functionally inert" are confirmed:
  - The `KrapperForce_*.h` siblings are written to `outputBase` in the `writeTo` loop *before* `CppCompiler.compile(cppFile, headers, ...)` is called. `CompileFlags.includeDirs` emits `-I<outputBase>` (parent-of-headers), so the `#include "KrapperForce_xyz.h"` lines in the `.cc` resolve at compile time.
  - `CppCompiler` compiles the `.cc` into `lib*.a` during `writeTo`. The consumer (cinterop) links the prebuilt `.a` and reads `feature.h` + `feature.def`; it does not recompile the `.cc`. The post-compile include paths are dead. The old `/tmp`-relative refs were dangling from the moment the run ended; the sibling paths are self-contained and deterministic.

**3. Hermeticity — ACHIEVED.**
`createTempDir` uses `AtomicInt.addAndGet(1)` (atomic increment, returns new value — no TOCTOU) plus `getpid()` → concurrent runs at distinct pids never collide; repeated calls in one run get a distinct counter. `generateIncludes()` defers `rmR()` inside `memScoped` — cleaned even on throw. `IndexedServiceImpl.close()` does `tempDir?.rmR()` on the non-null guard. The App try/finally covers ALL exit paths including the `return@runBlocking` early-exit (Kotlin `finally` runs on `return@label`), so the `kplusplusSync` Gradle path cleans up on both success and error.

### Quality
`createTempDir` is idiomatic: kdoc, clear ownership comment, `$TMPDIR` fallback correctly handles trailing slash + null/empty, lazy `tempDir()` initialization means a no-instantiation run touches nothing. The forced-header lifecycle comment thread (forcingHeaders → temp write → writeTo re-materialize → syntheticHeaders → compile) is well-documented. No concerns.

APPROVED — safe to merge.

PR #29, opened 2026-06-09
Reduce: consolidate the 3 top-level-split helpers into one (#17)

monkopedia-coder:
## What

Three functions hand-rolled the same "split on top-level commas, respecting bracket depth" logic. This extracts one shared helper and routes the two that share a shape through it.

### Shared helper
```kotlin
internal fun splitTopLevel(s: String, opens: String = "<", closes: String = ">"): List<String>
```
Splits `s` on top-level commas, tracking nesting depth across the position-paired bracket set `opens`/`closes`. Drops the delimiting commas, preserves every other char (incl. brackets), always returns >=1 segment, does NOT trim — callers keep their own trimming / trailing-blank handling.

### Consolidated (2 of 3)
- **`splitTopLevelArgs`** (KotlinWriter, `<>` only) -> `splitTopLevel(args)` then drop a trailing blank segment (preserving its old `isNotBlank()` final-segment rule).
- **`parseFunctionType`** arg-split (KotlinWriter, `()`+`<>`) -> `splitTopLevel(inner, "(<", ")>").map { it.trim() }` (keeps all segments incl. trailing blank, trims each — exactly as before).

### Left out (1 of 3) — and why
- **`findTemplates`** (`model/WrappedKotlinType.kt`) is a genuinely different shape and is **not** folded in:
  - returns `List<IntRange>` (index ranges), not `List<String>`;
  - uses recursive `findEnd` index math that **throws** on unbalanced `>`, and that helper is **shared with `findQualifiers`**;
  - tracks only `<>` (no `()`), and **includes the leading comma** in each range (the caller `parseTypes` does `.trimStart(',')`).
  Forcing it into the string-splitter shape would change `parseTypes`' contract and silently drop the throw-on-malformed behavior. Correctness over a forced total consolidation — it stays as-is.

## Behavior-preserving — byte-identical
`featuregen/krapped` md5 manifest is **byte-for-byte identical** between clean `725f76e` (converged) and this change: 141 files, manifest md5 `4e5258dab1b16eb92e89221ebe6a2e13` on both. `diff` reports zero differences.

## Net line delta
`1 file changed, +44/-44` raw; the real change removes two duplicated depth-tracking loops (~26 + ~18 lines) in favor of one ~30-line shared helper plus two short delegations — the duplicated loop now exists once instead of twice.

## Verification (all green, baseline counts)
- (cold worktree needed the #25 manifest-convergence sync passes to reach a stable 17 instantiations; not part of this change)

refs #17

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -859,35 +859,11 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
     }
 
     // Split a comma-separated template-arg list at top level only (so a nested
-    // `Inner<a, b>` arg stays intact).
+    // `Inner<a, b>` arg stays intact). A trailing blank segment is dropped (so an
+    // empty arg list yields no entries); segments are returned untrimmed.
     private fun splitTopLevelArgs(args: String): List<String> {
-        val result = mutableListOf<String>()
-        var depth = 0
-        val current = StringBuilder()
-        for (c in args) {
-            when (c) {
-                '<' -> {
-                    depth++
-                    current.append(c)
-                }
-
-                '>' -> {
-                    depth--
-                    current.append(c)
-                }
-
-                ',' -> if (depth == 0) {
-                    result.add(current.toString())
-                    current.clear()
-                } else {
-                    current.append(c)
-                }
-
-                else -> current.append(c)
-            }
-        }
-        if (current.isNotBlank()) result.add(current.toString())
-        return result
+        val segments = splitTopLevel(args)
+        return if (segments.last().isBlank()) segments.dropLast(1) else segments
     }
 
     /**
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
**Reviewed as monkopedia-reviewer. LGTM — approving.**

**PR:** #29 — Reduce: consolidate the 3 top-level-split helpers into one
**Category:** refactor
**Tier:** 1 (kplusplus baseline `* : tier 1`)
**Verdict:** Approve + merge

---

**Correctness checks (all pass):**

- `splitTopLevel` uses `c in opens` / `c in closes` (string-contains), so multi-char `"(<"` / `")>"` correctly increments depth for either `(` or `<` and decrements for either `)` or `>` — exactly matching the original `parseFunctionType` inline `'(', '<' -> depth++` / `')', '>' -> depth--` logic.
- `splitTopLevelArgs` delegation: `splitTopLevel(args)` (defaults `<>` only) → drop trailing blank segment when `isBlank()`. Matches the original `if (current.isNotBlank()) result.add(...)` final-segment guard exactly.
- `parseFunctionType` delegation: `splitTopLevel(inner, "(<", ")>")`.map { it.trim() }` — keeps all segments including trailing blank, trims each. Matches original `inner.substring(start, i).trim()` + final `inner.substring(start).trim()` exactly.
- `findTemplates` (WrappedKotlinType.kt) is not in the diff — correctly left out: it returns `List<IntRange>`, uses recursive `findEnd` that throws on unbalanced `>`, includes leading commas, and is shared with `findQualifiers`. Not the same shape; partial-clean call is right.
- Only `KotlinWriter.kt` is in the diff.
- Byte-identical output (141 files, md5 `4e5258dab1b16eb92e89221ebe6a2e13`) accepted. Tests: krapper_gen 304/0, featuregen 186/0, ktlint green.

**Quality:** `splitTopLevel` is clean, correctly `internal`, not over-engineered. Doc comment is accurate.

PR #30, opened 2026-06-09
Codegen: get/at-keyed concrete element + RangeReturn drop visibility (#7 items 3,5)

### LINKED ISSUE #7: [clang-campaign] Codegen-correctness cluster: enum.fromValue crash, keyword method names, get/at iterator, op-table typo, RangeReturn log
**Campaign codegen-correctness cluster (Clang self-bootstrap).** Real bugs found by the audit; mostly small. (Originally dispatched but the agent was rate-limited before committing — none of these landed.)

1. **enum `fromValue` crashes on an unknown int** — `KotlinWriter.renderEnum` emits `entries.first { it.value == v }`; a flag/bitmask enum or a newer-ABI value throws `NoSuchElementException` at runtime via `generateReturn`'s isEnum branch. → `firstOrNull ?: <safe>` (prefer making the path nullable if the getter type already is).
2. **Method names that are Kotlin HARD keywords aren't escaped** — `escapeKotlinKeyword()` is wired only into `define()` (params); `KotlinWriter.kotlinMethodName()` isn't, so a C++ method `in`/`is`/`when` emits `fun in(...)` (won't parse).
3. **`operator[]` concrete-element special-case misses `get`/`at`-keyed containers** — `generateBasicMethod` keeps the concrete element type only for `ResolvedOperator.IND`, but `detectIndexIterable` also accepts `get`/`at`; a `get`/`at`-keyed container with a base-typed element gets `next(): Concrete = at(i): BaseApi` (won't typecheck).
4. **`kotlinMethodName` operator-table typo** — maps `==`→`_cmd` (should be `_cmp`, as `cleanupName` has). Three parallel hand-maintained operator-symbol tables now exist (`cleanupName`/`kotlinMethodName`/`cleanupOperatorName`) — at least cross-reference them.
5. **`RangeReturn.elementOfIterator` drops silently** (`:107-110`) — add a `Log.w` so a dropped range method is discoverable.

Each: featuregen fixture that fails without the fix; update `CppCodeTests`/golden expectations if emitted code changes. Baseline: krapper 293/0, featuregen 179/0.

monkopedia-coder:
Lands the remaining two items of #7 (items 1/2/4 already landed in v2).

## Item 3 — `operator[]` concrete-element special-case missed `get`/`at`-keyed containers
`generateBasicMethod` kept the CONCRETE element type only for `ResolvedOperator.IND`, but `detectIndexIterable` ALSO accepts a `get`/`at` index-get. A `get`/`at`-keyed container whose element is a base class emitted `next(): Concrete = at(i): BaseApi` — the get's return was upcast to the base interface, while the synthesized `iterator()`'s `next()` declares the concrete element type, so it didn't typecheck (the featuregen module failed to compile).

Fix: thread the container identity into the method. New `keepConcreteElement` flag on `generateBasicMethod`, set from a new `isIndexGetMethod(cls, method)` predicate that reuses `detectIndexIterable`'s exact selection (now also exposing the chosen index-get method). The `get`/`at` index-get gets the same concrete-element treatment as `[]`.

## Item 5 — `RangeReturn.elementOfIterator` dropped silently
It returned `null` when it recognized an `iterator_range<It>` but couldn't recover the element from `It`. Now records the drop into `DropLedger` (PARSE phase, with a clear reason) so it's discoverable. The drop DECISION is unchanged. `DropLedger.record` is non-suspend, so it's callable from this non-suspend path (resolves the earlier suspend-`Log.w` blocker; `DropLedger` merged via #6).

## Fixtures / tests
- **featuregen** `CvGetAtBaseElemTest` + a `ShapeBag` fixture (a `get`/`at`-keyed container with a `Shape*` base element). Without the fix, the synthesized iterator's `next()` vs the upcast `at()` don't typecheck and the module fails to compile; with it the concrete Shapes round-trip and `area()` virtual-dispatches. Verified in the generated binding: `at(i): Shape?` and `next(): Shape? = at(__i++)` line up.
- **krapper_gen** `ReturnShapeTest.unextractableIteratorRangeRecordsDrop`: an `iterator_range` over an opaque iterator (no `value_type`/template-arg) lands a PARSE `DropLedger` record with a reason, while the sibling plain method still binds.

## Verification (REAL gates)
- `:krapper_gen:nativeTest` — **305/0** (304 baseline + 1 new)
- `:featuregen:nativeTest` — **188/0** (186 baseline + 2 new)

No goldens changed: the `keepConcreteElement` flag defaults false, so only `get`/`at` base-element containers (absent from `CppCodeTests`/`KotlinCodeTests`) are affected. Cold worktree needed the #25 manifest-convergence sync passes (not part of this change).

Closes #7 (items 3 + 5 — the remainder).

### DIFF
--- a/featuregen/src/cppMain/include/strings_feature.h
+++ b/featuregen/src/cppMain/include/strings_feature.h
@@ -371,6 +371,27 @@ struct ShapeProbe {
     }
 };
 
+// CV-getat-base-elem (#7 item 3): a `get`/`at`-keyed index container whose ELEMENT
+// type is a BASE class (`Shape*`, which has a generated `ShapeApi` interface). The
+// generator upcasts the `at(...)` return to `ShapeApi?`, but the synthesized
+// `iterator()`'s `next()` declares the CONCRETE element type — so WITHOUT the item-3
+// fix the emitted `next(): Shape? = at(i): ShapeApi?` doesn't typecheck and the
+// featuregen build fails to compile. With the fix the `at`/`get` get keeps the
+// concrete element type (matching `operator[]`), so the container binds and the stored
+// concrete Shapes round-trip (area() virtual-dispatches through the element).
+//
+// Index keyed by `at`/`get` (NOT `operator[]`) with a `size_t` size + `size_t` index,
+// so detectIndexIterable selects the `at` fallback rather than the IND operator path.
+struct ShapeBag {
+    Shape* items[3];
+    std::size_t n;
+    ShapeBag() : n(0) { items[0] = items[1] = items[2] = nullptr; }
+    void add(Shape* s) { if (n < 3) items[n++] = s; }
+    std::size_t size() const { return n; }
+    Shape* at(std::size_t i) const { return i < n ? items[i] : nullptr; }
+    Shape* get(std::size_t i) const { return at(i); }
+};
+
 // IH-flatten (3-level chain): Animal <- Dog <- Puppy. `Animal` is polymorphic
 // (virtual speak()/legs() + virtual dtor); `Dog` overrides speak(); `Puppy`
 // overrides speak() AGAIN and inherits legs() unchanged. This exercises the
--- a/featuregen/src/nativeTest/kotlin/CvGetAtBaseElemTest.kt
+++ b/featuregen/src/nativeTest/kotlin/CvGetAtBaseElemTest.kt
@@ -1,0 +1,53 @@
+import kotlinx.cinterop.memScoped
+import root.Circle
+import root.Rectangle
+import root.ShapeBag
+import kotlin.math.PI
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+// CV-getat-base-elem (#7 item 3): ShapeBag is a `get`/`at`-keyed index container
+// (size_t size + size_t index, no operator[]) whose ELEMENT type is the BASE class
+// `Shape*`. WITHOUT the item-3 fix, the generator upcast the `at`/`get` return to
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
**Approved — tier 1 (feature_internal / bug_fix)**

**Item 3 (get/at concrete element):** Correctness verified. `IndexIterable` gains `indexGet: ResolvedMethod`; `isIndexGetMethod` uses Kotlin referential identity (`===`) to match exactly the one method `detectIndexIterable` selected — not any method named `get`/`at`. The `keepConcreteElement` flag threads cleanly through `generateBasicMethod`; defaults false, so only the one selected index-get is affected and no existing golden output changed. The `operator==ResolvedOperator.IND` path is untouched. With both `at` and `get` in `ShapeBag`, declaration order selects `at` as the index-get, so `at` stays concrete and `get` stays base-typed — correct and consistent with what the PR description states.

**Item 5 (RangeReturn drop visibility):** The `DropLedger.record` call is purely additive — `return null` is unchanged, the drop decision is unchanged, and `DropLedger.record` is non-suspend (plain `fun`, no coroutine context needed). Resolves the earlier `Log.w`-suspend blocker cleanly.

**Tests:** Both tests bite. `CvGetAtBaseElemTest` is a compile-time witness to the fix (the module wouldn't compile without it) plus behavioral asserts on virtual dispatch. `unextractableIteratorRangeRecordsDrop` directly asserts the PARSE ledger record and confirms the sibling `count` method still binds.

**Minor note (non-blocking):** `isIndexGetMethod` calls `detectIndexIterable` once per non-operator method on its class, making the detection run N+2 times instead of 2. Acceptable for real C++ class sizes; worth a caching pass if hot paths appear later.

PR #32, opened 2026-06-09
Cold-start: converge the SYNC_REQUIRED manifest deterministically in one invocation (#25)

### LINKED ISSUE #25: kplusplusSync cold-start: nested-generic manifest needs multiple passes to converge (fresh worktree friction)

monkopedia-coder:
## Problem (#25)
On a fresh checkout/worktree the featuregen SYNC_REQUIRED manifest (`krapped/requested.txt`) is discovered **lazily over compile passes**. The FIR plugin only learns it needs a *nested* generic (e.g. `std::vector<std::vector<int>>`) once the *inner* binding (`std::vector<int>`) has been generated and compiled against — `CppVectorMapping.elementsOf` reads the inner type's `@CppBinding` spec, which requires that inner binding on the classpath. So a cold `krapped/` needs several `sync → compile → sync` cycles to reach the full closure, and the first run looks like a failure. Distinct from #16 (staleness flake); this is the **cold-start convergence** gap.

### Reproduced on this base (cold `krapped/`), 3 invocations needed:
| invocation | `requested.txt` | result |
|---|---|---|
| 1 | 0 → 10 specs | FAIL `kplusplus-sync-required: std::vector<int>` |
| 2 | 10 → 11 specs | FAIL `kplusplus-sync-required: std::vector<std::vector<int>>` |
| 3 | 11 (stable) | BUILD SUCCESSFUL |

(11 discovered specs + 6 build-script `instantiate(...)` seeds = the 17-instantiation fixpoint.)

## Convergence mechanism
Pin the converged closure as a **committed, version-controlled input**. `krapped/requested.txt` is un-ignored (the rest of `krapped/` — generated `src/`, the `.def`, `generated.txt`, `fixups.json` — stays machine-generated and ignored). A cold checkout now starts with the **whole** SYNC_REQUIRED set, so the first `kplusplusSync` generates everything up front and a single `:featuregen:nativeTest` converges in **one pass** — no manual re-sync, no multi-pass friction.

This rides entirely on the existing sync wiring (the task already unions `requested.txt` with the build-script seeds and regenerates the full set each run) — **no Kotlin or build-logic changes**, just the manifest data + `.gitignore`. It also tightens **bootstrap determinism**: the instantiation set is now a fixed, reproducible input rather than something rediscovered over a variable number of passes. The FIR plugin still appends newly-discovered specs, so adding a test with a new instantiation self-heals the manifest — commit the updated file to keep cold starts single-pass (documented inline in `.gitignore`).

## Cold-start proof (after)
Simulated a fresh checkout (`git clean -fdx featuregen/krapped` → `krapped/` holds only the committed manifest), then **one** invocation:
```
$ JAVA_HOME=/usr/lib/jvm/java-17-openjdk ./gradlew :featuregen:nativeTest
BUILD SUCCESSFUL
```
- `requested.txt` unchanged afterward (11 specs, **no growth** = fully converged in one pass)
- `git diff` on the manifest empty (append found every spec already present)

## Verification
- `:featuregen:nativeTest` — **188 / 0** (0 failures, 0 errors, 0 skipped), single cold invocation
- `:krapper_gen:nativeTest` not re-run: this change touches no krapper_gen code or build logic (only `featuregen/.gitignore` + a featuregen data file), so it cannot affect that suite.

closes #25


monkopedia-reviewer (approved):
**Approved — tier 1 auto-merge.**

### Correctness

**1. `.gitignore` change is correct.**
The switch from `/krapped/` to `/krapped/*` is the load-bearing move. The old rule ignored the `krapped/` directory itself, which causes gitignore to stop recursing — making any `!` negation inside it a no-op. `/krapped/*` instead ignores each item *inside* krapped/ as individual entries while leaving the directory visible, so `!/krapped/requested.txt` correctly un-ignores exactly that file. Verified: `src/`, `generated.txt`, `fixups.json`, `.def`, `.a`, `.cc`, `.h` all remain matched by `/krapped/*` and stay gitignored.

**2. Committed manifest is the correct, complete converged closure.**
11 entries, all valid C++ template specs, no absolute paths, no platform-specific content. These are the lazily-discovered specs; the 6 build-script seeds (`std::unique_ptr<int>`, `std::vector<Thing*>`, `Box<int>`, `Box<Point>`, `Box<Box<int>>`, `Pair2<int, double>`) continue to come from `instantiate()` calls in `build.gradle.kts` and are not needed in the file. The nested cases (`std::vector<std::vector<int>>`, `std::vector<std::__cxx11::basic_string<char>>`) are present — these are exactly the lazily-discovered closure that required 3 cold passes before.

**3. Determinism holds.**
`KPlusPlusCompilerGradlePlugin.kt:149` unions `manifestSpecs + seededInstantiations`, dedupes, and **sorts alphabetically** before invoking krapper_gen — so the krapper_gen run order is deterministic regardless of append order in `requested.txt`. On a cold checkout: 11-entry manifest + 6 seeds = 17 specs, all sorted, all generated in the first kplusplusSync; FIR compile sees all bindings on classpath → no SYNC_REQUIRED fires → single-pass convergence.

**4. Committed-lockfile approach is correct for the bootstrap goal.**
Pinning the instantiation closure as a version-controlled input is exactly right for the self-bootstrap determinism requirement (#25 / broader strategy). The maintenance discipline is clearly documented inline in `.gitignore` (append-on-next-build, commit the updated file). No silent-stale risk: the FIR plugin self-heals by appending, and `git status` surfaces the change.

### Quality
Inline comment in `.gitignore` is thorough — problem statement, mechanism, and maintenance discipline all captured. No Kotlin or build-logic changes means krapper_gen test suite exclusion is justified.

PR #34, opened 2026-06-09
Hermeticity: thread process-global state through context (#11 part 2)

monkopedia-coder:
## Hermeticity: thread process-global state through a context (#11 part 2)

Part 2 of #11: get the two process-global mutable top-levels off being loose,
leak-prone state that makes runs order-sensitive and blocks in-process test
isolation / Gradle-daemon reuse.

### What moved + how
Both globals are now gathered behind a single resettable `GenerationContext` object:
- `existingTypes` — the `WrappedType` intern/dedup cache (was a top-level `val` in `WrappedType.kt`) → `GenerationContext.internedTypes`.
- `rootPackageOverride` — was a top-level `var` in `WrappedKotlinType.kt` → `GenerationContext.rootPackage` (private setter, routed through `reset()`).

`IndexedServiceImpl.init` calls `GenerationContext.reset(config.rootPackage)` right next to the existing `DropLedger.reset()`. The intern cache is now **cleared per run**, so sequential in-process runs no longer leak it into one another — the order-sensitivity called out in `docs/clang-bootstrap.md`.

### Why this shape (existing carrier, not a threaded context param)
The reality-check flagged threading a context through `WrappedType.invoke` as a potential rabbit hole. It is — and it's the *same* rabbit hole for **both** globals: each is read from the context-less `WrappedType.invoke` / `WrappedType.kotlinType` / `fullyQualifiedType` surfaces, which are called pervasively (dozens of sites) with no carrier to thread. `rootPackageOverride` is read inside `fullyQualifiedType`, itself called from the context-less `.kotlinType` getter, so it is **not** more threadable than `existingTypes`.

So rather than a sprawling signature change for no behavioral gain, this uses the codebase's own established idiom for exactly this problem: the sibling `DropLedger` / `Log` objects, which the `DropLedger` KDoc documents as "the generator runs one request at a time under `runBlocking`, so a shared collector is the cohesive carrier without threading a handle through every signature." `GenerationContext` is that carrier for these two.

### What's left / deferred
Concurrent in-process runs (two generations in one process at the same time) still need true isolation (thread-local or a genuinely threaded context). That's a larger design change deferred from this PR; noted inline in `docs/clang-bootstrap.md`. The realistic scenarios today — separate-process binary invocations and sequential in-process tests — are now hermetic.

### Behavior-preserving (byte-identical)
Captured an md5 manifest of the generated `featuregen/krapped/src` `.kt` on clean `00278f7`, regenerated after the change, and diffed: **119 generated `.kt` files byte-for-byte identical**.

### Verification
- `:krapper_gen:nativeTest` — **305 / 0**
- `:featuregen:nativeTest` — **188 / 0**

Refs #11 (part 2 of N) — does not close it.


monkopedia-reviewer (approved):
Approved - refactor, tier 1.

Reviewed against the five scrutiny points:

1. Per-run reset correctness: GenerationContext.reset in IndexedServiceImpl.init calls both internedTypes.clear() and rootPackage assignment atomically, sitting beside DropLedger.reset(). Fires before any resolution - no cross-run leakage.

2. Behavior-preserving cache clear: internedTypes is a within-run dedup, not cross-run persistent state. Clearing between runs forces a per-run rebuild with identical effect. 119-file byte-identical proof confirms no output change.

3. rootPackage migration complete: every rootPackageOverride read site (fullyQualifiedType in WrappedKotlinType.kt, free-functions container in KotlinWriter.kt) updated to GenerationContext.rootPackage. Private setter enforces that only reset() can write it.

4. DropLedger/Log idiom is correct here: threading a context param through WrappedType.invoke/kotlinType/fullyQualifiedType would be dozens of call-site changes for zero behavioral gain under the one-request-at-a-time runBlocking model. The resettable-object carrier is the established idiom. Concurrent in-process case is properly documented inline and in docs/clang-bootstrap.md as a deferred follow-up.

5. RootPackageTest: all rootPackageOverride assignments correctly become GenerationContext.reset. The AfterTest reset calls GenerationContext.reset(null) which resets both fields - intern cache cleared AND rootPackage nulled, not just the old bare null assignment. Tests are non-tautological.

No issues found. Clean refactor.

Issue #10, opened 2026-06-08
[clang-campaign] SCALE tail: broad-binding correctness (the 284-error / bare-enum-qualification thesis)


monkopedia-coder:
**Campaign SCALE tail — broad-binding correctness ("massive binding" thesis).** The demo runs with a 16-class slice, but the campaign's stated product goal is binding LARGE libraries. Forcing the full transitive Clang surface (~1269 classes via INCLUDE_MISSING) currently yields ~284 generated-C++ compile errors (measured 2026-06-06, recorded in `docs/clang-bootstrap.md` Backlog as the T1.3-clang/"harden broad binding" P2 tail). It is **dominated by 179× "use of undeclared identifier"** — the doc notes this is "likely ONE systematic bug (bare enum constants / static members emitted unqualified)." That single fix is the highest-leverage scale unlock.

**Do:** root-cause + fix the bare-enum/static-member qualification bug first (biggest cluster), then iterate the remaining categories (incomplete-type, deleted-ctor residuals, fn-ptr, etc.) via the rebuild-kexe + real-Clang re-probe loop documented in the doc. Large + iterative. Acceptance: the forced broad surface compiles (error count → low single digits), proving the binding pipeline scales. NOTE: this is the bulk of "make it releaseable as a scale story" — size accordingly.

monkopedia-coder:
## #10 re-investigation (2026-07 fresh, current `main` @ c20c97d, LLVM 22.1.6): the broad force RUNS — no crash/empty-output/SEGV — but is CPU-bound for tens of minutes in the forcing resolve. Fresh error count NOT obtainable in-env; root cause below.

Re-ran the broad-surface harness on current `main` in a dedicated worktree (`-PenableClang`, `Xmx12g`). Key updated facts that make the **old 284/179 "use of undeclared identifier" thesis STALE** and re-frame the measurement blocker:

### 1. krapper_gen no longer self-parses — the force is now a 2-stage `krapper_parse` → `krapper_gen` pipeline
[...]

PR #35, opened 2026-06-09
Fix INCLUDE_MISSING cycle guard: key the in-progress marker on the leaf, not the wrapped type (#10 step 1)

monkopedia-coder:
## What

Corrects a mis-keyed cycle guard in the `INCLUDE_MISSING` expansion mapper (`krapper_gen/.../Resolver.kt`).

The mapper marks an in-progress leaf in `ResolveTracker.otherResolved` while it expands a missing referenced type. The re-entry guard `canResolve` follows a wrapped type down to its **bare leaf** and queries that set by the leaf string. But the marker was set/cleared on the **outer wrapped carrier** `t` (e.g. `const clang::X &`), not the leaf `it` (`clang::X`):

```
otherResolved.add(t.toString())     // "const clang::X &"
...
otherResolved.remove(t.toString())
```

For any class reached **only** through a wrapped carrier, the marker therefore never matched the leaf the guard checks — it was effectively dead. A base/member edge re-entering the in-flight class was not deduped at the marker level, so the class got re-resolved redundantly.

## Fix (2 lines + a defensive guard)

- Key the marker on the leaf `it` on both add and remove (matching what `canResolve` queries).
- Add a defensive short-circuit: if the leaf is already in progress, take the cycle exit (`ElementUnchanged`) instead of recursing.

## Test

Adds `ParseTest.testIncludeMissingCyclicReferenceTerminates`: a self-referential `const Node&` class reached only via a wrapped carrier from a seed class, under `INCLUDE_MISSING`. Asserts the cyclic path **terminates and binds**.

## Honest scope note (please read)

The root-cause hand-off predicted that a minimal mutually-referential pair would **SIGSEGV** without this fix. I could **not** reproduce that at unit scale. I tested four distinct minimal cyclic fixtures — pointer pair, const-ref pair, inheritance cycle, and self-reference — and **all terminate even with the buggy keying**. The `resolvedClasses` cache + `mappingCache` + the bare own-type marker (which is correctly keyed by coincidence, since `t == it` for a bare type) bound the redundant work to a constant factor in small graphs. The native stack overflow that blocks forcing the broad Clang surface is an **emergent property of its dense ~1269-class graph**, not reproducible in a small fixture.

So this PR is framed as **the correct cycle-guard keying + dedup, with regression coverage of the cyclic `INCLUDE_MISSING` path** — not as a fixture that crashes the old code on its own. A full broad-surface run (seed the `clang::Decl` hierarchy under `INCLUDE_MISSING` and confirm resolution terminates + capture the real C++ error tail, replacing the stale 284) is the right follow-up verification; I was unable to complete it in this turn because the broad clangwalk config needs the `only(...)` Decl seed kept (removing it empties the seed) and an environment hiccup cut the run short.

## Verification

- `:krapper_gen:nativeTest` — **306/0** (305 + the new test)
- `:featuregen:nativeTest` — **188/0**; `krapped/src` **byte-identical** (working tree shows only the 2 changed source files — normal generation unchanged)

## Note for reviewers

A sibling branch `fix-include-missing-cycle-10b` exists (parallel/retry of this same task). This PR (`fix-include-missing-cycle-10`) is the one to review; please de-dup if the other surfaces.

Refs #10

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
@@ -561,7 +561,20 @@ private fun typeMapper(policy: ReferencePolicy): TypeMapping {
                         if (context.tracker.resolvedClasses[it.toString()] != null) {
                             return@operateOn RemoveElement
                         }
-                        context.tracker.otherResolved.add(t.toString())
+                        // Cycle guard: an outer frame may already be expanding this very
+                        // leaf (Clang's AST is densely cyclic — mutual Decl refs, CRTP
+                        // bases, pointer/ref members back into the hierarchy). If so, don't
+                        // recurse; the outer frame will store it. Key on the leaf `it` —
+                        // exactly what canResolve queries — NOT the wrapped carrier `t`
+                        // (e.g. `const clang::X &`). When a class is reached only through a
+                        // wrapped carrier, a `t`-keyed marker never matches the leaf the
+                        // guard checks, so a base/member edge re-entering the in-flight
+                        // class isn't deduped and resolution descends until the native
+                        // stack overflows.
+                        if (context.tracker.otherResolved.contains(it.toString())) {
+                            return@operateOn ElementUnchanged
+                        }
+                        context.tracker.otherResolved.add(it.toString())
                         try {
                             // Reference expansion: keep the legacy hard-fail when a
                             // pulled class's primary base can't resolve (don't
@@ -579,7 +592,7 @@ private fun typeMapper(policy: ReferencePolicy): TypeMapping {
                             context.tracker.resolvedClasses[resolved.type.toString()] = resolved
                             context.tracker.classes[wrapper.type.toString()] = wrapper
                         } finally {
-                            context.tracker.otherResolved.remove(t.toString())
+                            context.tracker.otherResolved.remove(it.toString())
                         }
                     } catch (original: Throwable) {
                         try {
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
**Approved. tier 1 / bug_fix / auto-merge.**

The keying fix is correct.

operateOn recurses through all wrapper layers (WrappedModifiedType, isReference, isPointer) before invoking the lambda — so inside t.operateOn { ... }, 'it' is the bare leaf (e.g. clang::X) while 't' is the outer carrier (e.g. const clang::X &). canResolve queries otherResolved by bare-leaf string (the public path falls through to type.unconst.name, the private path does otherResolved.contains(str)). The old t.toString() marker key was therefore dead for any type reachable only through a wrapped carrier: marker keyed on "const clang::X &" while guard queried "clang::X". For bare types t == it, so the fix has zero effect on non-wrapped paths — consistent with the byte-identical featuregen output.

The defensive short-circuit before the add is belt-and-suspenders: uses the same it.toString() key, returns ElementUnchanged (leave reference in place, outer frame stores the class), matching the mapper's existing return idiom. After the fix the outer !canResolve(it) check would already short-circuit re-entry (since otherResolved now contains the leaf), but the explicit guard is a clean safety net.

Test testIncludeMissingCyclicReferenceTerminates is valid as regression/smoke coverage of the cyclic INCLUDE_MISSING path. The honest scope note (doesn't crash the old code at unit scale; the native stack overflow is an emergent property of the dense ~1269-class graph) is correctly documented and is not a basis for requesting changes — the fix is right, the test asserts termination-and-bind of the now-correctly-guarded path, and the broad-surface follow-up is appropriately deferred.

306/0 krapper_gen, 188/0 featuregen, ktlint green.

PR #36, opened 2026-06-09
Resolver: skip-not-crash on unmodelable callback params (self-bootstrap GAP A)

monkopedia-coder:
### Root cause
Under `referencePolicy = IGNORE_MISSING` (the `clangwalk/` self-host harness), a method whose parameter is an unmodelable callback — `llvm::function_ref<R(Args)>` / `std::function<R(Args)>` — was **not** skip-not-crashed.

The callback's template argument is a **function type** (`int (Thing *)`) the type model can't bind, so the param fails to resolve. But the `hasDefault` cursor heuristic misreads the callback signature's parenthesized parameter list as a default value (the param's "default" surfaces as e.g. `Thing*`). So a genuinely-**required** param looked `isOmittableDefault` and was silently **trimmed** from the signature, while the emitted C++ glue still called the original method with its full arity:

```
withCallback(a)   // 2-arg method called with 1 arg → "too few arguments"
```

That arity mismatch fails to compile and aborts the whole build (the clangwalk stage-1 exit-134 on `clang::ASTContext::adjustType` and `forEachMultiversionedFunctionVersion`, previously papered over with `RemoveMethodByCName` fixups).

### Fix
`WrappedArgument.typeCarriesFalseDefault` already classifies array-bound and non-type-template-arg tokens as *false* defaults (a misread of a type token as a value). Extend it to also recognize a **function-type template argument** — structural signal: its spelling carries a `(`; plain / pointer / reference / template type spellings never do. A function-type arg is never a genuine, trimmable default, so the param is no longer `isOmittableDefault` and the **existing** param-resolution path drops the **whole method** (skip-not-crash, recorded in the `DropLedger`) — exactly like an unresolvable return type.

The legitimate omittable-default path is preserved: a genuinely-defaulted trailing param over a *type* arg (e.g. `buildASTFromCode`'s defaulted params, `shared_ptr<X>`) still trims and binds. `INCLUDE_MISSING` / featuregen output is unchanged.

### Test
`ParseTest.testUnmodelableCallbackParamDropsWholeMethod` — a self-contained `FnRef<int(Thing*)>` mimic (partial specialization over a function type) under `IGNORE_MISSING` asserts:
- the callback-param method is **dropped** and **ledgered** (no arity-mismatched short call),
- the plain sibling method binds,
- a genuinely-defaulted sibling (`Holder<Thing> s = Holder<Thing>()`) **still** binds (legitimate omit path intact).

### Verify (JDK 17, clean base off 6152cf5)
`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` — **BUILD SUCCESSFUL**
- krapper_gen **307/0** (baseline + the new test)
- featuregen **188/0**, `kplusplusSync` byte-identical (no git delta)
- ktlintCheck clean

Note: the `clangwalk/fixups.json` `removeMethod` workarounds for `adjust_type` / `for_each_multiversioned_function_version` become unnecessary after this fix, but clangwalk is on a separate branch and was left untouched (not verified end-to-end in this PR's budget).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/WrappedMethod.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/WrappedMethod.kt
@@ -623,9 +623,32 @@ class WrappedArgument(
             // Secondary path: when the value instead survives as a literal STRING token
             // (the spelling-reconstruction path can yield `Mask<4>` with arg `"4"`),
             // [isNonTypeTemplateArg] classifies that token. Either path catches it.
+            //
+            // Third path — an unmodelable CALLBACK param (`llvm::function_ref<R(Args)>`,
+            // `std::function<R(Args)>`): the template arg is a FUNCTION TYPE (`int (Thing
+            // *)`), which the type model can't bind, so the param fails to resolve. But the
+            // hasDefault cursor heuristic misreads the callback signature's parenthesized
+            // param list as a default value (e.g. `cb`'s "default" surfaces as `Thing*`), so
+            // a callback param with NO real C++ default looks omittable. Trimming it then
+            // emits a SHORT call (`m(a)` for a 2-arg method) — an arity mismatch that fails
+            // to compile and aborts the whole build (the clangwalk self-host hit this on
+            // `clang::ASTContext::adjustType` / `forEachMultiversionedFunctionVersion`). A
+            // function-type arg is never a genuine, trimmable default, so treat it as a false
+            // default: the whole method drops (skip-not-crash, ledgered) instead.
             return referent is WrappedTemplateType &&
-                referent.templateArgs.any { it == UNRESOLVABLE || it.isNonTypeTemplateArg }
+                referent.templateArgs.any {
+                    it == UNRESOLVABLE || it.isNonTypeTemplateArg || it.isFunctionTypeArg
+                }
         }
+
+    // A template-argument WrappedType that is a FUNCTION TYPE — the `R(Args...)` callback
+    // signature inside `function_ref<R(Args...)>` / `std::function<R(Args...)>`. It surfaces
+    // as a bare type spelling carrying a parenthesized parameter list (`int (Thing *)`);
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
**Tier 1 — approve + merge**

**Correctness**

The `typeCarriesFalseDefault` extension correctly identifies a function-type template argument (structural signal: its spelling carries `(`) as a false default, preventing the arity-mismatched short-call codegen path from firing. The reasoning is sound: plain / pointer / reference / template type spellings never contain `(`, so the character is a clean discriminator for function-type args.

The test covers all three required cases:
- `withCallback` drops whole-method (skip-not-crash, ledgered with a non-blank reason) ✓
- `normal` (plain sibling) still binds ✓
- `withDefaulted` (genuine `Holder<Thing> s = Holder<Thing>()` default over a TYPE arg) still trims and binds — proving the legitimate omittable-default path is intact ✓

INCLUDE_MISSING / featuregen output is unchanged (kplusplusSync byte-identical).

**Quality**

`isFunctionTypeArg` follows the exact same shape as the existing `isNonTypeTemplateArg` extension property just below it — consistent, idiomatic, concise. The "Third path" comment in `typeCarriesFalseDefault` extends the existing numbered annotation style cleanly. Test follows established file conventions (memScoped, runBlocking, DropLedger.reset, random() temp-file names).

**Non-blocking note**

`WrappedTypedefRef.toString()` returns `"unresolved_typedef($usr)"` which also contains `(`. If such a type happened to be a template arg of an unresolvable param with a genuine C++ default, the method would now drop instead of trimming. This is a conservative-direction edge case (drop, not wrong compilation), affects only IGNORE_MISSING, and is very unlikely in practice — not blocking. Worth recording for future tightening.

**Build (independent verification, JDK 17)**

`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` → BUILD SUCCESSFUL (6m): nativeTest executed (all green), kplusplusSync regenerated 17 instantiations byte-identical, nativeTest UP-TO-DATE, ktlintCheck clean.

PR #37, opened 2026-06-09
krapper_gen: bind by-value std::string returns under IGNORE_MISSING (self-bootstrap GAP B)

monkopedia-coder:
## Self-bootstrap GAP B: by-value `std::string` returns dropped under scoped IGNORE_MISSING

A method returning `std::string` **by value** — e.g. `clang::QualType::getAsString()` and `clang::NamedDecl::getNameAsString()`, the type-string + name extraction at the crux of the clangwalk self-host front-end — was **silently DROPPED** under a scoped (`only(...)`) import + `IGNORE_MISSING` (the `clangwalk/` harness config). This is the wall for the front-end campaign.

### Confirmed root cause
libclang spells `std::string`'s canonical type as the libstdc++ ABI-tagged template **`std::__cxx11::basic_string<char>`** (NOT the literal `std::string`). Empirically confirmed via the scoped repro: the return surfaces as `std::__cxx11::basic_string<char>` (a `WrappedTemplateType`), `getNameAsString` drops, and a sibling `int`-returning method survives.

Under a scoped import whose allowlist doesn't bind `basic_string`, that template matches neither the STRING fast-path (`isString == name == "std::string"`) nor any tracked class, so a missing-tolerant policy treats it as a missing dependency and drops the whole method. (This is also why the brick-1 in-build-script workaround of adding `"std::string"` to `only()` failed — name mismatch.)

### Fix
In `typeMapper` (`Resolver.kt`), for the missing-tolerant policies (IGNORE/OPAQUE/THROW), normalize an **unbound** canonical `basic_string<char,...>` to the always-available `std::string` *before* the policy decision, so it rides the EXISTING `STRING -> const char* -> Kotlin String` marshalling (the same destination `rewriteViewReturns` routes `llvm::StringRef` to). No marshalling logic is duplicated.

- **Gated on the `basic_string` class being unresolvable**, so `INCLUDE_MISSING` (featuregen) — which binds `basic_string` as the wrapped `Basic_string__Char` class — is left untouched and its emitted output stays **byte-identical**.
- **char-only**: `wstring`/`u16string`/`u32string` (`basic_string<wchar_t>`/`<char16_t>`/`<char32_t>`) are not `char*` strings and are left alone.
- Reuses `operateOn` as a non-removing recursive walk, so it reaches the `basic_string` node even for the multi-arg `<char, char_traits<char>, allocator<char>>` spelling, and re-wraps through by-value / `const` / `const&` carriers.
- A small wrapper surfaces the normalized type as a `ReplaceWith` (the mapper otherwise returns `ElementUnchanged`, which `map()` reads as "keep the original" and would discard the rewrite).

### Test
`ReturnShapeTest.scopedStringByValueReturnBindsAsString` locks the `IGNORE_MISSING` + `only()` path: a by-value and a top-level-`const` by-value `std::string` return must SURVIVE resolution and surface as Kotlin `String`.

### Verify (single foreground run, JDK 17)
`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck`
- krapper_gen nativeTest: **308 / 0**
- featuregen kplusplusSync: **byte-identical** (no `krapped/` drift)
- featuregen nativeTest: **188 / 0**
- krapper_gen ktlintCheck: clean

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
@@ -522,8 +522,8 @@ data class ResolveContext(
 }
 
 private fun typeMapper(policy: ReferencePolicy): TypeMapping {
-    return when (policy) {
-        ReferencePolicy.IGNORE_MISSING -> return { t, context ->
+    val mapper: TypeMapping = when (policy) {
+        ReferencePolicy.IGNORE_MISSING -> { t, context ->
             t.operateOn {
                 if (context.tracker.canResolve(it, context)) {
                     ElementUnchanged
@@ -533,7 +533,7 @@ private fun typeMapper(policy: ReferencePolicy): TypeMapping {
             }
         }
 
-        ReferencePolicy.OPAQUE_MISSING -> return { t, context ->
+        ReferencePolicy.OPAQUE_MISSING -> { t, context ->
             t.operateOn {
                 if (context.tracker.canResolve(it, context)) {
                     ElementUnchanged
@@ -543,7 +543,7 @@ private fun typeMapper(policy: ReferencePolicy): TypeMapping {
             }
         }
 
-        ReferencePolicy.THROW_MISSING -> return { t, context ->
+        ReferencePolicy.THROW_MISSING -> { t, context ->
             t.operateOn {
                 if (context.tracker.canResolve(it, context)) {
                     ElementUnchanged
@@ -553,7 +553,7 @@ private fun typeMapper(policy: ReferencePolicy): TypeMapping {
             }
         }
 
-        ReferencePolicy.INCLUDE_MISSING -> return { t, context ->
+        ReferencePolicy.INCLUDE_MISSING -> { t, context ->
             t.operateOn {
                 if (it.isArray) return@operateOn RemoveElement
                 if (!context.tracker.canResolve(it, context)) {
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed to main-merge standard. APPROVE.

**Classification:** `bug_fix` — targeted fix for std::string-by-value methods silently dropped under IGNORE_MISSING + scoped import (self-bootstrap GAP B).  
**Tier:** 1 (kplusplus baseline `* : tier 1`).

---

### Correctness — all checks clear

**char-only gate** (highest-risk): `leaf.templateArgs.firstOrNull()?.toString() == "char"` correctly discriminates. `wchar_t`/`char16_t`/`char32_t` first-args don't match → wstring/u16string/u32string are left alone. ✅

**INCLUDE_MISSING gate**: `if (policy == ReferencePolicy.INCLUDE_MISSING) return mapper` is a direct early return before the wrapper — featuregen's path is completely untouched. ✅

**canResolve gate**: normalization only fires when the basic_string class is UNBOUND in the context. ✅

**ReplaceWith fix**: when normalization rewrote `t` but the inner mapper returns `ElementUnchanged`, the outer wrapper correctly surfaces `ReplaceWith(normalized)` so the rewrite isn't discarded by `map()`. Traced the logic; necessary and correct. ✅

**Multi-arg form**: `handleTemplate` processes args first then calls `typeHandler` on the full `WrappedTemplateType` — the first-arg check captures `basic_string<char, char_traits<char>, allocator<char>>` correctly. ✅

**const carrier**: `WrappedModifiedType` dispatches to `baseType.operateOn()` + `wrapOnReplace` — `const std::__cxx11::basic_string<char>` → `const std::string` correctly. ✅

**double-return cleanup**: refactoring `return when { ... -> return { ... } }` to `val mapper = when { ... -> { ... } }` is idiomatic and necessary for the wrapper pattern. ✅

### Quality

Code reads consistently with surrounding style. KDoc, CHAR_BASIC_STRING_BASES constant, and the inline `ReplaceWith`/`ElementUnchanged` comment are appropriate for the complexity. No style ledger violations.

### Build (independent, JDK 17)

- `krapper_gen:nativeTest`: 307/0, `scopedStringByValueReturnBindsAsString` present and passing (2.001s)  
- `featuregen:kplusplusSync`: `git status` shows zero drift — byte-identical ✅  
- `featuregen:nativeTest`: 188/0  
- `krapper_gen:ktlintCheck`: clean

PR #38, opened 2026-06-09
krapper_gen: don't normalize a pointer-carried basic_string<char> (GAP B follow-up)

monkopedia-coder:
## What
GAP B (#37) normalizes an unbound canonical `std::__cxx11::basic_string<char>` to `std::string` so it rides the STRING -> `const char*` boundary. But `normalizeMissingCharString` descended through **every** carrier — including a **pointer**. A `std::string*` out-param (e.g. `clang::Decl::getAvailability(std::string*)` / `isDeprecated(std::string*)`) therefore got normalized, and since the STRING arg cast-var is declared with the pointer-typed `targetType`, the wrapper emitted:

```cpp
std::string* x = std::string(arg);   // value assigned to a pointer — won't compile
```

…which aborts the **entire** generated-wrapper build. This was the wall for the clangwalk self-host harness once the `QualType`/`ASTContext` surface is bound.

## Fix
`normalizeMissingCharString` now descends only carriers the `const char*` boundary can actually marshal (by-value, `const`, reference, and value-position template args) and **hard-stops at a pointer/array carrier** — leaving the unbound `basic_string` in place so the arg/method drops cleanly (the pre-GAP-B behavior) instead of emitting broken codegen. By-value and `const std::string&` still normalize (their cast-var is value-typed — an lvalue that binds fine).

## Safety
The mapper short-circuits before `normalizeMissingCharString` for `INCLUDE_MISSING` (featuregen's policy), so **featuregen output stays byte-identical**; the rewrite only ever ran under the missing-tolerant policies.

## Test
`ReturnShapeTest.scopedStringPointerParamDropsNotCrashes` locks it: under `IGNORE_MISSING` + `only(...)`, a `std::string*` out-param method DROPS while a by-value `std::string` return and a `const std::string&` in-param on the same class survive.

## Verify (JDK 17)
- `:krapper_gen:nativeTest` green
- `:krapper_gen:ktlintCheck` clean
- `:featuregen:kplusplusSync` byte-identical + `:featuregen:nativeTest` green

Proven end-to-end: the clangwalk stage1 walk now binds `clang::NamedDecl::getNameAsString()` and `clang::QualType::getAsString()` and prints real names + type spellings (`freeFunction : void (int)`, `globalVar : int`).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
@@ -649,28 +649,44 @@ private val CHAR_BASIC_STRING_BASES = setOf("std::basic_string", "std::__cxx11::
  * `wstring`/`u16string`/`u32string` (`basic_string<wchar_t>`/`<char16_t>`/`<char32_t>`) are
  * NOT `char*` strings, so they are left untouched and follow the normal policy behavior.
  *
- * Reuses [operateOn] purely as a recursive walk: the handler never removes, so the rewrite
- * reaches the basic_string node even for the multi-arg `<char, char_traits<char>,
- * allocator<char>>` spelling (whose inner trait/allocator args would otherwise be the first
- * to drop), and re-wraps it through any by-value / `const` / `const&` carrier. Gated on the
- * basic_string class being UNRESOLVABLE: when it IS bound (an explicit allowlist entry) the
- * type is left as the wrapped class.
+ * Descends only carriers the STRING boundary can actually marshal: by-value, `const`, and
+ * reference (`&`), re-wrapping the basic_string node through them. A POINTER (`*`) or array
+ * (`[]`) carrier is a HARD STOP — a `std::string*` out-param (e.g. `clang::Decl::getAvailability(
+ * std::string*)`) can't ride the `const char*` boundary: the cast-var is declared with the
+ * pointer-typed `targetType`, so normalizing it would emit `std::string* x = std::string(arg)`,
+ * which does not compile and aborts the whole wrapper build. Leaving such a type unbound makes
+ * the arg/method DROP cleanly (the pre-GAP-B behavior) instead of producing broken codegen.
+ * Matches the basic_string node even for the multi-arg `<char, char_traits<char>,
+ * allocator<char>>` spelling. Gated on the basic_string class being UNRESOLVABLE: when it IS
+ * bound (an explicit allowlist entry) the type is left as the wrapped class.
  */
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
**Tier 1 — bug_fix — APPROVE**

Reviewed the full diff to merge standard. Both axes clean.

## Correctness (axis A)

**Carrier guard completeness:** The `normalizeMissingCharString` rewrite now correctly hard-stops at `WrappedModifiedType` with `modifier == "*"` or `"[]"`. This is the right layer and the right type: a pointer/array carrier produces a pointer-typed cast-var, so normalizing would emit `std::string* x = std::string(arg)` — value-to-pointer, won't compile. Hard-stopping leaves the unresolved `basic_string<char>` in place, which the downstream mapper drops cleanly under IGNORE/OPAQUE/THROW (pre-GAP-B behavior).

**`std::string&` (non-const lvalue ref) gap — not a gap:** `std::string&` is NOT hard-stopped by the fix — it falls through to the second `WrappedModifiedType` arm and normalizes. This is intentional and correct: `resolveArgument` strips the reference before determining the cast-var type (`val unreferencedType = if (type.isReference) type.unreferenced else type`), so the cast-var ends up value-typed (`std::string arg_cast = std::string(arg)`) and `obj->method(arg_cast)` compiles fine (named lvalue binds to non-const ref). No VALUE→REF miscompile. The task prompt concern is answered: this is not a missing guard.

**Other carriers verified:**
- `const std::string*` (`WrappedPrefixedType → WrappedModifiedType("*")`): prefix recurse hits the pointer hard-stop. ✓
- `std::string* const` (`WrappedModifiedType("*", …)`): outer modifier stops immediately. ✓
- `std::string**`: outer `*` stops immediately. ✓
- `const std::string&`: recurses through `WrappedModifiedType("&") → WrappedPrefixedType("const") → WrappedTemplateType` → normalizes. ✓
- Nested string in template args (e.g. `pair<basic_string<char>,int>`): handled by the non-string `WrappedTemplateType` arm descending into args. ✓
- INCLUDE_MISSING short-circuit unchanged (featuregen policy): confirmed byte-identical post-kplusplusSync.

**Test is non-tautological:** Pre-fix, the `operateOn`-based walk descended through pointer carriers and normalized `std::string*` → the cast-var was declared pointer-typed → `std::string* x = std::string(arg)` → compile error → build abort. Post-fix, the pointer arg drops cleanly and the test assertions hold. Would have failed pre-fix. ✓

## Quality (axis B)

The `when { ... }` rewrite is a clear improvement over the previous `operateOn`-based approach: direct, self-documenting, and easier to reason about the carrier-by-carrier descent. Comment density matches the surrounding code. Naming is consistent. No dead code, no redundancy.

PR #40, opened 2026-06-09
krapper_gen: cumulative forcing recovery — ALL range accessors of one class survive multi-instantiation (TRAVERSE gap)

monkopedia-coder:
## The bug (TRAVERSE gap)

When several element vectors are force-instantiated (`--instantiate`) over the **same** owning class, only **one** of that class's range accessors survives; the others drop with `Couldn't resolve return`. Against Clang's AST: `DeclContext::decls()` (`vector<Decl*>`) survives but `CXXRecordDecl::methods()` (`vector<CXXMethodDecl*>`) and `bases()` (`vector<CXXBaseSpecifier*>`) drop — so the self-bootstrap front-end can EXTRACT (getNameAsString/getAsString) but cannot TRAVERSE a record's methods.

## Root cause (re-confirmed in code)

- Each `requestInstantiation` runs its **own** `resolveForcing` with a **fresh per-call tracker** (`Resolver.kt` builds `baseContext...withClasses(classes)` once per call) holding only **that** request's container specialization.
- `resolveForcing` pass 3 re-resolves the already-bound class (fresh `mappingCache` + `evictKeys`) to recover a range method whose return container is the just-forced spec. It can only recover the accessor whose container is **in this tracker**; the others' return containers `canResolve == false`, so they re-drop.
- The cross-request merge `classes = classes + newClasses` (`IndexedServiceImpl.kt`) appends, and `KotlinWriter` dedups via `associateBy { it.type.toString() }` = **last-wins**. So the last-processed request's owning-class copy (recovering only its own accessor) overwrites the earlier ones. `decls()` survives only because `vector<Decl*>` sorts last.

## The fix — cumulative, order-independent (approach A via a shared accumulator)

Thread a shared `forcedContainerKeys` set across every `requestInstantiation`. Before each forcing run's pass-3 re-resolve, seed the **prior** requests' container keys into the tracker's `otherResolved` (the set `canResolve` short-circuits on); after the run, record the container(s) it forced. So whichever request runs **last** re-resolves the owning class against **every** forced container at once and recovers **all** its range accessors — independent of processing order (the set is cumulative; last always sees the full set).

Constraints honored:
- **Determinism:** recovery no longer depends on `vector<Decl*>` sorting last; the cumulative set is order-independent and stable across runs.
- **No re-explosion:** seeds are query-only (they never enter `resolvedClasses`), so nothing extra is emitted and the shared-tracker bound on pass 2 is untouched.
- **featuregen byte-identical:** seeding is gated to `IGNORE_MISSING` only. Under `INCLUDE_MISSING` (featuregen) nothing drops in pass 1 (methods materialize on-demand), so the path is unchanged. `decls()`/`redecls()` recovery and `evictKeys` semantics are preserved.

## Test

`ReturnShapeTest.multiContainerForcingRecoversAllRangeAccessorsUnderIgnoreMissing` — an `Owner` class (CXXRecordDecl analog) with two `iterator_range` accessors over distinct element types, both force-instantiated, driving `resolveForcing` once per container with the last-wins merge. **Fails pre-fix** (only the last-forced `bitems` survives — confirmed by temporarily disabling the seed: `AssertionError`), **passes post-fix** (both survive). Uses a single forcing parse to stay within the documented repeated-forcing-TU budget.

## Verify (JDK 17, one foreground run + a second sync for stability)

`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck`

- `:krapper_gen:nativeTest` — green (incl. the new test)
- `:featuregen:kplusplusSync` — **byte-identical**, zero `featuregen/krapped/` git drift, stable across two consecutive syncs
- `:featuregen:nativeTest` — green (UP-TO-DATE: generated sources unchanged)

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -92,6 +92,15 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
     // for an unrestricted DefaultFilter import (every non-std class is requested).
     private val requested = mutableListOf<String>()
 
+    // The container specializations forced by instantiation requests SO FAR, accumulated
+    // across every requestInstantiation call. Threaded into each resolveForcing so a later
+    // request's pass-3 re-resolve of an already-bound class can recover EVERY range accessor
+    // of that class (one per forced container) at once — not just the one this request forced,
+    // which the per-call fresh tracker + last-wins merge would otherwise leave as the sole
+    // survivor (e.g. CXXRecordDecl keeping only decls() and dropping methods()/bases()). See
+    // `resolveForcing`'s [forcedContainerKeys] doc.
+    private val forcedContainerKeys = mutableSetOf<String>()
+
     init {
         scope.defer {
             index.dispose()
@@ -254,7 +263,8 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
         val newClasses = scopedFound.resolveForcing(
             resolver,
             config.referencePolicy,
-            alreadyBoundKeys
+            alreadyBoundKeys,
+            forcedContainerKeys
         )
         classes = classes + newClasses.filter {
             it !is ResolvedMethod || it.forcingIdentity in boundMethodIds
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
@@ -184,11 +184,27 @@ suspend fun List<WrappedElement>.resolveAll(
  * bound classes are excluded from the count; the writer's last-wins dedup still picks up
  * the re-resolved bound classes, which is how `items()` is restored). The full
  * resolved set (minus the transient forcing struct) is returned for appending.
+ *
+ * [forcedContainerKeys] is a CUMULATIVE, in/out set of the container specs forced by every
+ * EARLIER instantiation request, shared across the whole run. It exists because each request
+ * gets its OWN `resolveForcing` with a FRESH tracker holding only THAT request's container,
+ * yet a single owning class can have SEVERAL range accessors over DIFFERENT containers (e.g.
+ * Clang's `CXXRecordDecl` with `methods() -> vector<CXXMethodDecl*>`, `bases() ->
+ * vector<CXXBaseSpecifier*>` and `decls() -> vector<Decl*>`). With only the current request's
+ * container in the tracker, pass 3 recovers THAT one range method and re-drops the others
+ * (their containers aren't visible), and the writer's last-wins dedup keeps whichever request
+ * ran last — silently losing the rest (`decls()` survived only by sorting last). Seeding the
+ * prior containers here (IGNORE_MISSING only — see pass 3) makes `canResolve` true for ALL of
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
## Review — APPROVE (tier 1)

**Category:** `bug_fix` — targeted fix for the TRAVERSE gap; no public API change.
**Tier:** 1 (kplusplus `* : tier 1` baseline; no UX/perf lever triggered).

---

### Correctness walkthrough

**Order-independence (3-spec proof):** With containers A/B/C over class X and class Y (only over B), processing order A→B→C:

- After A: `forcedContainerKeys={A}`. Last-wins X has only `aitems`.
- After B: seeds A into `otherResolved`, forces B. Pass 3 re-resolves X (both A via otherResolved and B via resolvedClasses recover) AND Y (B via resolvedClasses recovers `yitems`). `forcedContainerKeys={A,B}`. Last-wins X has {aitems,bitems}; Y has {yitems}.
- After C: seeds A+B into `otherResolved`, forces C. Pass 3 re-resolves ALL bound classes incl. X and Y. X gets all three (A/B via otherResolved, C via resolvedClasses). Y gets `yitems` (B via otherResolved). Last-wins copy is complete.

Permuting to B→A→C or C→A→B yields the same complete outcome because the cumulative set always makes the last request see every prior container. **Order-independent: confirmed.**

**Seeds never leak into output:** `canResolve` short-circuits on `otherResolved` (returns true immediately, line 72/97) without touching `resolvedClasses`. Under IGNORE_MISSING the typeMapper's only action on a canResolve=true result is `ElementUnchanged` — no `resolver.resolve()` call, no new entry in `resolvedClasses`. Seeded containers do not appear as phantom bindings in the emitted output. **Confirmed clean.**

**No re-explosion:** Seeding is into `otherResolved` only, not `resolvedClasses`. Under IGNORE_MISSING, canResolve=true only allows a method to survive; it never triggers on-demand class expansion (that's INCLUDE_MISSING territory). Pass 3 iterates the fixed pre-existing `boundClasses` set — seeding doesn't expand it. **~1269-class bound preserved: confirmed.**

**IGNORE_MISSING gating:** `if (policy == ReferencePolicy.IGNORE_MISSING)` is correct. featuregen (INCLUDE_MISSING) skips the seeding entirely; nothing dropped in pass 1 under INCLUDE_MISSING so there's nothing to recover and the gate is correct. **featuregen structurally untouched: confirmed.**

**Test faithfulness:** Uses real libclang parsing + the actual `resolveForcing` function, drives it once per container with a shared `forcedKeys` set (exact mirror of `IndexedServiceImpl.requestInstantiation`), and asserts the last-wins copy has BOTH accessors. Pre-fix: only the last-forced `bitems` survives (the seed disable makes `aitems` drop). Post-fix: both survive. **Faithful and would fail pre-fix: confirmed.**

**Code quality:** Comment density, naming, and structure are consistent with the surrounding file patterns. `mapNotNull { ... .takeIf { ... } }` is idiomatic. Default argument `= mutableSetOf()` maintains backward compatibility cleanly. No dead code, no ledger violations.

---

### Build verification (local, JDK 17)

- `:krapper_gen:nativeTest` — **GREEN** (incl. `multiContainerForcingRecoversAllRangeAccessorsUnderIgnoreMissing`)
- `:featuregen:kplusplusSync` × 2 — **byte-identical**, zero `featuregen/krapped/` git drift both runs
- `:featuregen:nativeTest` — **UP-TO-DATE** (no generated-source drift)
- `:krapper_gen:ktlintCheck` — **GREEN**

No remote CI configured (repo has only `publish.yaml` on release).

---

Merging per tier-1 policy.

PR #41, opened 2026-06-09
Resolver: drop callback-param methods when function_ref/std::function collapses to a bare type spelling (GAP-A residual)

monkopedia-coder:
## Problem (GAP-A residual)

A method whose parameter is an unmodelable callback `llvm::function_ref<R(Args)>` / `std::function<R(Args)>` (e.g. `clang::ASTContext::adjustType(QualType, llvm::function_ref<QualType(QualType)>)`) was still emitted with an under-arity call (`adjustType(Old)` — the callback param silently trimmed) → C++ compile error → build abort. This aborted the clangwalk self-host (exit-134) on `adjustType` / `forEachMultiversionedFunctionVersion`.

## Why #36 didn't catch it

PR #36 (ca4dd09) extended `WrappedArgument.typeCarriesFalseDefault` to treat a function-TYPE **template argument** as a false default, but gated it on `referent is WrappedTemplateType`. At Clang scale the callback param frequently does **not** surface structured: when libclang doesn't report the template args, `WrappedType.createForType` reconstructs the instantiation via `invoke(String)`, collapsing the whole type into a bare `WrappedTypeReference` whose spelling carries the parenthesized callback signature (`llvm::function_ref<QualType (QualType)>`). This is the same Clang-scale fallback `CollectInheritedProtocols`'s `SmallPtrSetImpl<...>&` already hits. The `referent is WrappedTemplateType` guard returned false, so the misread `hasDefault` cursor heuristic made the param look omittable and it was trimmed.

## Fix

Key the callback false-default signal on a `(` in the **referent's own spelling** (`isFunctionType`), which holds for both the structured `WrappedTemplateType` and the bare-string fallback — instead of only inspecting structured template args. This subsumes #36's per-template-arg check (the referent's `toString()` includes its args), so that branch is reduced to the `UNRESOLVABLE` / non-type-arg cases.

Since the callback param genuinely can't resolve under `IGNORE_MISSING`, the existing skip-not-crash path then drops the **whole method** (ledgered) — same intended behavior as #36.

### Why it can't regress legitimate defaults
`isOmittableDefault` is consulted **only** for a param that already failed to resolve. Legitimate omittable defaults — `const Alloc& = Alloc()`, `shared_ptr<X>`, `buildASTFromCode`'s defaulted trailing params — resolve fine and never reach this gate; and none of their **type** spellings contain `(` (the `()` lives in the default *value*, not the type). They stay omittable.

## Test

`FalseDefaultTest.callback_param_is_not_omittable` covers three shapes: the structured function-type template arg, the bare-string by-value `function_ref`, and the const-reference `adjustType` shape. The two bare-string asserts **fail pre-fix** (wrongly omittable) and pass post-fix; verified by reverting the fix and running just this test (`callback_param_is_not_omittable FAILED`).

## Verify (JDK 17)

`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` → **BUILD SUCCESSFUL**. krapper_gen 311/0, featuregen sync byte-identical (zero `krapped/` git drift, `:featuregen:nativeTest` UP-TO-DATE), ktlintCheck clean. INCLUDE_MISSING (featuregen) behavior unchanged.

## Follow-up

This makes clangwalk's `fixup { removeMethod("clang_ASTContext_adjust_type") }` workaround removable in a later clangwalk pass (not touched here — concurrent agent owns clangwalk).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/WrappedMethod.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/WrappedMethod.kt
@@ -578,10 +578,12 @@ class WrappedArgument(
         get() = hasDefault && !typeCarriesFalseDefault
 
     // True when [hasDefault] is a FALSE positive — the cursor heuristic misread a token in
-    // the param's type as a C++ default, so the param is not legitimately omittable. Three
+    // the param's type as a C++ default, so the param is not legitimately omittable. Four
     // shapes carry a false default: a non-const output reference (the heuristic can misfire
-    // on the `<...>` in `SmallPtrSetImpl<X>&`), an array bound (`int(&)[4]`), and a by-value
-    // template over a NON-TYPE arg (`Mask<4>`'s `4`). A by-value template over only TYPE args
+    // on the `<...>` in `SmallPtrSetImpl<X>&`), an array bound (`int(&)[4]`), an unmodelable
+    // callback (`function_ref<R(Args)>` / `std::function<R(Args)>`, whose parenthesized
+    // signature the heuristic reads as a default), and a by-value template over a NON-TYPE
+    // arg (`Mask<4>`'s `4`). A by-value template over only TYPE args
     // (`shared_ptr<PCHContainerOperations>`) and a CONST reference with a genuine default
     // (`const Alloc& = Alloc()`) keep their real default and stay omittable.
     private val typeCarriesFalseDefault: Boolean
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 (MAIN-MERGE) standard. **Approve + auto-merge.**

## Correctness

**Over-match analysis (key concern — verified safe).**
`isFunctionType` (`'(' in referent.toString()`) is only consulted inside `typeCarriesFalseDefault`, which is only called from `isOmittableDefault`, which is only reached at line 511 for a param that already returned `null` from `resolveArgument()` (i.e., resolution already failed). The two paths that could reach this check with a `(` in their spelling:

- **`WrappedFunctionPointer` (C function-pointer typedef):** `isNative = true`, resolves successfully — never reaches the false-default gate. Its `toString()` returns the typedef name (e.g. `IntTransform`), not the function signature — no `(` anyway.
- **Raw/unmodelable function types** (member function pointers, bare function types, any type that falls through `createForType` to `invoke(spelling)` with a spelling containing `(`): these cannot resolve in the current model. Treating them as false defaults is correct — the alternative is emitting a broken short call that aborts the build.

**Legit defaults preserved (verified).**
`const Alloc&`, `shared_ptr<PCHContainerOperations>`, and `buildASTFromCode`'s trailing defaulted params have no `(` in their **type** spellings. The `Alloc()` expression is the default **value**, not the type — the check operates on `type.unreferenced` at line 609, not on `defaultValue`. Confirmed by tracing the `typeCarriesFalseDefault` path for each case.

**`adjustType` const-reference shape (verified).**
`const llvm::function_ref<QualType (QualType)>&`: non-const check at line 605 passes (it IS const), `referent = const llvm::function_ref<...>` (WrappedPrefixedType), `referent.toString()` = `"const llvm::function_ref<QualType (QualType)>"` — contains `(` → correctly not omittable.

**#36 not regressed.** The structured template case (`std::function<int (Thing *)>`) is now caught earlier by the `referent.isFunctionType` check (the WrappedTemplateType's `toString()` includes its args). The remaining lambda `{ it == UNRESOLVABLE || it.isNonTypeTemplateArg }` correctly handles non-type value args. All 9 pre-existing FalseDefaultTest cases still pass.

## Build (independent, JDK 17)

`./gradlew :krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` → **BUILD SUCCESSFUL**

- krapper_gen: **311/0** (FalseDefaultTest 10/10, including new `callback_param_is_not_omittable`)
- featuregen sync: byte-identical (`nativeTest` UP-TO-DATE, zero `krapped/` git drift)
- ktlintCheck: clean

## Quality

Clean. The `isFunctionTypeArg` → `isFunctionType` rename is appropriate; the property is no longer template-arg–specific. Comment updates are accurate and complete. Lambda reduction (dropping `isFunctionTypeArg` from the template check) correctly reflects the subsumption. No dead code, no style deviations.

Issue #8, opened 2026-06-08
[clang-campaign] Make the stage1 walk representative (methods/bases at runtime) + a real clangwalk test


monkopedia-coder:
**Campaign demo representativeness (Clang self-bootstrap).** The stage1 walk (`clangwalk/src/.../AstWalk.kt`) only iterates `tu.asDeclContext().decls()` + reads `getDeclKindName()`/`getName()`. It NEVER calls `methods()` or `bases()` at runtime — though `build.gradle.kts` force-instantiates `vector<CXXMethodDecl*>` + `vector<CXXBaseSpecifier*>`. So 2 of the 3 "canonical walk" surfaces are compile-only, not run-verified, and the consumer has **no test** (just a `println`).

**Do:** (1) extend the walk to descend into a `CXXRecordDecl`'s `methods()` + `bases()` and assert non-empty; (2) give `clangwalk` an actual test that asserts the expected decls/kinds/names (not a println). Acceptance: the walk exercises decls + methods + bases at runtime under a test.

monkopedia-coder:
PR #42 (squash 878d107) landed the **methods/bases-at-runtime** half: the gated stage1 walk now iterates each `CXXRecordDecl`'s `methods()` (printing `getNameAsString()` + `getReturnType().getAsString()`) and `bases()` (`getType().getAsString()`), with names + return types matching Clang exactly — the reducer's core per-record loop on pure generated bindings.

Remaining scope for this issue: **a real automated clangwalk test** — an assertion-based test (runnable under `-PenableClang`) that runs the walk and checks the output, so the self-bootstrap proof is locked against regression rather than verified by eye. Leaving open for that piece.

monkopedia-coder:
Part 1 (descend a `CXXRecordDecl`'s `bases()` AND `methods()` at runtime) was already complete via PR #42. PR #109 adds **part 2**: the walk is now a REAL assertion test rather than a `println` demo.

[...]

PR #42, opened 2026-06-10
clangwalk: TRAVERSE a record — iterate methods() (name + return type) and bases() (#8)

monkopedia-coder:
## What

Combines the two halves of the self-bootstrap front-end's core loop that already landed on main — **EXTRACT** (#39: `getNameAsString()`/`getAsString()`) and **TRAVERSE** (#40: `methods()`/`bases()` survive multi-instantiation) — into the reducer's core per-record operation: walk a `CXXRecordDecl`, enumerate its methods, and extract each signature's name + return type, plus enumerate its bases.

For each walked `CXXRecordDecl` the stage1 walk now:
- iterates `bases()` and prints each base's `CXXBaseSpecifier::getType().getAsString()`;
- iterates `methods()` and prints each method's `NamedDecl::getNameAsString()` + `FunctionDecl::getReturnType().getAsString()`.

## Changes

- **Fixture** (`AstWalk.kt`): `Shape` gains methods of distinct return types (`int area()`, `const char* name()`, `double scale(double)`) and a derived `struct Circle : Shape` so `bases()` has content.
- **Bind `clang::FunctionDecl`**: it carries `getReturnType()` and bridges the `CXXMethodDecl -> NamedDecl` upcast chain. Without it the chain breaks at the unbound `FunctionDecl` and a walked `CXXMethodDecl` exposes neither name nor return type.
- **Drop the `clang_ASTContext_adjust_type` fixup**: PR #41 (drop callback-param methods when `function_ref`/`std::function` collapses to a bare type spelling) is the underlying cause this fixup papered over. Verified the release walk builds + runs byte-identical without it, and `clang_ASTContext_adjust_type` is now absent from the generated bindings (only the `function_ref`-free `adjust*` overloads remain).

## Verify (JDK 17, `:clangwalk:runReleaseExecutableKlinker -PenableClang`)

```
  [CXXRecord] Point
  [CXXRecord] Shape
      <method> area : int
      <method> name : const char *
      <method> scale : double
  [CXXRecord] Circle
      <base> Shape
      <method> radius : double
  [Var] globalVar : int
```

Names + return types + base match exactly what Clang reports for the fixture. **#40 delivered**: `methods()` and `bases()` both surfaced and survived together off the same owning class.

Gated behind `-PenableClang` (RELEASE only); the default LLVM-absent build is unaffected. Addresses #8.

### DIFF
--- a/clangwalk/build.gradle.kts
+++ b/clangwalk/build.gradle.kts
@@ -90,6 +90,12 @@ kplusplus {
         "clang::TagDecl",
         "clang::RecordDecl",
         "clang::CXXRecordDecl",
+        // FunctionDecl is CXXMethodDecl's direct base: it carries getReturnType() (the
+        // method's return QualType — the signature payload the reducer extracts) and bridges
+        // the CXXMethodDecl -> NamedDecl upcast chain (getNameAsString). Without it bound the
+        // chain breaks at FunctionDecl and a walked CXXMethodDecl exposes neither name nor
+        // return type.
+        "clang::FunctionDecl",
         "clang::CXXMethodDecl",
         "clang::FieldDecl",
         "clang::ValueDecl",
--- a/clangwalk/build.gradle.kts
+++ b/clangwalk/build.gradle.kts
@@ -112,13 +112,4 @@ kplusplus {
     instantiate("std::vector<clang::Decl*>")
     instantiate("std::vector<clang::CXXMethodDecl*>")
     instantiate("std::vector<clang::CXXBaseSpecifier*>")
-    // ASTContext::adjustType(QualType, llvm::function_ref<QualType(QualType)>) has an
-    // un-modelable function_ref second arg. With QualType now bound its return resolves, so
-    // the method is no longer dropped-by-unresolved-return — and the function_ref param is
-    // mis-flagged omittable-default (the cursor-children default heuristic fires on the
-    // function_ref's type tokens), emitting an under-arity `adjustType(Old)` call that won't
-    // compile. Drop it explicitly until that heuristic is tightened (a krapper_gen follow-up).
-    fixup {
-        removeMethod("clang_ASTContext_adjust_type")
-    }
 }
--- a/clangwalk/src/nativeMain/kotlin/AstWalk.kt
+++ b/clangwalk/src/nativeMain/kotlin/AstWalk.kt
@@ -22,7 +22,13 @@ import kotlinx.cinterop.memScoped
 fun main(args: Array<String>) = memScoped {
     val source = """
         struct Point { int x; int y; };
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
LGTM. Reviewed to merge standard — both axes clean.

**Gating:**  remains strictly opt-in under `-PenableClang`; confirmed absent from the default project graph and `:krapper_gen:nativeTest`+`:krapper_gen:ktlintCheck` green unaffected.

**Fixup removal:** verified. The release walk builds and runs correctly without the `clang_ASTContext_adjust_type` fixup. PR #41's function_ref→collapse drop means the method no longer appears in the generated bindings at all, so the manual removal guard was dead weight. Justified.

**Walk output:** `:clangwalk:runReleaseExecutableKlinker -PenableClang` produced exactly the expected output — `[CXXRecord] Shape` with `area`/`name`/`scale` methods, `[CXXRecord] Circle` with `<base> Shape` + `radius` method, `[Function] freeFunction`, `[Var] globalVar`. #40 TRAVERSE + #39 EXTRACT together, proven end-to-end.

**Quality:** `AstWalk.kt` is idiomatic Kotlin — null-safe `?.` chains, `?:` fallbacks, per-null `continue` guards in the loops, comment density consistent with the file's established pattern. The `FunctionDecl` addition to `only()` is minimal and correctly bridges the `CXXMethodDecl → NamedDecl` upcast chain. No libclang-C leak; pure generated-binding usage throughout.

monkopedia-reviewer (approved):
LGTM. Reviewed to merge standard — both axes clean.

**Gating:** :clangwalk remains strictly opt-in under `-PenableClang`; confirmed absent from the default project graph and :krapper_gen:nativeTest + :krapper_gen:ktlintCheck green unaffected.

**Fixup removal:** verified. The release walk builds and runs correctly without the `clang_ASTContext_adjust_type` fixup. PR #41's function_ref-collapse drop means the method no longer appears in the generated bindings at all, so the manual removal guard was dead weight. Justified.

**Walk output:** runReleaseExecutableKlinker -PenableClang produced exactly the expected output — [CXXRecord] Shape with area/name/scale methods, [CXXRecord] Circle with base Shape + radius method, [Function] freeFunction, [Var] globalVar. #40 TRAVERSE + #39 EXTRACT together, proven end-to-end.

**Quality:** AstWalk.kt is idiomatic Kotlin — null-safe ?. chains, ?: fallbacks, per-null continue guards in the loops, comment density consistent with the file's established pattern. The FunctionDecl addition to only() is minimal and correctly bridges the CXXMethodDecl -> NamedDecl upcast chain. No libclang-C leak; pure generated-binding usage throughout.

PR #43, opened 2026-06-10
krapper_gen: emit a plain method when an operator's idiomatic name collides (operator fun _get → fun _get)

monkopedia-coder:
## Root cause

When a class exposes BOTH a symbolic operator and a named sibling that owns the operator's idiomatic Kotlin name — e.g. `clang::TemplateArgumentList` has both `operator[](unsigned)` and a named `get(unsigned)` — the scope uniquifier allocates the canonical name (`get`) to the named sibling first, then renames the colliding `operator[]` to `_get`. But `generateOperator` still wrapped the emission in the `operator { }` modifier, producing `operator fun _get`, which is **illegal Kotlin** (the `operator` modifier is only legal on the canonical operator identifier). The K/N compile of the generated bindings then fails.

This was surfaced by binding `clang::TemplateArgumentList` during the self-bootstrap **type-decode probe** (P5, template arguments), where it blocked the front-end type-decode surface from compiling.

## Fix

When the operator's idiomatic name is already claimed in the current scope, emit a **plain (uniquified) method** instead of an `operator fun`:

- `Scope.kt`: add `isNameUsed(name)` so a caller can detect that a later `allocateName` of that name would be uniquified to `_<name>`.
- `KotlinWriter.kt` (`generateOperator`): gate the `operator { }` wrapper on the name not being a uniquified collision. When it is taken, drop the modifier.

The operator-call syntax (`a[i]`, `a + b`, ...) still works because it falls back to the named sibling that owns the canonical name, and index-iterator wiring already targets that sibling by name — so nothing observable is lost.

## Verification (JDK 17, foreground)

- `:krapper_gen:nativeTest` — **311/0** green
- `:featuregen:kplusplusSync` + `:featuregen:nativeTest` — **188/0** green, `featuregen/krapped/` **byte-identical** (zero git drift, full `--rerun-tasks`)
- `:krapper_gen:ktlintCheck` — clean

Generator-only change; clears a codegen blocker for the front-end type-decode surface in the Clang self-bootstrap.

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/builders/Scope.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/builders/Scope.kt
@@ -26,6 +26,14 @@ class Scope<T : LangFactory>(internal val parent: Scope<T>? = null) {
 
     private fun isUsed(name: String): Boolean = names.contains(name) || parent?.isUsed(name) == true
 
+    /**
+     * Whether [name] is already claimed in this scope (or an enclosing one) — i.e. a later
+     * [allocateName] of it would be uniquified to `_$name`. Lets a caller decide NOT to emit
+     * a construct that requires the exact name (e.g. the `operator` modifier, which is only
+     * legal on the canonical operator identifier — `operator fun _get` does not parse).
+     */
+    fun isNameUsed(name: String): Boolean = isUsed(name)
+
     fun allocateName(desiredName: String): String {
         if (desiredName.isEmpty()) return allocateName("v")
         if (isUsed(desiredName)) {
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -71,6 +71,7 @@ import com.monkopedia.krapper.generator.builders.pkg
 import com.monkopedia.krapper.generator.builders.property
 import com.monkopedia.krapper.generator.builders.qdot
 import com.monkopedia.krapper.generator.builders.reference
+import com.monkopedia.krapper.generator.builders.scope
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
## Review: approve ✓

**Category:** `bug_fix` — illegal Kotlin emission (`operator fun _get`) when operator's idiomatic name collides with a named sibling. Surfaced by the clang self-bootstrap type-decode probe.
**Tier:** 1 (kplusplus policy `* : tier 1`).

### Correctness

**Gate correctness (load-bearing check):** `scope.isNameUsed(kotlinType.name)` delegates to the same `isUsed()` logic that `allocateName` consults, so the gate is precisely consistent with the uniquifier. When the named sibling has already claimed the idiomatic name, `isNameUsed` returns `true` → `operator` wrapper is dropped → `generateBasicMethod` allocates `_name`. For normal operators (no collision), `isNameUsed` is `false` → `operator(emit)` fires unchanged. No regression risk to legitimate `operator fun` emissions.

**Fallback validity:** The named `get(idx)` sibling owns `operator fun get`; the synthesized iterator from `detectIndexIterable` already targets `getName = "get"` by name (derived from the operator's idiomatic name), and that call resolves to the named sibling — semantically equivalent to `operator[]`.

**Added import** (`scope`): previously only consumed indirectly via `define()`; direct call here requires the import. Correct and minimal.

**`isNameUsed()`:** thin public wrapper of `private fun isUsed`, kdoc explains the exact use case. Minimal addition.

### Quality

The `emit: KotlinCodeBuilder.() -> Unit = { ... }` lambda pattern with `if (nameTaken) emit() else operator(emit)` is consistent with the `METHOD` path in the same file. Comment block is informative without being redundant. No dead code, no duplication, no style-ledger violations.

### Build (independent, JDK 17)

- `:krapper_gen:nativeTest` — green (UP-TO-DATE against commit 97701ff)
- `:featuregen:kplusplusSync` — 17 instantiations regenerated, zero `featuregen/krapped/` git drift
- `:featuregen:nativeTest` — green (UP-TO-DATE)
- `:krapper_gen:ktlintCheck` — clean

LGTM.

Issue #44, opened 2026-06-10
[clang-campaign] Phase B: front-end module — typed recursive-descent WrappedTU producer on generated bindings


monkopedia-coder:
The constructive core of the self-bootstrap: a new front-end module that produces krapper_gen's `WrappedTU` parse-output tree from the **generated libclang-cpp bindings** (typed recursive descent: `decls()` → `methods()`/`fields()` → `parameters()` → types), replacing `Parsing.kt`'s libclang-C cursor walk. Resolve + codegen stay untouched (the swap point is the WrappedTU).

## Foundations (all PROVEN at runtime on generated bindings, 2026-06-09/10)
- **EXTRACT** — `getNameAsString()` / `QualType::getAsString()` (#37/#38/#39)
- **TRAVERSE** — `decls()` / `methods()` / `bases()` survive multi-instantiation (#40, demonstrated #42)
- **IDENTITY** — canonical `Decl::getID()` (Long, via `asDecl().getCanonicalDecl()`) is a drop-in for the USR-keyed `elementLookup`/`classMap`/template-substitution contract. Probe branch: `clangwalk/decl-identity-probe` (b9e19dc)
- **TYPE-DECODE** — qualifiers, shape classification, pointee, record→decl recovery, `FunctionProtoType` params/return, `TemplateSpecializationType` args all bind; **`getCanonicalType()` replaces the ~200-line typedef-reducer**. Probe branch: `clangwalk/type-decode-probe` (59ebf04)

## Brick breakdown (each its own PR; sole-agent, commit-before-build, finish-in-turn)
1. **Module scaffold**: new gated module (or grow clangwalk) producing a serialized `WrappedTU` for the simplest fixture (a free function + a trivial class); plumb the brick-5 identity key where USR goes today.
2. **Decl/class/method construction**: records, methods, fields, ctors/dtors, access, static-ness, namespaces — the `WrappedElement.map()` equivalent as typed descent.
3. **Type construction**: `QualType` → `WrappedType` (const/ref/ptr/array/builtin/record/enum), using canonical collapse instead of the typedef-reducer.
4. **Templates**: class-template specializations, template args → the WrappedTemplate model, substitution keyed on canonical getID.
5. **Enums + function pointers + default args**: replace the token-scrape with `ParmVarDecl::getDefaultArg()` + `EvaluateAsInt`.
[...]

PR #52, opened 2026-06-10
cppfrontend: structural type construction — decode the QualType tree into TypeFactories' WrappedType shapes (#44 brick 4)

monkopedia-coder:
Phase B brick 4 of #44: `cppfrontend`'s `QualType -> WrappedType` mapping is now STRUCTURAL — the type tree is decoded (qualifiers, pointee recursion, template args, array extents, typedef decls) instead of parsed back out of `getAsString()` spellings, mirroring `TypeFactories.createForType`/`invoke(CValue<CXType>)` rule-for-rule with in-code citations.

## Mirrored rules (each cited in TypeBuilder.kt)
- **Pointer / lvalue-reference** → `pointerTo`/`referenceTo(decode(pointee)).maybeConst` — invoke's `endsWith("*"/"&")` dispatch; a pointee's `const` nests INSIDE through the recursion (libclang `clang_isConstQualifiedType` semantics match `QualType::isConstQualified` exactly).
- **Rvalue refs** → `UNRESOLVABLE` (invoke's `RValueReference` throw under the call sites' `throwOnError=false`).
- **Template-specialization types** → `WrappedTemplateType(WrappedTypeReference(qualifiedName), args)` via `ClassTemplateSpecializationDecl::getTemplateArgs()`; non-type args dropped by `ArgKind.Type` — the bindings DID surface the nested `TemplateArgument::ArgKind` enum, so this mirrors the `getTemplateArgumentType -> CXType_Invalid -> filterNotNull` drop exactly (no residual).
- **Constant arrays** → the libclang `"elem [N]"` leaf (TestData golden `"int [5]"`), rebuilt from `ArrayType::getElementType()` + `ConstantArrayType::getZExtSize()`.
- **Record leaves** → `WrappedTypeReference(getQualifiedNameAsString())` — the Decl-side spelling of `fullyQualified`'s semantic-parent walk (`geo::Circle*` asserted).
- **Typedefs**: `size_type`/`difference_type` → `size_t`/`ptrdiff_t` (sizeTypedefElement); `size_t`/`ssize_t`/`ptrdiff_t`/`intptr_t`/`wchar_t` preserved as the alias NAME (cAliasTypedefElement — a `size_t` param emits `WrappedTypeReference("size_t")`, same as the libclang path); every other alias collapses to its underlying type (`ResolverBuilderImpl.visit`'s TypedefDecl branch).
- **G8 const-method rule** (ModelFactories.WrappedMethod) now applies over structural types: a const method's `Shape*` return decodes as `WrappedPrefixedType(const, WrappedModifiedType(*, Shape))` — OUTER const, asserted structurally.

## Bridges + documented divergences (Phase C normalizer entries)
- **`_Bool` vs `bool` (brick-3 finding)**: `QualType::getAsString()`'s no-policy overload prints with `PrintingPolicy(LangOptions())` (C mode). Resolution: **normalize the spelling** (`_Bool` → `bool`) rather than bind `PrintingPolicy` + `ASTContext::getPrintingPolicy()` — a whole class surface for one token; documented at `spellingOf()`.
- **`using`-alias order-dependence**: the libclang path collapses a `using` alias only if `visit()`'s `seenNames` cache already saw the underlying's canonical spelling — otherwise the alias NAME survives as a leaf. This front-end always collapses both alias kinds (deterministic); divergence documented in the typedef branch.
- **Enum leaves**: libclang emits `WrappedEnumType` (name + underlying int + constants); the constants' values need an `llvm::APSInt` bridge — structural decode emits the named reference, brick-5 note in place (no enum in fixture).
- **Bare fn-pointers**: opaque spelling leaf, same as libclang's `else` fall-through; the rich `WrappedFunctionPointer` only exists for fn-pointer TYPEDEFS (`functionPointerTypedefElement`) — `// brick-6:` comment, fixture case dropped as disproportionate (needs the underlying-proto decode + `isCFunctionPointerCompatible` recursion).
- **Missing generated down-cast** `Type.asTypedefType()`: TypedefType's base list doesn't survive resolution (`final` + private `TrailingObjects` base) so CastTargets never emits the helper — bridged via the bound static llvm `classof` + same-pointer re-view (exactly what the generated `B_dyncast_D` compiles to); krapper_gen gap noted in code.

## Fixture / checks delta
- Fixture grows `Scene` + in-fixture `Holder<T>` (chosen over `std::vector` to avoid std-header weight and keep the decoded shape byte-deterministic) + `typedef size_t`/`MyInt`/`using Real` + `geo::makeCircle()`.
- Self-checks **34 → 51**, all green — structural asserts (`is WrappedModifiedType && baseType is WrappedPrefixedType && ...`) plus toString round-trips per rule.
- `only()` grows by the probe-proven decode slice: `clang::Type`, `TypedefType`/`TypedefNameDecl`, `ArrayType`/`ConstantArrayType`, `ClassTemplateSpecializationDecl`/`TemplateArgument(List)`. No `:krapper_model` changes (featuregen gate not required).

## Verify
- Default unaffected: `:krapper_gen:nativeTest :krapper_model:build :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL.
- Gated: `:cppfrontend:runReleaseExecutableKlinker -PenableClang` — BUILD SUCCESSFUL, **ALL 51 SELF-CHECKS PASSED**.

```
[PASS] bool return: builtin leaf normalizes _Bool -> WrappedTypeReference("bool") — got bool
[PASS] makeCircle return spells the QUALIFIED record leaf: 'geo::Circle*' — got geo::Circle*
[PASS] Shape* return: WrappedModifiedType("*") over record leaf WrappedTypeReference("Shape") — got Shape*
[PASS] const method's Shape* return gains OUTER const: WrappedPrefixedType(const, Shape*) — got const Shape*
[PASS] const Shape& param: WrappedModifiedType("&") over WrappedPrefixedType(const) over Shape — got const Shape&
[PASS] size_t param preserved as the NATIVE alias leaf WrappedTypeReference("size_t") — got size_t
[PASS] size_type return normalizes to WrappedTypeReference("size_t") — got size_t
[PASS] typedef'd MyInt param collapses to the underlying builtin 'int' — got int
[PASS] `using Real` param collapses to the underlying builtin 'double' — got double
[PASS] int[4] field: WrappedTypeReference("int [4]") with isArray, size 4, element int — got int [4]
[PASS] Holder<Shape*> field: WrappedTemplateType(base WrappedTypeReference("Holder"), 1 arg) — got Holder<Shape*>
[PASS] Holder's template arg decodes structurally to *(Shape) — got Shape*
cppfrontend: ALL SELF-CHECKS PASSED
```

Decoded-structure sample (serialized WrappedTU, Scene):

```json
{ "kind": "method", "name": "cloneShape", "returnType": "const Shape*", "isConst": true },
{ "kind": "method", "name": "describe",
  "children": [ { "kind": "arg", "name": "shape", "type": "const Shape&", "usr": "cpp:3104" } ] },
{ "kind": "method", "name": "reserve",
  "children": [ { "kind": "arg", "name": "capacity", "type": "size_t", "usr": "cpp:3159" } ] },
{ "kind": "field", "name": "corners", "type": "int [4]" },
{ "kind": "field", "name": "shapes", "type": "Holder<Shape*>" }
```

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -117,8 +117,32 @@ kplusplus {
         "clang::ParmVarDecl",
         "clang::VarDecl",
         // QualType is the carrier for a decl's type; getAsString() (by-value std::string,
-        // normalized to a Kotlin String by #37) feeds the WrappedType(spelling) factory.
-        "clang::QualType"
+        // normalized to a Kotlin String by #37) spells the leaves.
+        "clang::QualType",
+        // NEW for brick 4 (structural type construction, TypeBuilder.kt): clang::Type
+        // carries the shape-classification predicates (isPointerType/isReferenceType/
+        // isRValueReferenceType/isFunctionPointerType/isConstantArrayType/isEnumeralType),
+        // getPointeeType(), and the record recovery (getAsCXXRecordDecl) — the decode
+        // surface proven on clangwalk/type-decode-probe (59ebf04).
+        "clang::Type",
+        // Typedef/alias decode: a sugared QualType's Type dyn-casts to TypedefType, whose
+        // getDecl() (a TypedefNameDecl — covers both `typedef` and `using`) provides the
+        // alias NAME (the size_t/size_type preservation rules) and getUnderlyingType()
+        // (the canonical-collapse recursion). TypeDecl (already bound) bridges the
+        // TypedefNameDecl -> NamedDecl upcast chain.
+        "clang::TypedefType",
+        "clang::TypedefNameDecl",
+        // Constant-array decode: ConstantArrayType::getZExtSize() is the extent and
+        // ArrayType (its direct base) carries getElementType() — together they reconstruct
+        // the libclang "elem [N]" array leaf structurally.
+        "clang::ArrayType",
+        "clang::ConstantArrayType",
+        // Template-specialization decode (probe-proven): a template-typed field/param's
+        // canonical record dyn-casts to ClassTemplateSpecializationDecl ->
+        // getTemplateArgs() -> TemplateArgument::getAsType() per argument.
+        "clang::ClassTemplateSpecializationDecl",
+        "clang::TemplateArgument",
+        "clang::TemplateArgumentList"
     )
     // Materialize the range returns the walk iterates. Brick 3 walks members through
     // DeclContext::decls() (source order, all member kinds) instead of methods(), so the
--- a/cppfrontend/include/clang_slice.h
+++ b/cppfrontend/include/clang_slice.h
@@ -1,4 +1,6 @@
 #include <clang/Frontend/ASTUnit.h>
 #include <clang/AST/DeclCXX.h>
+#include <clang/AST/DeclTemplate.h>
 #include <clang/AST/ASTContext.h>
+#include <clang/AST/Type.h>
 #include <clang/Tooling/Tooling.h>
--- a/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
+++ b/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
@@ -22,6 +22,10 @@ import com.monkopedia.krapper.generator.model.WrappedElement
 import com.monkopedia.krapper.generator.model.WrappedField
 import com.monkopedia.krapper.generator.model.WrappedMethod
 import com.monkopedia.krapper.generator.model.WrappedNamespace
+import com.monkopedia.krapper.generator.model.type.WrappedModifiedType
+import com.monkopedia.krapper.generator.model.type.WrappedPrefixedType
+import com.monkopedia.krapper.generator.model.type.WrappedTemplateType
+import com.monkopedia.krapper.generator.model.type.WrappedTypeReference
 import com.monkopedia.krapper.generator.resolvedmodel.MethodType
 import kotlin.system.exitProcess
 import kotlinx.cinterop.memScoped
@@ -33,6 +37,11 @@ import kotlinx.serialization.json.Json
 // with metadata), ctors (default + copy) and a virtual dtor, a deleted copy assignment,
 // a static method, public/private fields (incl. the const-field metadata rules), an
 // abstract class, namespace nesting with block merging, and a derived class's bases.
+// Brick 4 grows it with the STRUCTURAL type surface (Scene + the aliases + Holder): a
+// `const Shape&` param, `Shape*` returns (one through the G8 const-method rule), a
+// constant-array field, size_t/size_type preservation, typedef/`using` collapse, and a
+// template-typed field. Holder is defined in-fixture rather than using std::vector so the
+// case carries no std-header weight and the decoded shape is byte-deterministic.
 private val FIXTURE_HEADER = """
     void freeFunction(int);
 
@@ -68,7 +77,28 @@ private val FIXTURE_HEADER = """
 
     namespace geo {
         bool enabled();
+        Circle* makeCircle();
     }
+
+    typedef unsigned long size_t;
+    typedef int MyInt;
+    using Real = double;
+
+    template <typename T>
+    struct Holder { T value; };
+
+    struct Scene {
+        typedef unsigned long size_type;
+        Shape* current();
+        Shape* cloneShape() const;
+        void describe(const Shape& shape);
+        void reserve(size_t capacity);
+        void setId(MyInt id);
+        void setScale(Real scale);
+        size_type size() const;
+        int corners[4];
+        Holder<Shape*> shapes;
+    };
 """.trimIndent()
 
 private var failures = 0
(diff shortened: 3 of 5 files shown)


monkopedia-reviewer (changes requested):
**Tier 1 — request_changes before auto-merge (one comment fix).**

## Shape-fidelity spot-checks (all pass)

Verified 4 structural mirrors against `TypeFactories.kt` + the model type classes:

**1. `const Shape&` const-nesting order.**
`TypeFactories` path: `referenceTo(invoke(pointeeType)).maybeConst(ref.isConst)`. For `const Shape&`, the pointee QualType is `const Shape`, so the recursion builds `const(Shape)` = `WrappedPrefixedType("const", Shape)` first, then `referenceTo(...)` wraps it → `WrappedModifiedType("&", WrappedPrefixedType("const", Shape))`. The reference itself is not const. `TypeBuilder.kt` produces the identical tree (const nests INSIDE the `&`). Self-check asserts the exact structure plus `toString()` = `"const Shape&"`. ✓

**2. `size_t` preservation as native alias.**
`TypeFactories.cAliasTypedefElement` returns `WrappedType("size_t")` = `WrappedTypeReference("size_t")` (native). `TypeBuilder` hits the same `C_ALIAS_TYPEDEFS` set and returns `WrappedTypeReference("size_t")` with `isNative = true` (size_t is in `WrappedTypeReference`'s NATIVE list). Self-check asserts both name and `isNative`. ✓

**3. Array leaf spelling with space (`"int [4]"`).**
`TypeFactories` lets `ConstantArrayType` fall through to `WrappedType(spelling)` where libclang spells `"int [5]"` (TestData golden, space before `[`). `TypeBuilder` reconstructs `"$element [${getZExtSize()}]"` = `"int [4]"` — same space-before-bracket format. `WrappedTypeReference.arraySize` / `arrayType` parse correctly off that. ✓

**4. G8 const-method outer-const on `Shape*`.**
`ModelBuilder.buildMethod` applies `WrappedType.const(returnType)` when `isConst && (isPointer || isReference)`. `buildWrappedType(Shape*)` = `WrappedModifiedType("*", Shape)`. `.isPointer` = true on `WrappedModifiedType("*",…)`. So `const(WrappedModifiedType("*", Shape))` = `WrappedPrefixedType("const", WrappedModifiedType("*", Shape))` = OUTER const. Matches libclang path exactly. ✓

## classof bridge — sound, but the comment contains an incorrect class declaration

`TypeBuilder.kt` lines 45-52, the comment justifying the raw pointer re-view states:

```
`class TypedefType final : public Type, ..., private llvm::TrailingObjects<...>`
```

**The actual declaration in `/usr/include/clang/AST/TypeBase.h:6078` is:**

```cpp
class TypedefType final
    : public TypeWithKeyword,
      private llvm::TrailingObjects<TypedefType, ...>
```

TypedefType does NOT directly inherit from `Type`. The chain is:
```
TypedefType → TypeWithKeyword → KeywordWrapper<Type> → Type (first base)
```

`KeywordWrapper<T>` inherits `T` first, then `KeywordHelpers` (a struct of only static methods → EBO); `TypeWithKeyword` has `KeywordWrapper<Type>` as its only base; `TypedefType` has `TypeWithKeyword` as its first (primary) base. So through the chain the address equality `&TypedefType == &Type` still holds, and the bridge **is** sound at runtime.

However, the comment says "Type is TypedefType's **primary** base" — technically TypeWithKeyword is TypedefType's primary base, and Type is reachable through two additional levels of indirection. For a raw pointer re-view the justification comment must be accurate: a maintainer who checks the header will see `public TypeWithKeyword`, not `public Type`, and have to re-derive the chain from scratch.

**Request:** correct lines 46-47 and 51 to state the actual direct parent and the full derivation chain. Something like:

```kotlin
// `class TypedefType final : public TypeWithKeyword, private llvm::TrailingObjects<...>`
// where TypeWithKeyword : KeywordWrapper<Type> : Type (first base). Through the chain
// Type remains the address-identical primary base (EBO on KeywordHelpers — static
// methods only), so the same-pointer re-view is safe.
```

## Other findings

- **`:krapper_model` delta:** zero (verified `git diff main -- krapper_model/` produces no output). ✓
- **All 4 documented divergences present** (rvalue-ref → UNRESOLVABLE, `_Bool`→`bool` PrintingPolicy bridge, `using`-alias order-dependence, enum leaf deferred to brick-5). ✓
- **Fixture + self-checks:** 34→51, structural asserts + `toString` round-trips. Checks are substantive (they would fail with a different fixture). ✓
- **Idiom, naming, comment density:** consistent with `ModelBuilder.kt` and codebase style. ✓

## Build results (independent, JDK 17)

- `:krapper_gen:nativeTest :krapper_model:build :krapper_gen:ktlintCheck` → **BUILD SUCCESSFUL** (default unaffected ✓)
- `:cppfrontend:runReleaseExecutableKlinker -PenableClang` → **BUILD SUCCESSFUL, cppfrontend: ALL SELF-CHECKS PASSED** (all 51) ✓

---
One blocking item: fix the `asTypedefTypeOrNull` comment to state the actual inheritance chain so the pointer re-view safety justification is verifiable. Everything else is sound.

monkopedia-reviewer (approved):
Re-review of cbae01f (comment-only fixup).

**Delta check:** The commit touches exactly one region — the comment block above `asTypedefTypeOrNull` in `TypeBuilder.kt`. Zero code lines changed.

**Header accuracy (TypeBase.h:6078):**
- Declaration now reads `class TypedefType final : public TypeWithKeyword, private llvm::TrailingObjects<...>` — matches the header exactly.
- Chain `TypedefType → TypeWithKeyword → KeywordWrapper<Type> → Type` confirmed from lines 5956 and 5937.
- `KeywordHelpers` (also at 5912) is a struct with only `static` methods and no data members — empty by the standard definition, so EBO applies and the address-identity argument is sound.

**Build re-verification not required:** the prior review already confirmed both CI gates green and 51/51 self-checks pass on the unchanged binary. A comment-only patch cannot affect any of those results.

All findings from the CHANGES_REQUESTED cycle are resolved. Approving and merging tier-1 (squash).

PR #53, opened 2026-06-10
cppfrontend: template declarations + enum construction (#44 brick 5)

monkopedia-coder:
Phase B brick 5 of #44: template DECLARATIONS and enum construction in `cppfrontend/`, mirrored rule-for-rule against the libclang-C front-end with in-code citations.

## Template declarations (ModelFactories' ClassTemplate branch)
- A `ClassTemplateDecl` becomes the model's `WrappedTemplate(spelling)` (ModelFactories `CXCursor_ClassTemplate` + the `WrappedTemplate` factory); its TYPE params become `WrappedTemplateParam(spelling, usr)` (`CXCursor_TemplateTypeParameter`); non-type/template-template params are skipped exactly as map's `else -> null` does.
- The pattern record's bases + members run through the SAME shared class-construction path (`addBasesAndMembers`, refactored to take `parent: WrappedElement` + `ClassMetadata` so WrappedClass and WrappedTemplate share every member branch and metadata side-effect — the decl-side equivalent of mapAll's recursion + the `(parent as? WrappedTemplate)?.metadata` mirrors).
- **Dependent-type shape (the question the brick asked):** the libclang path produces `WrappedTemplateRef(param decl's USR)` — NOT `type-parameter-0-0`, NOT the bare name `T` (TypeFactories' `CXCursor_TemplateTypeParameter` branch). Mirrored: a `TemplateTypeParmType` decodes to `WrappedTemplateRef("cpp:<canonical-id>")`, read off the SUGARED type (the canonical one carries no decl — `assert(!TTPDecl == Canon.isNull())`), keyed to the SAME identity as the matching `WrappedTemplateParam.usr` so substitution can match. Self-checked: field/return/`const T&` arg/member typedef all key to the param's usr.
- Typedef ELEMENTS (map's `CXCursor_TypedefDecl` branch, level-agnostic): `WrappedTypedef(name, reducer-stack target)` at TU/class/template level; a `using` alias produces NO element (no `CXCursor_TypeAliasDecl` branch) — mirrored with a `Kind.Typedef` gate. The target runs the shared `buildTypedefTargetType` (size_type/difference_type/C-alias rules + underlying collapse), so Scene's `size_type` element lands on `size_t` exactly like `sizeTypedefElement`.

## Enums (TypeFactories' CXCursor_EnumDecl branch)
- An `EnumDecl` itself produces **NO element** — faithful to ModelFactories.map, which has no EnumDecl branch (`else -> return null`). The payload rides on every enum-typed LEAF as `WrappedEnumType(cppName, underlying, constants)`:
  - `cppName` = qualified name (`fullyQualified` ≡ `getQualifiedNameAsString`; nested enum spells `Palette::Flavor`),
  - `underlying` = `EnumDecl::getIntegerType()` (≡ `clang_getEnumDeclIntegerType`): explicit for fixed enums, clang's computed type otherwise,
  - `constants` = the `EnumConstantDecl` children's (name, value) pairs.
- **Value-extraction choice (documented):** libclang uses `getInitVal().getSExtValue()` (CIndex.cpp). Here the value reads `llvm::APSInt::getExtValue()` — the smallest bindable surface (APSInt's OWN method; `getSExtValue` lives on the unbound `llvm::APInt` base) and sign-correct. Identical for every signed constant and unsigned constants below the top bit; for an unsigned constant WITH the top bit set libclang reports a sign-mangled negative — recorded as a Phase C normalizer entry (mask: sign-extend the libclang value by the underlying width).
- Expected underlyings/values were pinned against a real libclang oracle run on the exact fixture (`clang_getEnumDeclIntegerType` + `clang_getEnumConstantDeclValue`): Color → `unsigned int`, RED=0/GREEN=5; Mode → `unsigned char`, A=0/B=1; Status → `long`, OK=-2; Palette::Flavor → `unsigned int`, qualified spelling.

## only() / bridges / fixups
- only() += `clang::TemplateDecl`, `ClassTemplateDecl`, `TemplateParameterList`, `TemplateTypeParmDecl`, `TemplateTypeParmType`, `EnumDecl`, `EnumConstantDecl`, `llvm::APSInt`.
- `clang::EnumType` is deliberately NOT bound (generator gap #1): it declares no own ctors (only `using TagType::TagType`), so nothing marks `hasConstructor` and krapper synthesizes an ill-formed `new clang::EnumType()` default-construct wrapper. The enum decl is reached through the already-bound `Type::getAsTagDecl()` + generated `Decl.asEnumDecl()` instead.
- Fixups (generator gap #2, v8-precedent `removeMethod`): four of `llvm::APSInt`'s operators emit invalid Kotlin — `operator==(int64_t)`/`operator<(int64_t)` (collision-suffixed `override fun _equals`/`operator fun _compareTo`) and prefix `++`/`--` (nullable `inc()`/`dec()` returns). All four removed by uniqueCName; none is needed by the value bridge. Both gaps documented in build.gradle.kts for a krapper_gen follow-up.
- Probe-confirmed cast surface: `Decl.asEnumDecl/asEnumConstantDecl/asTemplateTypeParmDecl` and `Type.asTemplateTypeParmType` ARE generated (used directly); `Decl.asClassTemplateDecl` is NOT (`RedeclarableTemplateDecl` unbound breaks the chain) — bridged with the TypedefType-pattern kind-gated re-view; `getTemplateParameters()` reached by TemplateDecl re-view (primary-base, address-identical).

## Fixture + self-checks: 52 → 72 (ALL PASS)
```
template<T> class Holder (5 children)
  typedef value_type = template<cpp:2974>
  val value: template<cpp:2974>
  fun get(): template<cpp:2974> [METHOD] const
  fun set(v: const template<cpp:2974>&): void [METHOD]
class Palette (7 members)
  val primary: Color
  fun cycle(from: Color): Color [METHOD]
  fun currentMode(): Mode [METHOD] const
  fun status(): Status [METHOD] const
  fun flavor(): Palette::Flavor [METHOD] const
```
Serialized template (the params/members/dependent-type shape):
```json
{ "kind": "template", "name": "Holder", "children": [
    { "kind": "templateParam", "name": "T", "usr": "cpp:2974" },
    { "kind": "typedef", "name": "value_type", "type": "template<cpp:2974>" },
    { "kind": "field", "name": "value", "type": "template<cpp:2974>" },
    { "kind": "method", "name": "get", "returnType": "template<cpp:2974>", "methodType": "METHOD", "isConst": true },
    { "kind": "method", "name": "set", "returnType": "void", "children": [
        { "kind": "arg", "name": "v", "type": "const template<cpp:2974>&", "usr": "cpp:3158" } ] } ] }
```
Key self-check lines:
```
[PASS] TU contains exactly one WrappedTemplate: Holder
[PASS] no WrappedClass shadows Holder (implicit specialization never surfaced) — got [Shape, Scene, Palette]
[PASS] field 'T value': WrappedTemplateRef keyed to the param's usr (substitution contract)
[PASS] in-template `typedef T value_type`: WrappedTypedef -> WrappedTemplateRef(param usr)
[PASS] TU has exactly 8 children (the 3 enum DECLs contribute NO elements)
[PASS] Color constants mirror the oracle: RED=0, GREEN=5
[PASS] scoped `enum class Mode : unsigned char` arg: EXPLICIT underlying + A=0, B=1
[PASS] unscoped-explicit `enum Status : long`: underlying 'long', SIGNED value OK=-2
[PASS] nested enum spells QUALIFIED: 'Palette::Flavor' (fullyQualified rule), MILD+BOLD
cppfrontend: ALL SELF-CHECKS PASSED
```

## Verify
- Default unaffected: `./gradlew :krapper_gen:nativeTest :krapper_model:build :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL (`:krapper_model` untouched, no featuregen gate needed).
- Gated: `./gradlew :cppfrontend:runReleaseExecutableKlinker -PenableClang` — BUILD SUCCESSFUL, 72/72 self-checks PASS.

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -155,16 +155,19 @@ kplusplus {
         "clang::TemplateParameterList",
         "clang::TemplateTypeParmDecl",
         "clang::TemplateTypeParmType",
-        // NEW for brick 5 (enums): an enum-typed leaf reaches its EnumDecl through
-        // EnumType::getDecl() (the Type-side dyn-cast is bridged by TypeClass check —
-        // TypeBuilder); EnumDecl carries getIntegerType() (the underlying type,
+        // NEW for brick 5 (enums): an enum-typed leaf reaches its EnumDecl through the
+        // already-bound Type::getAsTagDecl() + a kind-gated re-view (TypeBuilder).
+        // clang::EnumType is deliberately NOT bound: it declares no constructors of its
+        // own (only inherited `using TagType::TagType`), so no ctor cursor ever marks
+        // hasConstructor and krapper synthesizes a `new clang::EnumType()` default-
+        // construct wrapper that C++ rejects (implicitly deleted — generator gap noted
+        // on #44). EnumDecl carries getIntegerType() (the underlying type,
         // clang_getEnumDeclIntegerType's source) and its DeclContext holds the
         // EnumConstantDecls. llvm::APSInt is EnumConstantDecl::getInitVal()'s carrier;
         // only its OWN surface is needed (getExtValue — see TypeBuilder's value bridge),
         // so the heavy llvm::APInt base stays unbound.
         "clang::EnumDecl",
         "clang::EnumConstantDecl",
-        "clang::EnumType",
         "llvm::APSInt"
     )
     // Materialize the range returns the walk iterates. Brick 3 walks members through
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -142,7 +142,30 @@ kplusplus {
         // getTemplateArgs() -> TemplateArgument::getAsType() per argument.
         "clang::ClassTemplateSpecializationDecl",
         "clang::TemplateArgument",
-        "clang::TemplateArgumentList"
+        "clang::TemplateArgumentList",
+        // NEW for brick 5 (template DECLARATIONS): a `template <typename T> class` is a
+        // ClassTemplateDecl whose getTemplatedDecl() is the pattern CXXRecordDecl;
+        // getTemplateParameters() lives on TemplateDecl (the primary base, reached by
+        // re-view — see ModelBuilder) and yields a TemplateParameterList (size/getParam).
+        // TemplateTypeParmDecl is each type param's decl; TemplateTypeParmType is the
+        // DEPENDENT `T` in member signatures, whose getDecl() keys the WrappedTemplateRef
+        // back to the matching WrappedTemplateParam.
+        "clang::TemplateDecl",
+        "clang::ClassTemplateDecl",
+        "clang::TemplateParameterList",
+        "clang::TemplateTypeParmDecl",
+        "clang::TemplateTypeParmType",
+        // NEW for brick 5 (enums): an enum-typed leaf reaches its EnumDecl through
+        // EnumType::getDecl() (the Type-side dyn-cast is bridged by TypeClass check —
+        // TypeBuilder); EnumDecl carries getIntegerType() (the underlying type,
+        // clang_getEnumDeclIntegerType's source) and its DeclContext holds the
+        // EnumConstantDecls. llvm::APSInt is EnumConstantDecl::getInitVal()'s carrier;
+        // only its OWN surface is needed (getExtValue — see TypeBuilder's value bridge),
+        // so the heavy llvm::APInt base stays unbound.
+        "clang::EnumDecl",
+        "clang::EnumConstantDecl",
+        "clang::EnumType",
+        "llvm::APSInt"
     )
     // Materialize the range returns the walk iterates. Brick 3 walks members through
     // DeclContext::decls() (source order, all member kinds) instead of methods(), so the
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -170,6 +170,23 @@ kplusplus {
         "clang::EnumConstantDecl",
         "llvm::APSInt"
     )
+    // FIXUPS (documented generator gaps, #44 brick 5): krapper's operator generation
+    // emits invalid Kotlin for four of llvm::APSInt's C++ operators —
+    //  * operator==(int64_t) / operator<(int64_t): the Long overloads of operators whose
+    //    APSInt overload already claimed the idiomatic name (equals/compareTo) are
+    //    emitted as `override fun _equals` / `override operator fun _compareTo`
+    //    (override-nothing + illegal operator name + pointer-vs-Long argument mixups);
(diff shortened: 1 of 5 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. All faithfulness claims independently verified against source.

**No-EnumDecl-branch verification (CONFIRMED):** ModelFactories.kt has no `CXCursor_EnumDecl` branch — its `else -> return null` at line 386 is the sole fallthrough. Enum payload correctly rides only on typed leaves via TypeFactories' `CXCursor_EnumDecl` branch (line 200), producing `WrappedEnumType` there and nowhere else.

**WrappedTemplateRef reading (CONFIRMED):** TypeFactories.kt line 172-174 confirms the libclang path produces `WrappedTemplateRef(referencedDecl.usr.toKString())` for `CXCursor_TemplateTypeParameter`. The PR mirrors this faithfully with `WrappedTemplateRef(decl.asDecl().cppUsr())` using the same `cpp:<canonical-id>` key.

**Shared-path refactor safety (CONFIRMED):** `addBasesAndMembers` correctly threads `cls.metadata` from `buildClass`, so the class construction path is semantically unchanged. The type widening from `WrappedClass` to `WrappedElement` + `ClassMetadata` is a clean extraction; all pre-existing checks still pass.

**APSInt fixups vs. PR #43 relation (DOCUMENTED, NOT A REGRESSION):** These are distinct krapper_gen limitations from what #43 addressed. PR #43 fixed collision between an operator and a same-name non-operator member (the `_get` case). The APSInt fixups address two different patterns: (a) multiple operator overloads producing the same idiomatic name (`APSInt` overload vs `int64_t` overload of `==`/`<` → collision-suffixed `_equals`/`_compareTo` residuals that #43 doesn't reach), and (b) nullable `inc()`/`dec()` return types violating Kotlin's operator convention. These are independent krapper_gen gaps, correctly documented in build.gradle.kts for a follow-up, and the `removeMethod` workaround is appropriate scaffolding since none of the four is needed for the value bridge.

**APSInt getExtValue vs. getSExtValue:** Reasoning is sound. `APSInt::getExtValue()` sign-extends by actual signedness; `getSExtValue()` (libclang's path) sign-extends through bit width unconditionally. They agree for all signed constants and all unsigned constants below the top bit; the divergence (unsigned constant with top bit set) is correctly identified as a Phase C normalizer entry.

**Build verification (independent):**
- `./gradlew :krapper_gen:nativeTest :krapper_model:build :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL ✓
- `./gradlew :cppfrontend:runReleaseExecutableKlinker -PenableClang` — BUILD SUCCESSFUL, **72/72 SELF-CHECKS PASSED** ✓

No CI checks configured (publish.yaml only). Both axes clean.

PR #54, opened 2026-06-10
cppfrontend: function-pointer typedefs + default-argument capture (#44 brick 6)

monkopedia-coder:
Phase B brick 6 of #44: the two deferred pieces that make the libclang TOKEN-SCRAPE deletable — function-pointer typedefs and default-argument capture in `cppfrontend/`.

## Function-pointer typedefs
`TypeBuilder.functionPointerTypedef` mirrors `TypeFactories.functionPointerTypedefElement` rule-for-rule:
- **typedef-only**: keyed on `Kind.Typedef` exactly as the libclang reducer is keyed on `CXCursor_TypedefDecl`; a `using` fn-ptr alias never becomes a `WrappedFunctionPointer` on either path (it collapses to the bare proto leaf here — the already-documented using-collapse divergence).
- pointer→proto shape checks = the `CXType_Pointer` + `CXType_FunctionProto` kind tests; the pointee is read **canonically** because the declarator nests `ParenType` sugar (`Pointer -> Paren -> FunctionProto`) that `clang_getPointeeType` desugars on the libclang side.
- stage-1 gate `isCFunctionPointerCompatible` copied verbatim (`isVoid || (isNative && !isString) || (isPointer && pointed.isVoid)`): Notifier's `void*` Mode-1 context slot passes, ShapeVisitor's `Shape*` fails and falls through to the alias collapse (CB-cfnptr-richsig).
- `cName` = unqualified spelling, `cppName` = `getQualifiedNameAsString()` (≡ the cursor `fullyQualified` walk) — exercised by namespace-scoped `geo::Predicate`.
- `FunctionProtoType` needed **no** classof bridge: unlike `TypedefType`, its chain survives resolution (generated `Type.asFunctionProtoType()` + inherited `getReturnType()` are real).

## Default arguments — the token-scrape deletion payoff
- `hasDefault` is now `ParmVarDecl::hasDefaultArg()` — the first-class Sema-computed fact, replacing `ModelFactories.defaultExpr`'s cursor-children heuristic. **The #36/#41 false-default workarounds are confirmed unnecessary on this path**: `hasDefaultArg()` cannot fire for an integer token inside the param's TYPE (array bound, non-type template arg, `function_ref` signature), so the model-side `typeCarriesFalseDefault` repairs have nothing to repair once the libclang front-end goes.
- `defaultValue` mirrors `ModelFactories.defaultValue`'s token-join contract via a **documented bridge**: an inline `kppbridge::defaultArgText` helper in the slice header wrapping `Lexer::getSourceText(CharSourceRange::getTokenRange(getDefaultArgRange()))`. Binding `clang::Lexer` + `clang::SourceManager` + `clang::LangOptions` wholesale (three very large classes for one static call, each a likely APSInt-style fixup-discovery source) is recorded as a deferred surface on #44.
- Verified value shapes, asserted equal to the libclang token-join strings: `5`, `-1`, `RED`, `nullptr`, `Palette()`, plus the multi-default trailing run `configure(int a, int b = 1, Shape* c = nullptr)` (whose tail satisfies `isOmittableDefault` — the trailing-defaulted-omit shortcut's contract).

## Phase C normalizer entries
- **default-value whitespace**: libclang joins tokens with `""`; `getSourceText` preserves inter-token whitespace. Identical for compactly-written defaults; a spaced default (`= Palette ( )`) diverges — normalize by stripping whitespace.
- **`_Bool` on the bare fn-ptr leaf**: extended the existing PrintingPolicy bridge (`spellingOf`) to fn-ptr proto spellings (`bool (*)(double)`), same divergence family as brick 3.
- (existing) using-alias collapse stays deterministic here vs order-dependent on the libclang path.

## Surface / fixture delta
- only(): + `clang::FunctionType`, `clang::FunctionProtoType`, `kppbridge::defaultArgText`. No new fixups. Slice header gains an include guard (the generated wrapper includes it through differently-spelled paths; the inline helper made that a same-TU redefinition).
- Fixture: + `Callback`/`Notifier`/`ShapeVisitor`/`HandlerFn`/`geo::Predicate` + `Dispatcher` (5 default shapes + trailing run); serializer carries `hasDefault`/`defaultValue` per arg.
- Self-checks: 72 → 91.
- `:krapper_model` untouched (WrappedFunctionPointer/WrappedArgument already carry the fields) — no featuregen gate needed.

## Verify
- Default unaffected: `:krapper_gen:nativeTest :krapper_model:build :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL.
- Gated: `:cppfrontend:runReleaseExecutableKlinker -PenableClang` — BUILD SUCCESSFUL, **ALL 91 SELF-CHECKS PASSED**.

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -171,9 +171,9 @@ kplusplus {
         "llvm::APSInt",
         // NEW for brick 6 (function-pointer typedefs): a typedef over a pointer-to-
         // function-proto decodes through FunctionProtoType — getNumParams()/getParamType()
-        // — with getReturnType() on FunctionType (its address-identical primary base;
-        // FunctionProtoType's own base list has the same multi-base CastTargets gap as
-        // TypedefType, bridged by classof + re-view in TypeBuilder).
+        // + the inherited getReturnType(). Unlike TypedefType, this chain survives
+        // resolution intact (the generated Type.asFunctionProtoType() dyn-cast is real);
+        // FunctionType is bound as the bridging base.
         "clang::FunctionType",
         "clang::FunctionProtoType",
         // NEW for brick 6 (default arguments): the smallest bindable surface for the
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -168,7 +168,19 @@ kplusplus {
         // so the heavy llvm::APInt base stays unbound.
         "clang::EnumDecl",
         "clang::EnumConstantDecl",
-        "llvm::APSInt"
+        "llvm::APSInt",
+        // NEW for brick 6 (function-pointer typedefs): a typedef over a pointer-to-
+        // function-proto decodes through FunctionProtoType — getNumParams()/getParamType()
+        // — with getReturnType() on FunctionType (its address-identical primary base;
+        // FunctionProtoType's own base list has the same multi-base CastTargets gap as
+        // TypedefType, bridged by classof + re-view in TypeBuilder).
+        "clang::FunctionType",
+        "clang::FunctionProtoType",
+        // NEW for brick 6 (default arguments): the smallest bindable surface for the
+        // default's source text — an inline helper in the slice header wrapping
+        // Lexer::getSourceText over ParmVarDecl::getDefaultArgRange() (see clang_slice.h
+        // for why Lexer/SourceManager/LangOptions aren't bound wholesale yet).
+        "kppbridge::defaultArgText"
     )
     // FIXUPS (documented generator gaps, #44 brick 5): krapper's operator generation
     // emits invalid Kotlin for four of llvm::APSInt's C++ operators —
--- a/cppfrontend/include/clang_slice.h
+++ b/cppfrontend/include/clang_slice.h
@@ -1,3 +1,11 @@
+// Include guard required: the generated wrapper includes this header both directly and
+// through each KrapperForce_*.h (via differently-spelled relative paths, so #pragma once
+// can't be trusted to dedupe) — without a guard the inline kppbridge helper below is a
+// same-TU redefinition. The clang/LLVM includes self-guard, which is why the header
+// didn't need one before it carried a definition of its own.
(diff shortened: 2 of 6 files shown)


monkopedia-reviewer (approved):
Tier 1 — auto-merge.

**isCFunctionPointerCompatible verbatim**: confirmed identical body between TypeFactories.kt (libclang path) and TypeBuilder.kt: `isVoid || (isNative && !isString) || (isPointer && pointed.isVoid)`.

**Default-value exact strings**: live run confirms all 5 shapes (`5`, `-1`, `RED`, `nullptr`, `Palette()`) match the libclang token-join contract exactly. The documented whitespace divergence for spaced defaults is a Phase C normalizer entry — not hidden, not papered over inside the builder. Honest.

**ParenType desugar**: `getCanonicalType()` is scoped to `functionPointerTypedef` only; regular pointer types in `buildWrappedType` still use non-canonical `getPointeeType()` (typedef alias preservation intact, brick-4 unaffected).

**kppbridge inline bridge**: correct. `getDefaultArgRange()` avoids the `getDefaultArg()` assert on unparsed/uninstantiated states; `CharSourceRange::getTokenRange()` wraps correctly; `.str()` copies the `StringRef` to owned storage — no lifetime issue.

**Include guard**: correct and well-explained. `#ifndef` is the right form when the header is included through differently-spelled relative paths (where `#pragma once` inode-dedup is unreliable).

**`isOmittableDefault` contract** (ModelResolution:611): `typeCarriesFalseDefault` returns false for all fixture arg types (`int`, `Shape*`); `configure`'s defaulted tail correctly reads as omittable.

**`:krapper_model` zero delta**: confirmed — no files in krapper_model/ changed.

**Self-checks 72→91**: confirmed by count and live run. No new fixups. No spurious `hasDefault` on pre-brick-6 args.

**Build results (local, JDK 17)**:
- `:krapper_gen:nativeTest :krapper_model:build :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL
- `:cppfrontend:runReleaseExecutableKlinker -PenableClang` — BUILD SUCCESSFUL, ALL 91 SELF-CHECKS PASSED

Issue #46, opened 2026-06-10
[clang-campaign] Phase D: featuregen byte-identical parity under --frontend=cpp (convergence tail)


monkopedia-coder:
The parity oracle for the front-end swap: `featuregen` (all 17 instantiations, 188 tests) must generate **byte-identical** output under `--frontend=cpp` vs `--frontend=libclang`. The existing golden infra IS the contract — no new spec needed.

This is the open-ended convergence phase: wire the gate (one brick), then iterate "run diff harness (Phase C) → fix first divergence as a targeted PR → re-run" until identical. Expect real-world C++ complexity to surface here (this is also where we learn how much of #10's broad-binding tail actually matters).

Exit criteria: featuregen byte-identical + krapper_gen nativeTest green under BOTH front-ends, twice-stable (determinism).

Blocked by #44 + #45. Followed by Phase E (#47): flip + delete libclang cinterop.

monkopedia-coder:
Phase D convergence-tail items from the brick-7 golden compare (PR #55 — the first real two-front-end tree-diff):

1. **Template-param default types** — the libclang path reconstructs `WrappedTemplateParam.defaultType` from cursor-walk type-residue children (`CXCursor_TypeRef`/`CXCursor_TemplateRef` elements attached by `ModelFactories.mapAll`, dropped by the compare's N3 normalizer); the cpp front-end doesn't build template-param defaults at all yet. The golden fixture has no defaulted template params, so the compare is currently blind here — needs a fixture row (`template <typename T = int> ...`) + cpp-side construction when the front-end grows there.

2. **Using-alias collapse determinism (N5)** — a `using` alias type-use is ORDER-DEPENDENT on the libclang path (`ResolverBuilderImpl.visit` leaves the alias-name leaf, e.g. `HandlerFn`), while the cpp path deterministically collapses to the canonical spelling (`int (*)(double)`). Normalized in the golden compare by a fixture-scoped alias table; the durable convergence (deterministic collapse on the libclang path, or alias preservation on the cpp path) deletes that table.

Also recorded (normalizer-permanent, no action needed): the libclang path keys dependent-T `WrappedTemplateRef`s on the param NAME via TypeFactories' `CXType_Unexposed`/`NoDeclFound` fallback rather than the USR branch; the resolver deliberately accepts both keys (`Parsing.typedAs` maps name AND usr), so the N2 positional-token rewrite is the honest, permanent bridge until the libclang front-end goes away.

[...]

Issue #45, opened 2026-06-10
[clang-campaign] Phase C: --frontend={libclang|cpp} flag + WrappedTU structural diff harness


monkopedia-coder:
Once Phase B (#44) produces a `WrappedTU` from the C++-AST front-end, make both front-ends selectable and comparable:

1. A `--frontend={libclang|cpp}` flag on krapper_gen (default `libclang`); both paths produce a `WrappedTU` and share resolve+codegen unchanged.
2. A structural WrappedTU tree-diff harness (element-by-element, identity-key-aware) that reports the first divergence with enough context to file a targeted fix — this is the debugging tool for Phase D convergence, so invest in good diff output.
3. CI/gating: the cpp front-end requires LLVM-22 + clang++, so it stays behind the existing `-PenableClang`-style gate until Phase E flips the default.

Blocked by #44. Followed by Phase D (#46): featuregen byte-identical parity under --frontend=cpp.

monkopedia-coder:
Finding from #44 brick 2 (PR #50): the model hierarchy is NOT @Serializable (only WrappedTypeReference is — parent back-links + non-data classes). The scaffold serialized via a frontend-side one-way DTO (`SerializedElement` in cppfrontend). For this ticket's tree-diff, BOTH front-ends must emit one canonical shape, so the durable home for that projection is `:krapper_model` (now standalone + ungated, so it's featuregen-byte-identical verifiable). Also note: identity fields use the documented `cpp:<canonical getID()>` convention on the cpp side vs libclang USRs — the diff must treat identity fields as maskable.

monkopedia-coder:
Brick 2 landed as PR #58 (THE HANDOFF). Scoped-out limitation, recorded precisely:

[...]

PR #55, opened 2026-06-11
THE GOLDEN TEST: both front-ends parse the same fixture, structural tree-diff under a cited normalizer ledger (#44 brick 7)

monkopedia-coder:
Phase B brick 7 of #44 — **THE GOLDEN TEST**: run BOTH front-ends (libclang-C and the cpp/generated-bindings one) on the same fixture and structurally compare their parse-output models. Phase B exit criterion / Phase C (#45) entry point, and the real automated regression lock for the front-end rewrite (the spirit of #8: any construction drift between ModelFactories/TypeFactories and ModelBuilder/TypeBuilder now fails a task; the clangwalk-specific test piece of #8 can be judged separately).

## What's in it

**(a) Canonical projection → `:krapper_model`** — `SerializedElement` + `WrappedElement.serialized()` moved from cppfrontend's local copy into `krapper_model/.../model/ModelSerializer.kt` (nativeMain — the projection extends the nativeMain `WrappedElement`). It is now THE canonical comparable shape both front-ends emit; cppfrontend's local copy is deleted.

**(b) krapper_gen `--dumpParsedModel <path>`** — follows the existing clikt option pattern in `App.kt`. Parse-only mode: the hook sits in `Parsing.kt parseHeader` **after the multi-file reduce completes the tree but BEFORE the pre-resolution rewrites** (`rewriteViewReturns`/`rewriteUniquePtrReturns` are krapper-specific marshalling baked onto the tree, not parse output — the cpp front-end doesn't perform them), writes the canonical-DTO JSON, and **exits** (documented on the flag + at the hook). Threaded via a CLI-scoped global rather than `KrapperConfig`, so the ksrpc service schema (`:slice`) is untouched by a debug-only flag. Flag absent ⇒ byte-identical behavior (featuregen gate below proves it).

**(c) The compare** — gated tasks in `cppfrontend/build.gradle.kts`:
- `goldenEmit` — the cppfrontend binary writes `fixture.h` (the binary emits the fixture itself, so both parses consume identical bytes) + `cpp.json`;
- `goldenDump` — the REAL `krapper_gen.kexe` parses that fixture with `--dumpParsedModel` → `libclang.json`;
- `goldenCompare` — tree-diff (`GoldenCompare.kt`) under the documented normalizer ledger; prints precise per-path divergence reports and exits non-zero on anything unexplained.

## Compare result

```
GOLDEN COMPARE: PASS — cpp front-end and libclang front-end parse output match
structurally under the documented normalizer ledger (N1-N5).
```

The first real two-front-end diff found **6 divergences beyond the documented list**, dispositioned per the brick's rules:

| Divergence | Disposition |
|---|---|
| `int [4]` vs `int[4]` (Scene.corners) | **FIXED** — cppfrontend construction bug: it pinned the OLD TypePrinter spelling from the krapper TestData golden ("int [5]"); live LLVM-22 libclang spells `int[4]`. TypeBuilder + self-check updated to the live oracle. |
| `template<cpp:N>` vs `template<T>` (4 sites in Holder: value_type typedef, `value` field, `get()` return, `set()` arg) | **NORMALIZED (N2)** — the libclang path keys dependent-T refs on the param NAME via TypeFactories' `CXType_Unexposed`/`NoDeclFound` fallback (which is what every dependent use inside a ClassTemplate's members actually hits), not the documented USR branch; the cpp path always keys identity. The resolver accepts BOTH keys (`Parsing.typedAs` maps name AND usr to the same substitution), so both forms rewrite to one positional token, scoped to the owning template. |
| `int (*)(double)` vs `HandlerFn` (Dispatcher.onHandle) | **NORMALIZED (N5)** — the documented using-alias collapse order-dependence: the libclang reducer leaves the alias-name leaf, the cpp path deterministically collapses to the canonical spelling. Fixture-scoped alias table; durable convergence filed as a Phase D item on #46. |

Negative test (manual): perturbing the libclang dump (`isAbstract` flip + a method rename) fails with exact paths — `tu/class[Shape].isAbstract: cpp=true libclang=false`, `tu/namespace[geo]/class[Circle]/method[radius].name: cpp=radius libclang=radiusX` — exit 1. The compare is not vacuous.

## Final mask/normalizer ledger (GoldenCompare.kt, each entry cites both sides)

- **N1 IDENTITY MASK** — usr fields: libclang USR strings vs `cpp:<canonical Decl id>` can never literally agree. Template-param usrs (cross-referenced by `WrappedTemplateRef` targets) rewrite to positional tokens (`tp:0`, document order — preserving the structural keying); arg usrs (referenced by nothing) mask to null.
- **N2 TEMPLATE-REF KEYING IN SPELLINGS** — `WrappedTemplateRef.toString()` embeds the ref key in type spellings; libclang keys on USR *or* bare name (the Unexposed/NoDeclFound fallback), cpp on identity; both rewrite to the positional token within the owning template's subtree (the resolver accepts both keys).
- **N3 CURSOR-WALK TYPE RESIDUE** — `ModelFactories.mapAll` attaches stray `CXCursor_TypeRef`/`CXCursor_TemplateRef` elements as children of methods/bases/typedefs; the cpp front-end constructs types only where the model references them. Non-model-kind children dropped from both sides. (Load-bearing ONLY under `WrappedTemplateParam.defaultType` — Phase D, see below.)
- **N4 DEFAULT-VALUE WHITESPACE** — libclang token-join (`joinToString("")`) vs cpp raw source text; whitespace stripped on both sides.
- **N5 USING-ALIAS COLLAPSE ORDER-DEPENDENCE** — alias-name leaf (libclang, order-dependent) vs deterministic canonical collapse (cpp); fixture-scoped alias table (`HandlerFn` → `int (*)(double)`).

(The two remaining documented Phase C divergences never fire on this projection: `_Bool`→`bool` is normalized at construction in TypeBuilder, and the APSInt unsigned-top-bit case isn't expressible in the DTO's type spellings.)

## Phase D convergence-tail items (filed on #46)

1. **Template-param default types** — the libclang path reconstructs `WrappedTemplateParam.defaultType` from the N3-dropped residue children; the cpp front-end doesn't build template-param defaults yet. The fixture has no defaulted template params, so the compare can't see it — needs a fixture row + construction when the front-end grows there.
2. **Using-alias collapse determinism (N5)** — converge by making the collapse deterministic (or preserving the alias) so the fixture-scoped table can be deleted.

## Verify

- FULL gate (krapper_model touched): `:krapper_gen:nativeTest` **311/0**, `:krapper_model:build` green, `:featuregen:kplusplusSync` + `:featuregen:nativeTest` **188/0** with **zero krapped/ drift** (byte-identical — git status clean after sync), `:krapper_gen:ktlintCheck` green.
- Gated: cppfrontend run **91/91 self-checks PASS**; `:cppfrontend:goldenCompare` **PASS** end-to-end at HEAD (emit → dump → compare).

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -205,3 +205,52 @@ kplusplus {
     instantiate("std::vector<clang::Decl*>")
     instantiate("std::vector<clang::CXXBaseSpecifier*>")
 }
+
+// ---- #44 brick 7: THE GOLDEN TEST (Phase B exit criterion / Phase C entry) ----
+// Run BOTH front-ends over the same fixture and structurally compare their parse-output
+// models through the canonical SerializedElement projection (:krapper_model):
+//   goldenEmit    — this front-end writes the fixture + its cpp.json (the fixture is
+//                   emitted by the binary itself so both parses consume identical bytes);
+//   goldenDump    — the REAL krapper_gen kexe parses that fixture with
+//                   --dumpParsedModel (parse-only mode, Parsing.kt) -> libclang.json;
+//   goldenCompare — the tree-diff under the documented normalizer ledger
+//                   (GoldenCompare.kt), non-zero exit on unexplained divergence.
+// This is the automated regression lock for the front-end rewrite (issue #8's spirit):
+// any construction drift between ModelFactories/TypeFactories and ModelBuilder/TypeBuilder
+// fails this task. Gated with the module (-PenableClang).
+val goldenDir = layout.buildDirectory.dir("golden").get().asFile
+val cppfrontendBinary = layout.buildDirectory.file("bin/klinker/cppfrontendRelease").get().asFile
+val krapperGenKexe = rootProject.layout.projectDirectory
+    .file("krapper_gen/build/bin/native/releaseExecutable/krapper_gen.kexe").asFile
+
+val goldenEmit = tasks.register<Exec>("goldenEmit") {
+    dependsOn("linkReleaseExecutableKlinker")
+    doFirst { goldenDir.mkdirs() }
+    commandLine(cppfrontendBinary.absolutePath, "--golden-emit", goldenDir.absolutePath)
+}
+
+val goldenDump = tasks.register<Exec>("goldenDump") {
+    dependsOn(goldenEmit, ":krapper_gen:linkReleaseExecutableNative")
+    commandLine(
+        krapperGenKexe.absolutePath,
+        "-h",
+        File(goldenDir, "fixture.h").absolutePath,
+        // cppfrontend's buildASTFromCode parses fixture.cc under the clang driver default;
+        // the fixture's surface is identical under c++14..c++20, pin c++17 to match the
+        // module's own binding standard.
+        "--std",
+        "c++17",
+        "--dumpParsedModel",
+        File(goldenDir, "libclang.json").absolutePath
+    )
+}
+
+tasks.register<Exec>("goldenCompare") {
+    dependsOn(goldenDump)
+    commandLine(
+        cppfrontendBinary.absolutePath,
+        "--golden-compare",
+        File(goldenDir, "cpp.json").absolutePath,
+        File(goldenDir, "libclang.json").absolutePath
+    )
+}
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -219,7 +219,8 @@ kplusplus {
 // any construction drift between ModelFactories/TypeFactories and ModelBuilder/TypeBuilder
 // fails this task. Gated with the module (-PenableClang).
 val goldenDir = layout.buildDirectory.dir("golden").get().asFile
-val cppfrontendBinary = layout.buildDirectory.file("bin/klinker/cppfrontendRelease").get().asFile
+val cppfrontendBinary = layout.buildDirectory
+    .file("bin/klinker/cppfrontendRelease/cppfrontend").get().asFile
 val krapperGenKexe = rootProject.layout.projectDirectory
     .file("krapper_gen/build/bin/native/releaseExecutable/krapper_gen.kexe").asFile
 
--- a/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
+++ b/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
@@ -480,13 +480,14 @@ fun main(args: Array<String>): Unit = memScoped {
         "got $setScaleArg"
     )
 
-    // Constant array: the libclang leaf spelling is "int [4]" (krapper_gen TestData
-    // golden "int [5]"); WrappedTypeReference parses element/extent off that name.
+    // Constant array: LLVM-22's libclang leaf spelling is "int[4]" — no space (pinned by
+    // the golden compare; the old TestData golden "int [5]" predates the TypePrinter
+    // change). WrappedTypeReference parses element/extent off that name either way.
     val corners = sceneFields.find { it.name == "corners" }?.type
     check(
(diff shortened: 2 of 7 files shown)


monkopedia-reviewer (approved):
Tier 1 — auto. Full review at merge standard. Both axes sound.

**Hook point + hermeticity (scrutiny item 1):** The `dumpParsedModelPath` hook in `Parsing.kt` lands exactly at the right seam — after `reduceRight` completes the multi-file TU, before `rewriteViewReturns`/`rewriteUniquePtrReturns`. Verified by direct reading of lines 554–588. Flag-absent path: `dumpParsedModelPath?.let { }` is a null-safe no-op — structurally identical to pre-PR when the flag is absent. Hermeticity: service mode returns (`if (serviceMode) return runService()`) BEFORE the `dumpParsedModelPath = dumpParsedModel` assignment in `run()`, so in service mode the global is always `null` and has exactly zero effect. This is the only new mutable top-level `var` in krapper_gen. No hermeticity regression.

**Featuregen byte-identical gate (scrutiny item 2):** `git status -- featuregen/krapped/` is clean after `kplusplusSync`. Zero drift. Independently verified.

**Normalizer ledger honesty (scrutiny item 3):**
- N1: narrow — only USRs; template-param USRs rewritten to positional tokens (resolver keying preserved), arg USRs masked. Honest.
- N2: narrow — only `"template<X>"` substrings within template subtrees. Honest — cites both sides' behavior (libclang CXType_Unexposed/NoDeclFound name-fallback vs cpp identity; resolver accepts both via `typedAs`).
- N3: drops non-MODEL_KINDS children. MODEL_KINDS is exhaustive for all modeled element types. Cannot mask a missing model element (all model kinds are in the set). Correctly only drops cursor-walk type residue (WrappedTemplateRef/WrappedType children of methods/bases). The "load-bearing under WrappedTemplateParam.defaultType" corner is correctly Phase D'd (#46) — the fixture has no defaulted template params, so N3 cannot hide real data. Not over-masking.
- N4: narrow — only defaultValue whitespace. N5: narrow — fixture-scoped alias table with one entry. Both honest.
- Negative test: perturbed dump exits 1 with exact paths. `goldenCompare()` calls `exitProcess(1)` on unexplained divergences; `Exec` task propagates non-zero exit as build failure. Compare is not vacuous.

**int\[4\] fix direction (scrutiny item 4):** The apparent tension resolves cleanly. LLVM-22's TypePrinter changed to drop the space (`int[4]`, not `int [4]`). The cppfrontend was constructing the OLD spelling from what the krapper TestData golden said. The golden compare found the divergence in the live run; fixed in TypeBuilder.kt. Now both sides emit `int[4]`. No normalizer needed — this was a real bug, now fixed. `WrappedTypeReference`'s `indexOf('[')` parser is space-insensitive so semantics were already correct; only the name string diverged.

**goldenCompare wiring (scrutiny item 5):** Task chain verified — `goldenEmit` writes `fixture.h` + `cpp.json`; `goldenDump` depends on both `goldenEmit` and `krapper_gen.kexe`, reads `fixture.h`, writes `libclang.json`; `goldenCompare` depends on `goldenDump`. One-source guarantee holds (same bytes, same dir). Exit propagation correct.

**:slice service schema:** `dumpParsedModel` is a Clikt option on `KrapperGen : CliktCommand()`, not in `KrapperConfig` (which is `@Serializable` commonMain). Schema untouched.

**Independent build results:**
- Full gate: `:krapper_gen:nativeTest` 311/0, `:krapper_model:build` green, `:featuregen:kplusplusSync` + `:featuregen:nativeTest` 188/0, zero krapped/ drift, `:krapper_gen:ktlintCheck` — BUILD SUCCESSFUL.
- Gated: 91/91 self-checks PASS; `:cppfrontend:goldenCompare` PASS end-to-end (emit → dump → compare).

No CI workflows on PRs in this repo (only `publish.yaml` on release). Local build is the gate.

PR #56, opened 2026-06-11
krapper_model: full-fidelity WrappedTU round-trip (DAG-aware ModelIo) + --roundTripModel oracle — byte-identical featuregen sync (#45 brick 1)

monkopedia-coder:
Phase C brick 1 of #45: full-fidelity `WrappedTU` round-trip serialization — the foundation of the `--frontend=cpp` handoff. Serialize a parsed WrappedTU to JSON, deserialize it back, and the pipeline produces **byte-identical** output.

## THE ORACLE RESULT

`featuregen kplusplusSync` run with the round-trip ON is **byte-identical** to the committed flag-off output, across the header import + all 17 instantiations (string/vector/map/set/pair/unordered/unique_ptr/Box/Pair2 — 18 parses, largest per-parse payload 1.8 MB JSON):

```
# via gradle (env var survives the single-use daemon):
KRAPPER_ROUNDTRIP_MODEL=1 ./gradlew --no-daemon :featuregen:kplusplusSync
# engagement-proof run (gradle 9 swallows the spawned kexe's stdout), same args the
# plugin passes, logging 18/18 "Round-tripped parsed model through ModelIo JSON":
KRAPPER_ROUNDTRIP_MODEL=1 krapper_gen.kexe featuregen -r INCLUDE_MISSING ... --header strings_feature.h --instantiate ... -o featuregen/krapped
diff -r <flag-off snapshot> featuregen/krapped   # -> empty, byte-identical
```

Also verified: flag-OFF output is byte-identical to main's (the schema additions alone change nothing).

## What the oracle caught (and the design that fixes it)

The first oracle run drifted massively — and not from a missing field: **instance identity is part of the model.** The parse output is a DAG (one element under two parents; interned/memoized types shared by every use site), elements are memoized ACROSS parses (`elementLookup`), and resolution MUTATES them in place (metadata flags, synthetic sizeOf/alignOf children, const-overload dedup, allocationStyle). A naive value-tree encode duplicated aliases and severed cross-parse identity (visible as runaway `_..._size_of` duplicate accumulation differences and a Drawable dispose delta). Decisions:

- **DAG-aware encoding**: pass 1 collects identity-shared nodes; the first occurrence is wrapped in `DefDto(id)`, later occurrences collapse to `RefDto(id)` back-references; decode rebuilds the exact aliasing. Encode/decode share one canonical traversal (fields, then children) so definitions always precede references.
- **Cross-parse identity (oracle harness only)**: `WrappedElement.remapMemoized` lockstep-walks original/restored trees (`forEachIdentityPair`) and points the cursor→element memo at the restored instances, so later parses accumulate state on the same objects resolution sees — exactly like a non-flag run. The production `--frontend=cpp` handoff will need stable cross-TU identities in the format itself (the USR→stable-Decl-key probe, next on #45).
- **Parent back-links**: not serialized; rebuilt bottom-up via `addAllChildren` (which re-parents every child), so the cycle never reaches the encoder.
- **DTO mirror, not `@Serializable` on the model**: the element classes carry a private base-constructor children list, parent links, non-constructor mutable state and caches; the mirror keeps the model untouched (flag-off provably unchanged) and doubles as the explicit, reviewable schema. Every field the ModelResolution.kt when-dispatch consumes is carried (incl. method flags, ClassMetadata, allocationStyle, template merge counter + `otherParams`, argument defaults/USRs, the full WrappedType hierarchy with enum constants and function-pointer cName/cppName).
- **Type interning**: `GenerationContext.internedTypes` only ever caches plain `WrappedTypeReference`s (value-equal data class) — referential identity of types is not load-bearing for equality, and within-model type aliasing is preserved by the def/ref encoding, so decode doesn't touch the intern cache.
- **Hook placement**: same point as `--dumpParsedModel` (post-reduce, pre-rewrites) — the rewrites replace method children with copies, so they must run on the restored tree to keep the memo and live tree aligned.

## Wiring

- `:krapper_model` `ModelIo.kt`: schema + `ModelIo.encodeToString` / `ModelIo.decodeFromString` — the deserialize entry point the future `--frontend=cpp` path consumes.
- `krapper_gen --roundTripModel` flag, or `KRAPPER_ROUNDTRIP_MODEL=1` env (reaches service-mode/plugin-spawned runs).
- `ModelIoTest` (new, 2 tests): hand-built model covering every node kind — JSON-stability + state/parents/aliasing restoration.

## Verify

- `:krapper_gen:nativeTest` 311/0, `:featuregen:nativeTest` 188/0, `:krapper_model:build` (+2 new tests), `:featuregen:kplusplusSync`, `:krapper_gen:ktlintCheck` — all green (JDK 17).
- Flag-off krapped/ byte-identical to main; flag-on byte-identical with 18/18 parses round-tripped.

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
@@ -165,6 +165,15 @@ class KrapperGen : CliktCommand() {
             "mode). :cppfrontend:goldenCompare diffs this libclang front-end's parse " +
             "output against the C++-AST front-end's."
     )
+    val roundTripModel by option(
+        "--roundTripModel",
+        help = "Debug/oracle flag (#45 brick 1): after each parse, serialize the full " +
+            "WrappedTU through the ModelIo round-trip JSON, deserialize it back, and run " +
+            "resolution+generation on the DESERIALIZED model. Output should be " +
+            "byte-identical to a run without the flag — the verification gate for the " +
+            "--frontend=cpp handoff format. Also enabled by KRAPPER_ROUNDTRIP_MODEL=1 " +
+            "in the environment (which reaches service-mode runs too)."
+    ).flag()
     val fixupFile by option(
         "--fixup-file",
         help = "Path to a JSON file containing a list of Fixup directives (see Fixup.kt). " +
@@ -181,6 +190,7 @@ class KrapperGen : CliktCommand() {
         // process exits inside parseHeader right after the JSON is written, so the
         // resolution/generation flow below never runs with the flag set.
         dumpParsedModelPath = dumpParsedModel
+        if (roundTripModel) roundTripParsedModel = true
         runBlocking {
             val service = KrapperServiceImpl()
             val resolvedModule = moduleName
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelFactories.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelFactories.kt
@@ -29,7 +29,6 @@ import clang.CX_CXXAccessSpecifier
 import com.monkopedia.krapper.generator.model.ElementIdentity
 import com.monkopedia.krapper.generator.model.MethodType
 import com.monkopedia.krapper.generator.model.RangeReturn
-import com.monkopedia.krapper.generator.model.forEachIdentityPair
 import com.monkopedia.krapper.generator.model.WrappedArgument
 import com.monkopedia.krapper.generator.model.WrappedBase
 import com.monkopedia.krapper.generator.model.WrappedClass
@@ -43,6 +42,7 @@ import com.monkopedia.krapper.generator.model.WrappedTU
 import com.monkopedia.krapper.generator.model.WrappedTemplate
 import com.monkopedia.krapper.generator.model.WrappedTemplateParam
 import com.monkopedia.krapper.generator.model.WrappedTypedef
+import com.monkopedia.krapper.generator.model.forEachIdentityPair
 import com.monkopedia.krapper.generator.model.type.WrappedTemplateRef
 import com.monkopedia.krapper.generator.model.type.WrappedTemplateType
 import com.monkopedia.krapper.generator.model.type.WrappedType
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelFactories.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelFactories.kt
@@ -26,8 +26,10 @@ import clang.CXCursorKind.CXCursor_TypedefDecl
 import clang.CXType
 import clang.CXTypeKind.CXType_Pointer
 import clang.CX_CXXAccessSpecifier
+import com.monkopedia.krapper.generator.model.ElementIdentity
 import com.monkopedia.krapper.generator.model.MethodType
 import com.monkopedia.krapper.generator.model.RangeReturn
+import com.monkopedia.krapper.generator.model.forEachIdentityPair
 import com.monkopedia.krapper.generator.model.WrappedArgument
 import com.monkopedia.krapper.generator.model.WrappedBase
 import com.monkopedia.krapper.generator.model.WrappedClass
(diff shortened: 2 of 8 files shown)


monkopedia-reviewer (approved):
## Review: krapper_model WrappedTU round-trip (DAG-aware ModelIo) — APPROVED

**Category:** `feature_internal` — new serialization infrastructure in :krapper_model, flag-gated oracle in :krapper_gen; no public API change.
**Tier:** 1 (kplusplus baseline `* : tier 1`).

---

### Correctness

**Flag-off purity ✓**  
Model classes are untouched as behavior-bearing definitions. The only changes to existing model files are `@Serializable` on `ClassMetadata` (data class, no runtime behavior change) and `WrappedEnumConstant` (same). Every DTO is a new, separate type mirror. The `roundTripParsedModel` global var follows the same CLI-scoped-global idiom as `dumpParsedModelPath` (flagged in the same comment block). The env-var initialization path (`KRAPPER_ROUNDTRIP_MODEL=1`) reaching service-mode runs is intentional and documented.

**DAG encode/decode ✓**  
The Def/Ref scheme is sound. IDs are assigned first-seen (deterministic). Canonical traversal (modelFields → children) shared by encode, decode, and the identity zip guarantees defs always precede refs — no forward-reference ambiguity. Cycle protection (ID assigned before recursing) degrades cleanly. The `ModelEncoder` `element()→DefDto(id, raw(element))` pattern correctly handles nested shared sub-elements inside a def's fields.

**Interning claim ✓**  
Confirmed by reading `WrappedType.invoke`: the `getOrPut` lambda uses non-local `return` (the function is `inline`) for `const `, `typename `, `*`, `&`, and empty cases — those paths return WITHOUT storing in `GenerationContext.internedTypes`. Only plain `WrappedTypeReference(type)` is ever put in the intern cache, keyed by string. `WrappedModifiedType`, `WrappedTemplateType`, `WrappedEnumType`, `WrappedFunctionPointer` are NOT interned — no referential identity assumption on types other than plain refs. The DAG encoding handles the type aliasing that IS load-bearing (shared WrappedTypeReference instances encoded def/ref) correctly.

**remapMemoized ✓**  
Called only inside the `if (roundTripParsedModel)` branch in Parsing.kt. Normal operation never touches it. The lockstep `forEachIdentityPair` walk is correct: visited-set on original identity prevents double-processing; the shape-mismatch `require` catches any round-trip schema gap.

**WrappedTemplate.addAllChildren ✓**  
The base-class `addAllChildren` bypasses `WrappedTemplate.addChild` (the param-merge hook), which is correct here: the merged params are already in `otherParams` (serialized separately in `TemplateParamDto.otherParams`) and the children list contains only the primary params. No double-merge.

**Identity comparisons ✓**  
All `===` uses in the production parse path (`Parsing.kt:933`, `KotlinWriter.kt:2372`) are on `WrappedMethod`/`ResolvedMethod` instances — not types. These fire after `remapMemoized` has updated the elementLookup memo, so they see the restored instances. Correct.

**ModelIoTest ✓**  
Both tests are meaningful:  
- `roundTrip_isStable`: JSON re-encoding of the restored tree is bit-identical — catches any field silently not surviving.  
- `roundTrip_restoresStateAndParents`: Exercises every node kind; checks all mutable fields (isVirtual, isConst, returnsPairSecond, etc.); verifies parent back-links; and critically `assertSame(args[0].type, enum.underlying)` pins the DAG aliasing (two `int` uses must decode as one instance).

---

### Independent build (JDK 17, no loops)

**Full gate (`--no-daemon`):**  
- `:krapper_gen:nativeTest` — **311/0** ✓  
- `:krapper_model:nativeTest` (ModelIoTest) — **2/0** ✓ (new)  
- `:featuregen:nativeTest` — **188/0** ✓  
- `:featuregen:kplusplusSync` (flag OFF) — **zero krapped/ drift** ✓  
- `:krapper_gen:ktlintCheck` — **clean** ✓  

**Oracle (`KRAPPER_ROUNDTRIP_MODEL=1`):**  
`KRAPPER_ROUNDTRIP_MODEL=1 ./gradlew :featuregen:kplusplusSync --no-daemon` — **zero krapped/ drift** ✓  
(Note: Gradle 9 swallows spawned kexe stdout, so the 'Round-tripped…' log lines don't surface through Gradle — per the PR description this is expected; the byte-identical output is the oracle.)

**timespec_size_of 324×:** Pre-existing — `git diff main HEAD -- featuregen/krapped/` is empty; the 324× occurrences in featuregen.h/featuregen.cc are identical in main. Should be filed as a separate bug (cross-instantiation accumulation) but does not touch this PR.

---

### Minor non-blocking note

`ModelIo.json` is `public` (no visibility modifier). The tests don't use it directly; making it `private` or `internal` would be tidier. Trivial toss-up — not blocking.

---

Merging.

PR #58, opened 2026-06-11
krapper_gen --frontend=cpp: THE HANDOFF — resolve+codegen on a cppfrontend-produced WrappedTU (#45 brick 2)

monkopedia-coder:
#45 Phase C brick 2 — THE HANDOFF: krapper_gen consumes a WrappedTU produced by the cpp front-end and runs its real pipeline on it. First time generated-bindings-parsed C++ flows through resolve+codegen.

## Pipeline

`cppfrontend (Clang C++ AST on generated bindings) → model.json (full-fidelity ModelIo JSON) → krapper_gen --frontend=cpp --parsedModel (libclang parse SKIPPED) → resolve → codegen → clang++ compiles the wrapper`

## What changed

- **cppfrontend**: `--golden-emit` now also writes `model.json` = `ModelIo.encodeToString(tu)` (the handoff format), alongside the lossy compare projection. ModelIo's DAG-aware Def/Ref encoding handles any instance sharing the same way it does libclang-interned types — no cppfrontend/ModelIo gaps surfaced; nothing needed fixing en route.
- **krapper_gen**: `--frontend={libclang|cpp}` (default libclang, default behavior bit-for-bit unchanged) + `--parsedModel <path>` (required with cpp). The hook sits at the top of `parseHeader`: decode → `rewriteViewReturns`/`rewriteUniquePtrReturns` on the loaded tree → `ParsedResolver` — the exact post-parse entry point the `--roundTripModel` oracle (brick 1) proved. CLI-scoped global like `dumpParsedModelPath`; ksrpc service schema untouched. libclang stays LINKED, just never CALLED on this path (flip/deletion is Phase E).
- **`:cppfrontend:handoffGenerate`** (gated, `-PenableClang`): the end-to-end proof. krapper_gen runs to completion on the cpp-front-end model and `writeTo`'s CppCompiler step compiles the generated `.cc` against `fixture.h` with clang++ (non-zero exit fails the task); `doLast` asserts the artifacts.

## Generated output

`fixture.h`, `fixture.cc`, `fixture.def`, `libfixture.a` (compiled), + 12 Kotlin files: `root_Shape.kt`, `root_ShapeApi.kt`, `geo_Circle.kt`, `geo_Functions.kt`, `root_Scene.kt`, `root_Palette.kt`, `palette_Flavor.kt`, `root_Color.kt`, `root_Mode.kt`, `root_Status.kt`, `root_Dispatcher.kt`, `CppBinding.kt`.

```kotlin
@krapper.CppBinding("geo::Circle")
class Circle(
    override val ptr: COpaquePointer,
    val memScope: MemScope,
) : root.ShapeApi {
    inline fun name(): String? {
        val str: CPointer<ByteVar>? = Shape_name(ptr)
        ...
```

**Cross-front-end identity check** (not just plausibility): ran the SAME fixture through the default libclang front-end and diffed the outputs — every generated Kotlin file and the wrapper sources are **byte-identical** modulo output-dir path artifacts (the `#include` relative path and the `.def`'s `-I`/`libraryPaths`).

## What dropped (all expected, ledger-cited)

Identical drop set on both front-ends (5 bound / 6 dropped):
- `corners: int[4]`, `shapes: Holder<Shape*>` — field-type drops; `Holder<T>` has no instantiation (IGNORE_MISSING) and array fields aren't marshalable. Same drops on the libclang path.
- `visit(void (*)(Shape *))` — CB-cfnptr-richsig: class-pointer param fails the stage-1 compat gate (both front-ends).
- `onHandle(...)` — the `using HandlerFn` alias; drops on both, spelled `int (*)(double)` (cpp, deterministic collapse) vs `HandlerFn` (libclang, order-dependent alias leaf) — the documented N-ledger using-collapse divergence, same semantics.

## Scoped out

`--instantiate` with `--frontend=cpp` is **rejected with a clear error**: instantiation forcing (`IndexedServiceImpl.requestInstantiation`) synthesizes a forcing header and re-parses it through libclang — entangled with the in-process parse + the cursor→element memo. The fixture needs no instantiations; this is the known next chunk on #45.

## Verify

- Full gate: krapper_gen `nativeTest` 311/0, featuregen `kplusplusSync` + `nativeTest` 188/0, `:krapper_model:build` green, `ktlintCheck` green, ZERO `krapped/` drift (flags default-off change nothing).
- Gated: cppfrontend 91 self-checks PASS, `goldenCompare` PASS, `handoffGenerate` green end-to-end (kexe completes, C++ compiles).

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -255,3 +255,50 @@ tasks.register<Exec>("goldenCompare") {
         File(goldenDir, "libclang.json").absolutePath
     )
 }
+
+// ---- #45 brick 2: THE HANDOFF (Phase C) ----
+// The first time generated-bindings-parsed C++ flows through krapper_gen's REAL pipeline:
+//   goldenEmit       — the cppfrontend binary parses the fixture with the Clang C++ AST
+//                      and writes model.json (full-fidelity ModelIo JSON) next to the
+//                      compare projection;
+//   handoffGenerate  — krapper_gen --frontend=cpp loads that model (libclang parse
+//                      SKIPPED), runs resolution + codegen, emits the Kotlin bindings +
+//                      C++ wrapper for the fixture, and COMPILES the wrapper against
+//                      fixture.h (writeTo's CppCompiler step — clang++ -c; a non-zero
+//                      exit fails the task), proving the handed-off model carries
+//                      everything the real pipeline consumes.
+// --instantiate is scoped out on this path (forcing re-parses synthesized headers
+// through libclang); the fixture needs no instantiations. Gated with the module
+// (-PenableClang), like the golden tasks above.
+val handoffDir = layout.buildDirectory.dir("handoff").get().asFile
+
+tasks.register<Exec>("handoffGenerate") {
+    dependsOn(goldenEmit, ":krapper_gen:linkReleaseExecutableNative")
+    doFirst { handoffDir.mkdirs() }
+    commandLine(
+        krapperGenKexe.absolutePath,
+        "-h",
+        File(goldenDir, "fixture.h").absolutePath,
+        // Same standard pin as goldenDump: the wrapper compile must match the parse.
+        "--std",
+        "c++17",
+        "--frontend",
+        "cpp",
+        "--parsedModel",
+        File(goldenDir, "model.json").absolutePath,
+        "-o",
+        handoffDir.absolutePath,
+        "fixture"
+    )
+    doLast {
+        // The run already failed on any resolve/codegen/compile error; assert the
+        // artifacts the next phase consumes actually materialized.
+        val expected = listOf("fixture.h", "fixture.cc", "fixture.def", "libfixture.a")
+        val missing = expected.filter { !File(handoffDir, it).exists() }
+        check(missing.isEmpty()) { "handoffGenerate: missing generated artifact(s): $missing" }
+        val kotlinFiles = File(handoffDir, "src").walkTopDown()
+            .filter { it.isFile && it.extension == "kt" }.toList()
+        check(kotlinFiles.isNotEmpty()) { "handoffGenerate: no Kotlin bindings generated" }
+        println("handoffGenerate: generated " + (expected + kotlinFiles.map { it.name }))
+    }
+}
--- a/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
+++ b/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 import clang.tooling.buildASTFromCode
+import com.monkopedia.krapper.generator.model.ModelIo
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. THE HANDOFF — clean. Build verified independently.

**Default-path purity:** `cppParsedModelPath?.let { ... }` guards the early-return; null (default) means the guard never fires and the libclang path is bit-for-bit unchanged. `featuregen:kplusplusSync` re-gen was zero-drift (17 instantiations, same as before); `nativeTest` UP-TO-DATE on both modules. `ktlintCheck` green.

**Entry point:** cpp path enters at the correct post-parse point — `rewriteViewReturns` + `rewriteUniquePtrReturns` run on the decoded tree, then `ParsedResolver` wraps it — matching the `--roundTripModel` oracle.

**Include plumbing:** `-h fixture.h` populates `IndexRequest.headers`; `writeTo` uses `request.headers` for the `#include` list in the generated C++ wrapper. This comes from the CLI, not the parse — clean on both paths.

**Byte-identical claim — verified independently:** Ran the same fixture through the libclang front-end (`krapper_gen.kexe -h fixture.h --std c++17`) and diffed against the `handoffGenerate` output. All **12 Kotlin files are byte-identical**. `fixture.h` is identical. `fixture.cc` non-include lines are identical. `fixture.def` differs only in the `-I<output-dir>` / `libraryPaths = <output-dir>` path artifacts (the two directories are different locations on disk). The second `#include` in `fixture.cc` is the relative path from output-dir to the golden fixture.h — varies correctly by output-dir. Nothing beyond path artifacts.

**Drop ledger:** Both paths bound 5, dropped 6. The using-alias divergence (`int (*)(double)` cpp vs `HandlerFn` libclang) is correctly documented as the N5 using-collapse difference — same semantics, different spelling.

**`--instantiate` rejection:** `require(instantiate.isEmpty())` fires a clear `IllegalArgumentException` with the full entanglement description. The crash+stacktrace output is consistent with the existing `error()` call pattern in App.kt (line 241: "A module name argument is required" produces the same).

**Gradle task wiring:** `handoffGenerate` depends on `goldenEmit` (task reference, not string) + `:krapper_gen:linkReleaseExecutableNative`. Gradle `Exec` fails on non-zero exit by default. `doFirst`/`doLast` placement correct. Module fully gated by `-PenableClang` at the `settings.gradle.kts` level — the task doesn't exist in a default build.

**`cppParsedModelPath` global:** consistent with existing `dumpParsedModelPath` / `roundTripParsedModel` in the same file.

**Build results (independent run, JDK 17):**
- Full gate: `krapper_gen:nativeTest` + `krapper_model:build` + `featuregen:kplusplusSync` + `featuregen:nativeTest` + `krapper_gen:ktlintCheck` → BUILD SUCCESSFUL
- Gated: `goldenEmit` (91 self-checks PASS) + `goldenCompare` (PASS) + `handoffGenerate` (PASS — C++ compiled, 16 artifacts generated)

Approving. Tier 1.

PR #59, opened 2026-06-11
Resolver: idempotent synthetic sizeOf/alignOf injection across re-resolutions (#57)

### LINKED ISSUE #57: [bug] Cross-instantiation accumulation: timespec_size_of emitted 324x with ~330-underscore name mangling in featuregen krapped/
Discovered by PR #56's round-trip oracle (and verified pre-existing on main — `git diff main -- featuregen/krapped/` empty): the committed featuregen `krapped/` output contains `timespec_size_of` emitted **324 times**, with progressively-growing underscore-mangled names (~330 underscores at the worst). This is a cross-instantiation accumulation: each of the 17 `requestInstantiation` passes appears to re-add a synthetic sizeOf/alignOf for the same class with a freshly-uniquified name instead of recognizing the existing one.

Benign today (byte-identical/deterministic, featuregen 188/0 green — it's bloat, not corruption), but it:
- inflates the generated bindings + def files,
- is exactly the kind of duplicate-accumulation the resolveForcing eviction work (#10/#11 family) was cleaning up,
- will look like a parity diff under Phase D if the cpp front-end doesn't reproduce the same accumulation order.

Likely suspects: the synthetic sizeOf/alignOf injection during resolution + the last-wins merge across `requestInstantiation` calls (cf. PR #40's cumulative forcing — this may be its remaining sibling on the SYNTHETIC-member side rather than the range-accessor side).

monkopedia-coder:
Fixes #57.

## Root cause
`WrappedClass.modifyMethodsIfNeeded` (ModelResolution.kt) runs on the **shared WrappedClass** on every resolve and unconditionally `addChild`ed a fresh synthetic sizeOf/alignOf pair. Each `requestInstantiation` re-resolves every already-bound class **twice** (`resolveForcing` pass 1 + pass 3), so the class accumulated +2 pairs per request. The per-pass `NameHandler` then `_`-uniquified the duplicate names instead of collapsing them. After featuregen's 17 instantiation requests, the per-pass emitted sections carried 1, 3, 5, ..., 35 copies — `timespec_size_of` appeared **324 times (= 18²)** in `featuregen.h`/`featuregen.cc`, the worst name with 34 leading underscores. Every bound class accumulated the same way: 48,890 synthetic `sizeof`/`alignof` definitions total in the pre-fix `.cc`.

## Fix shape
Check-before-add at the injection site: the pair is added only if the class doesn't already carry a `SIZE_OF` method. Synthetic injection is now idempotent across re-resolutions, so every re-resolved copy carries exactly **one clean-named pair** (`timespec_size_of`, zero extra underscores), regardless of instantiation count or order.

## Generated-output characterization (krapped/ is gitignored, so proven by regeneration diff)
Regenerated `featuregen/krapped/` at the parent commit vs this branch:
- `timespec_size_of`: 324 → 18 occurrences (one clean decl per per-pass section); total synthetic bodies in `.cc`: 48,890 → 2,888 across all 101 emitted classes (exactly one sizeOf+alignOf per class-section).
- `featuregen.h`: 138,135 → 46,131 lines; `featuregen.cc`: 277,491 → 93,483 lines.
- **Nothing else changed**: the non-synthetic line multiset of `.h` is identical pre vs post; the `.cc` multiset delta contains only the synthetic definitions themselves (`*_size_of`/`*_align_of` signatures, `return sizeof/alignof(X);` bodies, closing braces) — zero other lines. The generated Kotlin `src/`, `.def`, and forcing headers are byte-identical.
- Remaining `_`-prefixed names (`_IO_FILE_size_of`, `_pthread_cleanup_buffer_size_of`) are real class names, not uniquifier growth.

## Stability / determinism
`:featuregen:kplusplusSync` run three times on the fix (including after a pre-fix regeneration in between): all three outputs byte-identical (`diff -r` clean), so the result is re-run-stable and instantiation-order-independent.

## Verification
- `:krapper_gen:nativeTest` 312/0 (311 + new test), `:krapper_model:build`, `:featuregen:kplusplusSync`, `:featuregen:nativeTest` 188/0, `:krapper_gen:ktlintCheck` — all green (JDK 17).
- New test `ReturnShapeTest.repeatedForcingDoesNotDuplicateSyntheticMembers`: drives `resolveForcing` twice over the same bound class (mirroring the #40 multi-forcing pattern) and asserts the last-wins copy carries exactly one pair named `Holder_size_of`/`Holder_align_of` (pre-fix: 5 pairs with underscore growth).
- Gated sanity: `:cppfrontend:goldenCompare -PenableClang` passes (golden fixture has no instantiations — untouched, as expected).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelResolution.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelResolution.kt
@@ -299,20 +299,28 @@ private fun WrappedClass.modifyMethodsIfNeeded(baseClasses: List<WrappedClass>) 
             removeChild(it)
         }
     }
-    addChild(
-        WrappedMethod(
-            "sizeOf",
-            WrappedTypeReference("int"),
-            MethodType.SIZE_OF
+    // Idempotent: this runs on the SHARED WrappedClass on EVERY resolve, and a class is
+    // re-resolved once per requestInstantiation pass (resolveForcing pass 1 + pass 3).
+    // Unconditionally appending here accumulated a fresh sizeOf/alignOf pair per pass,
+    // which the per-pass NameHandler then `_`-uniquified — e.g. `timespec_size_of`
+    // emitted 324 times with up to 34 leading underscores after featuregen's 17
+    // instantiation requests (#57). Add the pair only once per class.
+    if (children.none { (it as? WrappedMethod)?.methodType == MethodType.SIZE_OF }) {
+        addChild(
+            WrappedMethod(
+                "sizeOf",
+                WrappedTypeReference("int"),
+                MethodType.SIZE_OF
+            )
         )
-    )
-    addChild(
-        WrappedMethod(
-            "alignOf",
-            WrappedTypeReference("int"),
-            MethodType.ALIGN_OF
+        addChild(
+            WrappedMethod(
+                "alignOf",
+                WrappedTypeReference("int"),
+                MethodType.ALIGN_OF
+            )
         )
-    )
+    }
 }
 
 suspend fun WrappedElement.createThisArg(resolverContext: ResolveContext): ResolvedArgument? {
--- a/krapper_gen/src/nativeTest/kotlin/ReturnShapeTest.kt
+++ b/krapper_gen/src/nativeTest/kotlin/ReturnShapeTest.kt
@@ -22,6 +22,7 @@ import com.monkopedia.krapper.generator.codegen.CppWriter
 import com.monkopedia.krapper.generator.codegen.File
 import com.monkopedia.krapper.generator.model.WrappedClass
 import com.monkopedia.krapper.generator.resolvedmodel.ArgumentCastMode
+import com.monkopedia.krapper.generator.resolvedmodel.MethodType
 import com.monkopedia.krapper.generator.resolvedmodel.ResolvedClass
 import com.monkopedia.krapper.generator.resolvedmodel.ResolvedElement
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Tier 1 — auto-merge. Reviewed to main-merge quality.

**Guard correctness:** The check-before-add keys on `(it as? WrappedMethod)?.methodType == MethodType.SIZE_OF` — an enum field set at construction, not the method name. A user-defined C++ method named "sizeOf" carries `MethodType.METHOD` (the default), not `MethodType.SIZE_OF`, so no false-positive is possible. SIZE_OF and ALIGN_OF are added atomically inside the same `if` block, so the SIZE_OF sentinel is sufficient. First resolution correctly adds the pair (guard passes on empty children); re-resolution skips correctly (SIZE_OF already present on the shared object).

**Change characterization (independently verified):**
- `timespec_size_of` in featuregen.cc: 18 occurrences ✓ (matches claim; one clean decl per section)
- featuregen.cc: 93,483 lines ✓; featuregen.h: 46,131 lines ✓
- Total synthetic bodies: 2,888 ✓
- Stability: second sync run is byte-identical to first ✓

**Test:** `repeatedForcingDoesNotDuplicateSyntheticMembers` drives the real `resolveForcing` path twice (not a unit shortcut — uses the same function featuregen's `requestInstantiation` calls), asserts exactly one clean-named pair via `uniqueCName`. Would fail pre-fix (5 pairs with underscore growth). 312/0, featuregen 188/0.

**Scope note:** The pre-existing 18-sections issue (non-synthetic decls like `timespec_new` appearing 18× — one per instantiation section) is confirmed pre-existing and correctly left out of scope here. Should be filed as a follow-up.

PR #61, opened 2026-06-11
krapper_gen: dedup per-instantiation class copies — one wrapper section per class (#60)

### LINKED ISSUE #60: [bug] Per-instantiation section duplication: non-synthetic decls (e.g. timespec_new) emitted 18x in featuregen's generated .h/.cc
Sibling of #57 (found during its fix, confirmed pre-existing and untouched by PR #59): featuregen's generated `featuregen.h`/`featuregen.cc` contain **18 sections per bound class** — the header import plus each of the 17 `requestInstantiation` passes re-emits the full wrapper section for every already-bound class, so non-synthetic decls like `timespec_new` appear 18×. The Kotlin side dedups (writer last-wins), and the C++ tolerates it (identical redefinitions in one TU? or section-guards — characterize first), but the generated C++ is ~18× larger than necessary (#59 shrank the .cc from 277k to 93k lines by fixing the SYNTHETIC half; the non-synthetic 18× remains).

Likely the same `classes + newClasses` append in IndexedServiceImpl + the writer emitting per-pass sections instead of deduping whole classes by type-key before the C++ write. Fix should preserve determinism (sync byte-identical 2×) and featuregen 188/0; krapped/ shrinks intentionally (characterize the diff like PR #59 did: only whole-section collapses, nothing else).

Also relevant to Phase D (#46): the cpp front-end shouldn't have to reproduce this pathology for parity — fixing it first keeps the parity target clean (same rationale as #57).

monkopedia-coder:
Fixes #60

## Confirmed mechanism

Every `requestInstantiation` appends a re-resolved copy of each already-bound class (`classes = classes + newClasses` in `IndexedServiceImpl`) — that append is the carrier of pass-3 recovery. After featuregen's header parse + 17 instantiation requests, the element list held **18 copies** of each main-bound class (78 classes at 18x, plus 2x/6x for containers bound mid-run; 101 distinct keys).

`KotlinWriter` already collapsed the copies (`associateBy { it.type.toString() }`, last-wins), but `HeaderWriter`/`CppWriter` (via `CodeGenerator.generate`) iterated the **raw list**, emitting a full wrapper section per copy:

* **.h** — 18 identical declarations per symbol (`timespec_new` at 18 places). Legal C redeclarations, so it compiled.
* **.cc** — 18 *definitions* per symbol, which only compiled because `Scope.allocateName` `_`-uniquified the collisions into dead name-distinct variants (`timespec_new`, `_timespec_new`, ... 17 underscores). The live symbol (the one the .h declares and Kotlin imports) was the **first** copy's body — i.e. the C++ side was effectively first-wins while the Kotlin side was last-wins.
* **.def** — module config only, no per-class content: unaffected (verified byte-identical).

## Fix

Dedup at the shared input — the `requestInstantiation` append — with the exact `KotlinWriter` semantics: **last copy wins** (cumulative forcing, #40, guarantees the final copy carries every recovered range accessor regardless of request order), kept at the key's **first-occurrence position** (matching `associateBy`'s LinkedHashMap order so Kotlin output is unchanged). One dedup; all three writers now describe the same final class, and the live .cc body is the last copy's — consistent with Kotlin at last.

## Diff characterization (regenerated at parent ff9c140 vs fix)

* `timespec_new`: **18 → 1** in both .h and .cc; every one of the 101 classes now has exactly **1** section (`.h` 46,131 → 3,729 lines; `.cc` 93,483 → 7,323).
* **Last-wins proof (.h):** the surviving section is **byte-identical to the pre-fix LAST section** for 92/101 classes — spot-checked `timespec`, `StringFeature`, `Vec2`, `std::vector<Thing*>`, and `RangeHolder` (the #40 pass-3-recovery class, where last differs from first: only the last copy carries `items()` — first-wins would have lost it). The 9 divergent classes (`Base`, `Shape`, `Animal`, ...) differ ONLY by removing *within-section* duplicated dyncast/upcast helper declarations — the same root cause (18 derived-class copies in `ClassLookup`), verified mechanically: zero added lines, removals all duplicate helper decls.
* **.cc:** surviving section == pre-fix last section exactly (18 classes) or modulo the underscore de-mangling (73; clean names now on the live definitions); the remaining 10 additionally drop the duplicated dyncast helper *definitions* (verified: residual diffs are pure duplicate removals).
* **Everything non-C++ byte-identical:** Kotlin `src/` (119 files), `.def`, forcing headers, `generated.txt`, `requested.txt` — only `.h`/`.cc` changed.
* **Stability:** two post-fix syncs → byte-identical tree; parent-commit regeneration also reproduced the saved baseline exactly.

Known sibling (out of scope, pre-existing): namespace **free functions** (`ResolvedMethod` elements, e.g. `hh_default_handler`) still duplicate 18x in .h/.cc *and* in the Kotlin facades (`_default_handler`...) — KotlinWriter never deduped those either, so fixing them changes the Kotlin baseline; worth its own issue.

## Test

`repeatedForcingEmitsOneSectionPerClass` (ReturnShapeTest, mirroring #59's pattern): two forcing rounds over a bound class whose range accessor only pass-3 recovery restores; asserts one copy per type-key, the recovered (last) copy survives, and `HeaderWriter` emits exactly one section per class.

## Verification

* `:krapper_gen:nativeTest` **313/0** (312 + new test), `:krapper_model:build`, `:featuregen:kplusplusSync`, `:featuregen:nativeTest` **188/0** on the new baseline, `:krapper_gen:ktlintCheck` — all green.
* Gated: `:cppfrontend:goldenCompare :cppfrontend:handoffGenerate -PenableClang` green (exercises the changed C++ writer end-to-end).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -267,9 +267,15 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
             alreadyBoundKeys,
             forcedContainerKeys
         )
-        classes = classes + newClasses.filter {
-            it !is ResolvedMethod || it.forcingIdentity in boundMethodIds
-        }
+        // One entry per class type-key, last copy wins (#60): without this, every
+        // writer received a fresh full copy of each bound class per instantiation
+        // request and the C++ writers emitted a complete wrapper section per copy
+        // (18 sections per class in featuregen). See dedupClassesLastWins.
+        classes = (
+            classes + newClasses.filter {
+                it !is ResolvedMethod || it.forcingIdentity in boundMethodIds
+            }
+            ).dedupClassesLastWins()
     }
 
     // The std headers needed to declare the types named in an instantiation
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/Resolver.kt
@@ -385,6 +385,35 @@ suspend fun List<WrappedElement>.resolveForcing(
     return resolved
 }
 
+/**
+ * Collapse the per-instantiation class copies to ONE entry per class type-key (#60).
+ *
+ * Each `requestInstantiation` appends a re-resolved copy of every already-bound class
+ * (that's how pass-3 recovery upgrades them), so after featuregen's header parse + 17
+ * instantiation requests the element list carries 18 copies of each main-bound class.
+ * KotlinWriter already collapses these via `associateBy { it.type.toString() }` —
+ * last copy wins — but HeaderWriter/CppWriter iterated the raw list, emitting a full
+ * wrapper section per copy: 18 identical declarations per symbol in the .h (legal
+ * redeclarations) and 18 definitions in the .cc that only compiled because the
+ * builder's `Scope.allocateName` `_`-uniquified the colliding names into dead
+ * `_timespec_new`/`__timespec_new`/... variants.
+ *
+ * Mirror the KotlinWriter semantics exactly: LAST copy wins (the most recent pass-3
+ * re-resolution is the most completely resolved one — cumulative forcing guarantees the
+ * final copy carries every recovered range accessor regardless of request order), kept
+ * at the key's FIRST position (matching `associateBy`'s LinkedHashMap order, so the
+ * Kotlin output is unchanged). Non-class elements pass through untouched.
+ */
+fun List<ResolvedElement>.dedupClassesLastWins(): List<ResolvedElement> {
+    val lastByKey = filterIsInstance<ResolvedClass>().associateBy { it.type.toString() }
+    val seen = mutableSetOf<String>()
+    return mapNotNull { element ->
+        if (element !is ResolvedClass) return@mapNotNull element
+        val key = element.type.toString()
+        if (seen.add(key)) lastByKey.getValue(key) else null
+    }
+}
+
 data class ResolveContext(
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. Tier 1 — kplusplus baseline.

**Live-symbol crux: latent bug repaired, not just bloat removal.**

Pre-fix: every `requestInstantiation` call appended a fresh re-resolved copy of each bound class; `KotlinWriter` collapsed them (`associateBy`, last-wins) but `HeaderWriter`/`CppWriter` iterated the raw list — emitting a full section per copy. `Scope.allocateName`'s `_`-uniquification meant only the FIRST definition was the live C++ symbol while Kotlin always described the LAST copy. For #40 recovery classes (e.g. `RangeHolder`/`Holder`) where later copies carry a range accessor (`items()`) that earlier copies lack, Kotlin called a method whose live C++ symbol was a stale first-copy body — a real inconsistency, now fixed. `featuregen`'s 188 green tests exercising through the generated Kotlin surface confirm the last-copy C++ bodies are correct.

**`dedupClassesLastWins` semantics:** verified exact parity with `KotlinWriter`'s `associateBy { it.type.toString() }` — `lastByKey` pre-built (last value, LinkedHashMap insertion order = first-occurrence position), then on iteration: first occurrence of a key substitutes the last copy at that position; duplicates → null. Kotlin output provably byte-identical.

**Non-class elements:** `ResolvedMethod` and others pass through the `!is ResolvedClass` guard untouched — free-function duplication is a pre-existing out-of-scope sibling (noted for its own issue).

**Test `repeatedForcingEmitsOneSectionPerClass`:** three meaningful assertions — (a) one copy per key, (b) last-wins for the recovered `items()` method, (c) `HeaderWriter` section count — using the established `createIndex`/`parseHeader`/`/tmp/` integration pattern.

**Build:** krapper_gen 313/0, featuregen 188/0, `:cppfrontend:goldenCompare` + `:cppfrontend:handoffGenerate -PenableClang` green.

PR #63, opened 2026-06-11
krapper_gen+cppfrontend: instantiation forcing on the cpp path — --frontend=cpp --instantiate end-to-end (#45 brick 3)

monkopedia-coder:
#45 brick 3 (Phase C): instantiation forcing on the cpp path — `--frontend=cpp --instantiate` works end-to-end on an instantiation-bearing fixture, the gate to Phase D.

## Design choice: (a) cppfrontend does the forcing parses — with an oracle-first proof split

Both candidate shapes were judged against the real code:

- **(b) model-side instantiation** was REJECTED on evidence: krapper's "instantiation" is *already* model-side substitution (`WrappedTemplate.typedAs`) — what the forcing re-parse actually provides is the **parsed std template declaration** (featuregen's headers don't include `<vector>`; the template only enters the model through the forcing TU's `#include <vector>`). There is no parse to delete — (b) collapses into (a) minus generality.
- **(a)**: cppfrontend synthesizes the SAME forcing header the libclang path synthesizes (`ForcingHeader`, extracted to `:krapper_model` so one definition serves both front-ends), parses it as a separate args-bearing TU (`kppbridge::buildASTWithArgs` — clang::tooling needs an explicit `-resource-dir` for the builtin headers behind `<vector>`), and emits the tree as ModelIo JSON. `krapper_gen --frontend=cpp --forcingModel '<spec>=<path>'` loads it instead of re-parsing through libclang; **everything downstream — the scoped collection, `resolveForcing`'s 3-pass policy dance, cumulative `forcedContainerKeys`, last-wins dedup — is the same code on both paths** (one seam in `requestInstantiation`).

## Cross-TU identity resolution

`Decl::getID()` is per-ASTContext (NOT stable across `buildASTFromCode` calls), so each payload is self-contained: nothing references identity ACROSS payloads, and krapper_gen merges the trees by **structural identity (type-string keys)** — `alreadyBoundKeys`, the tracker maps, and dedup are all string-keyed, which is exactly what the libclang USR memo's shared instances collapse to. Dependent param refs inside template members now key on the param **NAME** (`template<_Tp>`) — the libclang in-template fallback every member use actually hits, and the spelling baked into uniqueCNames (an identity key would leak nondeterministic `cpp:<id>` into generated symbols); `typedAs` maps name AND usr, so substitution is unaffected.

## Proven end-to-end (`:cppfrontend:handoffInstDiff`, gated `-PenableClang`)

Fixture: `Item` + `Bag::items() -> std::vector<Item*>` (featuregen's RangeHolder shape) + `--instantiate "std::vector<Item*>"`. `items()` drops on the main resolve (container unbound) and is recovered by pass-3 forcing — the full libclang flow, mirrored.

- **Gate 1 — THE FLOW PROOF (byte)**: the libclang baseline also dumps its parse trees (`--dumpModels`, main + forcing); fed back through `--frontend=cpp --forcingModel`, output is **byte-identical across all 9 files** (path-modulo .def). This isolates the file-handoff flow — fresh instances, no cursor→element memo — from front-end construction fidelity.
- **Gate 2 — the cppfrontend-model run (functional)**: the C++-AST front-end's own parses through the same flow materialize the specialization, recover `items()`, produce the SAME file set, core container surface present, wrapper compiled. `root_Bag.kt` / `root_Item.kt` / `CppBinding.kt` are **byte-identical**; `std_Vector__Item_P.kt` differs in **8 lines, all one cause**: two ctor uniqueCNames spell `const_allocator_type_and` (libclang keeps the alias-name leaf its order-dependent `visit()` collapse couldn't resolve) vs `const_template__Alloc_and` (this front-end's deterministic collapse). Public Kotlin API is identical.

**Why gate 2 is functional, not byte**: member uniqueCNames bake libclang's *lossy pre-substitution spellings* (`template<_Pointer>`, `_E*<...>`, Unexposed-fallback artifacts, order-dependent alias collapse — evidence table in the investigation). Byte-parity there means emulating that lossiness bug-for-bug, which this rewrite exists to delete. The flow itself is held to byte-parity by gate 1.

## What it took (cppfrontend fidelity work, each mirroring a cited libclang rule)
- CIndex `GetTemplateArguments` mirror (`kppbridge`): WRITTEN args preferred (`std::vector<Item*>` is ONE arg — the forcing key depends on it); dependent specializations (`vector<_Tp,_Alloc>` in member signatures) decode structurally.
- Redeclaration collapse for classes/templates (the USR-memo equivalent); params re-merge per redecl (`otherParams`), members attach from the defining redecl only.
- Protected/private BASE specifiers filtered (ModelFactories' access filter) — vector's `protected _Vector_base` would hard-fail INCLUDE reference-expansion.
- `referenceTypedefElement` mirrored (`reference`/`const_reference` -> `value_type&`) — the at/operator[]/front/back survival rule.
- Anonymous-record typedef-name rule (`max_align_t`).
- `std::initializer_list` UNRESOLVABLE by construction (the boundary can never marshal one; libclang drops the same members by spelling accident).

## Verify
- FULL default gate: krapper_gen **313/0**, featuregen **188/0** (+ kplusplusSync, ZERO krapped/ drift — the libclang path with the `ForcingHeader` refactor is byte-stable), krapper_model build, ktlintCheck — all green.
- Gated: cppfrontend self-checks ALL PASS, goldenCompare PASS (normalizer unchanged — it already rewrote both ref-key forms), handoffGenerate (base) green, handoffInstDiff both gates green.

## Remainder (filed on #45)
- Gate-2 byte-parity = the alias-collapse divergence (N5): durable convergence needs either deterministic collapse on the libclang path or alias preservation here — Phase D tail.
- `pointerTypedefElement`/`assocTypedefElement` mirrors (unique_ptr/map/set instantiations) + `WrappedTemplateParam` defaultType capture: needed for featuregen's full 17-instantiation set, not for vector.
- Multi-instantiation cumulative recovery on the cpp path is wired (`forcedContainerKeys` is front-end-agnostic) but only single-instantiation is fixture-proven.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ gradle-app.setting
 local.properties
 .claude/worktrees/
 .kotlin/
+*.hprof
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -180,7 +180,12 @@ kplusplus {
         // default's source text — an inline helper in the slice header wrapping
         // Lexer::getSourceText over ParmVarDecl::getDefaultArgRange() (see clang_slice.h
         // for why Lexer/SourceManager/LangOptions aren't bound wholesale yet).
-        "kppbridge::defaultArgText"
+        "kppbridge::defaultArgText",
+        // NEW for #45 brick 3 (instantiation forcing): the forcing-parse entry point —
+        // buildASTFromCode with REAL driver args ('\n'-joined; -resource-dir + -std), the
+        // smallest bridge until std::vector<std::string> params are bindable (see
+        // clang_slice.h).
+        "kppbridge::buildASTWithArgs"
     )
     // FIXUPS (documented generator gaps, #44 brick 5): krapper's operator generation
     // emits invalid Kotlin for four of llvm::APSInt's C++ operators —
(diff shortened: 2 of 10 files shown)


monkopedia-reviewer (approved):
## Review: krapper_gen+cppfrontend brick 3 — APPROVE

**Tier 1 (kplusplus override: all PRs tier 1)**
**Category: feature_internal** — new capability behind `--frontend=cpp --instantiate`; no public-API change

### Default-path purity (production risk #1) — CLEAN

The `requestInstantiation` seam is correct. The new branch is:
```kotlin
val resolver = cppForcingModelPaths[target]?.let { path -> loadParsedModel(path) }
    ?: run { /* original libclang parse, unchanged */ }
```
`cppForcingModelPaths` defaults to `emptyMap()` in Parsing.kt, so the `?.let` short-circuits to `null` on the libclang path, and the original `scope.parseHeader(…)` runs exactly as before. The fidelity mirrors (redecl collapse, protected-base filter, referenceTypedef, anonymous-record typedef-name, initializer_list UNRESOLVABLE) are ALL confined to cppfrontend's `ModelBuilder.kt` / `TypeBuilder.kt` — none touch shared resolver code. Zero-drift confirmed (featuregen 188/0 with full sync).

### ForcingHeader extraction — move-not-modify ✓

`ForcingHeader.forceName()`, `contentFor()`, and `stdHeadersFor()` are byte-identical lifts from the old `IndexedServiceImpl` inlines. The `STD_TYPE_HEADERS` list, filter logic, and buildString content are unchanged. Proved by zero featuregen sync drift.

### Gate 1 mechanics (`--dumpModels`) — clean ✓

Default-off global (`dumpModelsDir = null`). Set before the frontend check in App.kt. Dump is placed at the same post-reduce, pre-rewrite point as the other oracle hooks; `loadParsedModel` re-applies the two rewrites on load. The cpp path early-returns at `cppParsedModelPath?.let { return … }` before reaching the dump code — so `--dumpModels` is physically inert on the cpp path even if both flags are passed.

### Param-NAME keying — safe ✓

`TypeBuilder` now emits `WrappedTemplateRef(name)` (e.g. `"T"`) instead of `WrappedTemplateRef(cppUsr)`. Verified in `Parsing.typedAs` (lines 876-880): the substitution map is built as `mapping[templates[i].name] = mappedType` AND `mapping[templates[i].usr] = mappedType` (plus `otherParams`). So name-keyed refs hit the mapping exactly as usr-keyed refs do. Cross-template collision is impossible: each `typedAs` call builds its own local mapping scoped to that template's params. The `uniqueCNames` parity requirement (leaking `cpp:<id>` into generated symbols would be a real defect) is the load-bearing argument, and it holds: the name baked into the pre-substitution spelling IS the key.

### Gate 2 8-line divergence — documented and parked ✓

Gate 2 passed as FUNCTIONAL; gate 1 (byte) passed for the file-handoff flow. The alias-spelling difference (`const_allocator_type_and` vs `const_template__Alloc_and`) is an N5-class consequence of libclang's order-dependent alias-collapse vs this front-end's deterministic decode — correctly NOT normalized away in the builder (emulating lossiness bug-for-bug would defeat the rewrite). Parked on #45.

### hprof incident — history clean ✓

Verified: no blob in the branch's object graph exceeds 100KB (max commit object size ~1.5KB). `*.hprof` added to `.gitignore` in the final commit. Strip was effective before push.

### Build verification — all green

- FULL gate: `krapper_gen nativeTest` **313/0** ✓, `featuregen nativeTest` **188/0** ✓, `featuregen:kplusplusSync` ZERO drift ✓, `krapper_model:build` ✓, `ktlintCheck` ✓
- Gated (`-PenableClang`): `cppfrontend:goldenCompare` PASS ✓, `cppfrontend:handoffGenerate` green ✓, `cppfrontend:handoffInstDiff`:
  - Gate 1 (byte): *"byte-identical (path-modulo .def) across 9 files"* ✓
  - Gate 2 (functional): *"cppfrontend-model run materialized the specialization + recovered items() (same file set, core surface present)"* ✓

Quality is consistent with surrounding code; comments are dense but necessary given the complexity of the fidelity mirroring. No idiom, naming, or duplication issues.

PR #64, opened 2026-06-11
cppfrontend: Phase D entry — remaining fidelity mirrors + the first full featuregen parity attempt (#46)

monkopedia-coder:
Phase D entry brick (#46): the remaining cppfrontend fidelity mirrors (#45 tail) + the FIRST full featuregen parity attempt under `--frontend=cpp`.

## Part 1 — the remaining mirrors (each citing its TypeFactories/ModelFactories source)
- **pointerTypedefElement** → `TypeBuilder.pointerTypedef`: a DEPENDENT `pointer`/`const_pointer` member alias reduces to `element_type*` (unique_ptr's `get()`/`release()`/`operator->` survival); dual-kind `element_type` sibling lookup (`using` included), concrete pointer typedefs fall through untouched.
- **assocTypedefElement** → `TypeBuilder.assocTypedef`: dependent `key_type`/`mapped_type`/`value_type` reduce to the enclosing template's params (map shape: 0/1; key-only set shape: both → 0), params reached through `getDescribedClassTemplate()`; name-keyed per this front-end's in-template convention (typedAs accepts name or USR).
- **WrappedTemplateParam.defaultType** → `kppbridge::defaultArgType` (the TemplateArgumentLoc unwrap) + `TypeBuilder.defaultTypeResidue`: reproduces the libclang CURSOR-WALK RESIDUE shapes the model's `defaultType` getter reads, pinned against the live oracle (`U = Alloc<T>` → `[UNRESOLVABLE, tref(T)]`; `U = T`/`U = T*` → `[tref(T)]`; builtin → none); attached to the canonical param instance so a default first declared on a later redeclaration (basic_string after `<iosfwd>`) still lands.
- Fixture + self-checks grown 91 → **100** (SmartPtr/RawHolder/MiniMap/MiniSet/DefBox/DefPair, all pre-validated against the libclang parse of the same fixture).

## Part 1.5 — fixes that fell out of the parity run (cheap single-cause items, fixed inline)
- **Non-type template args decode to UNRESOLVABLE** (probed: libclang gives `Mask<4>` → `Mask<unresolveable>`, member drops); dropping the arg materialized a broken zero-arg `Flags<>` binding that failed the wrapper compile.
- **T1.3 range rewrite mirrored** (`rangeVectorReturn`/`elementOfIterator`/`stripDecoration`): `iterator_range<It>` returns materialize into `std::vector<Elem*>` at model construction — RangeHolder::items() now byte-matches.
- **Nested classes** (`Outer::Inner` binds; `Wrapper<T>::Inner` correctly does NOT — the nested-in-class-template skip).
- **Inline-namespace-preserving qualified names** (`kppbridge::qualifiedName`): `std::__cxx11::basic_string` now spells with the inline namespace, so std::string members resolve and Basic_string__Char/Wchar_t materialize on the cpp path.
- **`type-parameter-` leaf guard**: a canonical spelling still carrying a dependent placeholder is UNRESOLVABLE (was baking `type-parameter-0-0` into uniqueCNames → unordered_map wrapper-compile crash).

## Part 2 — the parity worklist generator
- `cppfrontend --parity-emit`: arbitrary header + per-spec ForcingHeader-shared forcing parses, ONE process per payload (crash isolation).
- `:cppfrontend:featuregenParity`: for each of featuregen's REAL 17 instantiation requests (requested.txt + the 6 seeded `instantiate()` calls) PLUS the ALL unit (the real sync's shape), generate through BOTH front-ends with an identical CLI tail and diff (byte-compare, `.def` path-modulo, unified diffs persisted per unit). Exits nonzero only on REGRESSION vs the committed `parity-expectations.txt` ledger — CI-able while Phase D converges; improvements print "ratchet the ledger".

## THE SCOREBOARD (first honest run)
**0 byte-identical / 18 diff / 0 generation-failed** — every unit (including ALL with all 17 instantiations) generates AND compiles end-to-end on `--frontend=cpp`. Per-spec: all 17 specs land in the small-categorized-diff class (~930–1100 diff lines each, dominated by one shared header-wide baseline; ALL = 3037 lines across 16 files).

### Categorized divergence families (the Phase D worklist, also filed on #46)
- **D1 initializer_list materialization** (~50% of diff lines): libclang's order-dependent `visit()` collapse RESOLVES `initializer_list` at featuregen scale (materializing `std_Initializer_list__*` + the ilist ctor/assign/insert overloads); the cpp front-end's documented deterministic UNRESOLVABLE rule drops them (the same rule that keeps the bag-fixture gate byte-identical, where libclang DROPS them — the order-dependence is the wall; convergence = deterministic collapse on the libclang side or scale-aware handling, tracked with N5).
- **D2 incidental INCLUDE_MISSING std surface** (~40%): `std::abs`×5/`wcschr`/`wcspbrk`/`wcsrchr`/`wcsstr`/`wmemchr`/`get_new_handler`/`set_new_handler` + `_IO_FILE` bind on libclang only; `__convert_from_v` + `__locale_struct` on cpp only — suspected LinkageSpecDecl/using-decl walk differences in ModelBuilder.addContextDecls + reference-graph deltas.
- **D3 dependent member-alias spellings on surviving members** (small): both sides keep a member but bake different pre-substitution arg spellings into its uniqueCName (set::insert's USR-keyed `template<c:...stl_set.h@3743>` vs the cpp overload set; unordered_map's `typename _Hashtable::hasher` vs `unresolveable`).
- **D4 out-of-line virtual destructor**: the libclang cursor walk REPARENTS `Drawable`'s out-of-line-defined dtor to the TU and loses it ("Can't find parent type" → no dispose()); the cpp path keeps it and emits dispose()/owned() — correct-by-construction; decide normalizer vs libclang-side fix.

## Verify
- FULL gate: krapper_gen nativeTest **313/0**, featuregen kplusplusSync + nativeTest **188/0**, krapper_model build, ktlintCheck — green, ZERO krapped/ drift.
- Gated: **100/100** self-checks, goldenCompare PASS (N1–N5), handoffGenerate, handoffInstDiff (oracle byte-identical + functional), featuregenParity green vs ledger.

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -181,6 +181,10 @@ kplusplus {
         // Lexer::getSourceText over ParmVarDecl::getDefaultArgRange() (see clang_slice.h
         // for why Lexer/SourceManager/LangOptions aren't bound wholesale yet).
         "kppbridge::defaultArgText",
+        // NEW for Phase D (#46): WrappedTemplateParam.defaultType capture — the
+        // TemplateArgumentLoc unwrap for TemplateTypeParmDecl::getDefaultArgument()
+        // (see clang_slice.h).
+        "kppbridge::defaultArgType",
         // NEW for #45 brick 3 (instantiation forcing): the forcing-parse entry point —
         // buildASTFromCode with REAL driver args ('\n'-joined; -resource-dir + -std), the
         // smallest bridge until std::vector<std::string> params are bindable (see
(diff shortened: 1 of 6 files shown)


monkopedia-reviewer (approved):
Phase D entry: reviewed to merge standard.

**Purity** — all 6 changed files are cppfrontend-scoped; zero krapper_gen/krapper_model/featuregen drift (kplusplusSync regenerated cleanly, git status clean).

**Ratchet ledger** — `parity-expectations.txt` semantics verified: ranks identical<diff<fail, regression = worse-than-expected exits nonzero, improvement prints ratchet message and exits 0. Committed state matches the observed run (0 identical / 18 diff / 0 fail, all [as expected]).

**Five inline fixes** — each spot-checked against its libclang source:
- Non-type template args → UNRESOLVABLE: `map { UNRESOLVABLE }` vs old `mapNotNull { null }`. Correct: libclang yields CXType_Invalid for these (Mask<4> → Mask<unresolveable>); dropping the arg materialized a broken zero-arg Flags<> that failed wrapper compile.
- T1.3 iterator_range mirror: `rangeVectorReturn`/`elementOfIterator`/`stripDecoration` structurally match ModelFactories.kt (pointer-peel, value_type alias, first-template-arg). RangeHolder::items() byte-matches.
- Nested classes: Outer::Inner binds via the new WrappedClass branch; Wrapper<T>::Inner correctly skipped (parent is WrappedTemplate, not WrappedClass).
- Inline-namespace qualified names: `qualifiedName()` bridge sets PrintingPolicy::SuppressInlineNamespace=None, exposing std::__cxx11::basic_string so std::string members resolve.
- type-parameter- leaf guard: `if (spelling.contains("type-parameter-")) return UNRESOLVABLE` before leaf construction. Correct: DependentNameType members are CXType_Unexposed on the libclang path (dropped); the cpp path was baking `type-parameter-0-0` into uniqueCNames.

**D4 dtor spot-check** — strings_feature.h:427 declares `virtual ~Drawable();` in-class; :431 defines `inline Drawable::~Drawable() {}` out-of-line. ModelResolution.kt:331 shows libclang loses it via "Can't find parent type" (createThisArg → parentClass null). Cpp path keeps the dtor under Drawable's CXXRecordDecl — correct-by-construction. D4 claim is honest.

**D1-D4 categorization** — parity output confirmed: only-libclang files are _IO_FILE.kt + std_Initializer_list__*.kt (D1+D2), only-cpp = __locale_struct.kt (D2). No uncategorized residue visible.

**defaultTypeResidue confinement** — only adds children to WrappedTemplateParam; consumed only by WrappedTemplateParam.defaultType → typedAs. Does not leak into generated output decisions beyond what parity requires. Kind.TemplateTypeParm guard (line 232) is in place before the TemplateTypeParmDecl cast (line 252) — type-safe.

**Build** — 100/100 self-checks (exit 0), goldenCompare PASS, handoffInstDiff byte-identical + functional, featuregenParity 0 fail / 18 diff [all as expected], krapper_gen 313/0, featuregen 188/0 (kplusplusSync regenerated 17 specs cleanly), ktlintCheck clean. No CI checks configured for this gated module — local gates are the full verification surface.

PR #65, opened 2026-06-11
cppfrontend: D2 convergence — LinkageSpecDecl traversal + diff-line-ratcheted parity ledger (#46)

monkopedia-coder:
Phase D convergence brick for #46: family **D2 — the incidental INCLUDE_MISSING std surface** (~40% of the parity diff lines at entry).

## Root causes pinned (per mechanism)

**D2a — the cpp walk skipped `LinkageSpecDecl` (FIXED).** ModelBuilder.addContextDecls never descended into `extern "C"`/`extern "C++"` blocks, so the cpp model was missing every std function libstdc++ declares inside one — `<cwchar>`'s `std::wcschr/wcspbrk/wcsrchr/wcsstr/wmemchr`, `<bits/std_abs.h>`'s `std::abs` x7, `<new>`'s `get/set_new_handler`, `<type_traits>`' `__aligned_storage_default_alignment` (all wrapped in `extern "C++" { namespace std { ... } }`), plus `std::div`/`__gnu_cxx::div`/`__gnu_cxx::__is_null_pointer` — AND glibc's `struct _IO_FILE`, whose *definition* sits inside `extern "C"`: the cpp model carried only the TU-scope forward decl, so the directly-bound class was EMPTY (32 members on the libclang side) and dropped as "Empty class", which is exactly the `root__IO_FILE.kt` file-set divergence.

The fix mirrors libclang's semantics precisely: mapAll's `map(parentCursor) ?: return` has no LinkageSpec branch, so a linkage spec's DIRECT children never attach — but the walk descends, and anything one level deeper attaches to its own mapped parent. `addContextDecls(attach=false)` therefore creates namespaces/classes/templates DETACHED (memo only, members still walked) and skips typedefs/free functions; `attachIfNew` reproduces the lazy `parent.addChild` at the first sighting outside a linkage spec. Post-fix parse-model diff: free functions/templates/typedefs are at **zero set-difference** between the two front-ends on the featuregen TU (was 12 missing free-function entries), and `_IO_FILE` carries 32 children on both sides.

**D2b — libclang reference-spelling artifacts break its OWN INCLUDE_MISSING lookups (LEDGER-ACCEPTED, cpp side is more correct).** Two instances, both cited in the ledger:
* `std::__convert_from_v`'s `__cloc` arg leaf is spelled `struct __locale_struct` by libclang (the `__c_locale` typedef collapse keeps the elaborated keyword; USR `c:c++locale.h@2318`), which misses the class-map key `__locale_struct` — KRAPPER_DEBUG_RESOLVE shows `Method failed from argument __cloc: const struct __locale_struct*&: Missing type`, so libclang silently DROPS the function. The cpp leaf is `__locale_struct`, resolves, and emits the function + `root___locale_struct.kt` (every unit).
* `std::__exception_ptr::exception_ptr` is misnamed **`exception_ptr::exception_ptr`** in the libclang model (out-of-line-member spelling artifact in `bits/exception_ptr.h`), so libclang's lookup for `std::current_exception`/`rethrow_exception`'s return type fails and both functions + the class are dropped; the cpp path binds them (+ `std::type_info` via `__cxa_exception_type`) — `std::unique_ptr<int>` and ALL units only (`<memory>` pulls `<exception>`).

Mirroring either spelling would be bug-for-bug emulation of the lossiness the bootstrap exists to delete (same reasoning as D4 and handoffInstDiff gate 2), so these are accepted as diffs with citations, not converged.

## The ratchet (diff lines, before -> after)

| unit | before | after |
|---|---|---|
| Box&lt;int&gt; / Box&lt;Point&gt; / Box&lt;Box&lt;int&gt;&gt; / Pair2&lt;int, double&gt; / std::unique_ptr&lt;int&gt; | 955 | 336 / 336 / 336 / 336 / 533* |
| std::pair&lt;int, int&gt; | 933 | 336 |
| std::map&lt;int, double&gt; / std::map&lt;int, int&gt; | 955 | 368 |
| std::set&lt;int&gt; | 1001 | 414 |
| std::unordered_map&lt;int, int&gt; | 963 | 376 |
| std::unordered_set&lt;int&gt; | 1006 | 419 |
| std::vector&lt;Point&gt; / &lt;double&gt; | 1069 | 482 |
| std::vector&lt;Thing*&gt; | 992 | 405 |
| std::vector&lt;int&gt; | 1032 | 445 |
| std::vector&lt;string&gt; | 1069 | 450 |
| std::vector&lt;vector&lt;int&gt;&gt; | 1109 | 522 |
| **ALL** | **3037** | **1383** |

\* unique_ptr carries the exception_ptr/type_info D2b residue (the two extra cpp-side files).

Total across the 17 per-spec units: 16,936 -> 6,648 diff lines (ALL: 3,037 -> 1,383, **-54%**). `root__IO_FILE.kt` is gone from every unit's file-set diff. Per-family decomposition of a representative unit (Box&lt;int&gt; featuregen.cc): D1 initializer_list 132, D2b 37, D4 5, other 0 — D2a residue is **zero**.

**Ledger hardening:** `parity-expectations.txt` now carries a per-unit `diff:<N>` bound and featuregenParity enforces it — a unit that stays in the diff class but diverges by MORE lines than recorded is a regression; fewer lines prompts the ratchet. Intra-diff improvements were previously invisible to the gate.

## Entanglements noted
The remaining diff lines decompose into D1 (initializer_list — dominant share), D3 (alias spellings in basic_string uniqueCNames), D4 (ledgered) and the two D2b accepts above. No D2 work was deferred into D1/D3.

## Verify
* FULL gate: krapper_gen nativeTest **313/0**, featuregen nativeTest **188/0**, kplusplusSync + krapper_model build green, ktlintCheck green, **zero krapped/ drift** (no shared krapper_gen/krapper_model code touched).
* Gated: self-checks **100 PASS / 0 FAIL**, goldenCompare green, handoffGenerate green, handoffInstDiff oracle **byte-identical across 9 files** + functional gate green, featuregenParity **0 fail / 18 diff / 0 identical, 0 regressions** against the ratcheted ledger.

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -524,8 +524,11 @@ tasks.register("featuregenParity") {
         // Byte-compare the two outputs: same file set, every file identical (the .def
         // path-modulo — it bakes the output dir's absolute path; .a excluded — derived
         // from the compared .cc). On divergence, a unified diff per file lands under
-        // diffs/<unit>/ for the worklist analysis.
-        fun diffUnit(unit: String, libclang: File, cpp: File): Pair<String, String> {
+        // diffs/<unit>/ for the worklist analysis. Third element: the diff-line count,
+        // ratcheted per-unit against the ledger's `diff:<N>` bound (D2 brick) so an
+        // intra-diff regression — more divergence while staying in the diff class —
+        // fails the task just like a class regression.
+        fun diffUnit(unit: String, libclang: File, cpp: File): Triple<String, String, Int> {
             fun files(dir: File) = dir.walkTopDown()
                 .filter { it.isFile && it.extension != "a" }
                 .map { it.relativeTo(dir).path }.toSortedSet()
@@ -560,15 +563,17 @@ tasks.register("featuregenParity") {
                 }
             }
             return if (notes.isEmpty() && diffFiles == 0) {
-                "identical" to "byte-identical (path-modulo .def) across ${a.size} files"
+                Triple("identical", "byte-identical (path-modulo .def) across ${a.size} files", 0)
             } else {
-                "diff" to (
-                    notes + "$diffLines diff line(s) across $diffFiles file(s)"
-                    ).joinToString("; ")
+                Triple(
+                    "diff",
+                    (notes + "$diffLines diff line(s) across $diffFiles file(s)").joinToString("; "),
+                    diffLines
+                )
             }
         }
 
-        data class Verdict(val unit: String, val cls: String, val detail: String)
+        data class Verdict(val unit: String, val cls: String, val detail: String, val lines: Int)
         val verdicts = mutableListOf<Verdict>()
         val units = specs.map { it to listOf(it) } + ("ALL" to specs)
         for ((unit, instSpecs) in units) {
@@ -580,13 +585,14 @@ tasks.register("featuregenParity") {
             if (libExit != 0) {
                 verdicts += Verdict(
                     unit, "fail",
-                    "LIBCLANG BASELINE failed (exit $libExit — infra, see $libLog)"
+                    "LIBCLANG BASELINE failed (exit $libExit — infra, see $libLog)",
+                    0
                 )
                 continue
             }
             val missing = instSpecs.filter { (forcingPaths[it] ?: "EMIT_FAILED") == "EMIT_FAILED" }
             if (missing.isNotEmpty()) {
-                verdicts += Verdict(unit, "fail", "cppfrontend model emit failed for $missing")
+                verdicts += Verdict(unit, "fail", "cppfrontend model emit failed for $missing", 0)
                 continue
             }
             val cppLog = File(logsDir, "${paritySlug(unit)}_cpp.log")
@@ -599,11 +605,11 @@ tasks.register("featuregenParity") {
             )
             if (cppExit != 0) {
                 val tail = cppLog.readLines().lastOrNull { it.isNotBlank() }.orEmpty()
-                verdicts += Verdict(unit, "fail", "cpp generation failed (exit $cppExit): $tail")
+                verdicts += Verdict(unit, "fail", "cpp generation failed (exit $cppExit): $tail", 0)
                 continue
             }
-            val (cls, detail) = diffUnit(unit, outLib, outCpp)
-            verdicts += Verdict(unit, cls, detail)
+            val (cls, detail, lines) = diffUnit(unit, outLib, outCpp)
+            verdicts += Verdict(unit, cls, detail, lines)
         }
 
         // Scoreboard + the regression gate against the committed expected-state ledger.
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Tier 1 — auto-approve. Full independent build verified.

**Attach semantics (D2a):** The `attachIfNew` / `attach: Boolean` approach correctly mirrors libclang's behavior. libclang's `map(linkage_spec_cursor)` returns null → `addToParent` skips (direct children of a linkage spec never attach as model children), but the walk still descends and one-level-deeper elements attach to their own mapped parent. The cpp implementation matches exactly: `attach=false` for immediate linkage-spec children, defaulting to `attach=true` for the namespace/class contents. Double-attach is impossible by construction — `attachIfNew` guards with `element.parent == null`, and `addChild` sets `parent`; once attached the guard never fires again.

**D2b spot-check (confirmed live):** featuregenParity output shows `only-cpp=[src/root___locale_struct.kt]` in every unit — direct empirical confirmation that libclang drops `std::__convert_from_v` due to the `struct __locale_struct` elaborated-keyword artifact while the cpp path resolves it correctly. The `std_Type_info.kt` / `std___exception_ptr_Exception_ptr.kt` pair appears only in `std::unique_ptr<int>` and ALL, consistent with the `exception_ptr` misnaming claim. Both D2b ledger-accepts are proven libclang defects, not unexplained diffs.

**Ledger hardening:** The `diff:<N>` ratchet semantics are correct. Parser handles both `diff` (no bound) and `diff:N` (with bound) correctly. `lines > expLines` → REGRESSION (with counter increment); `lines < expLines` → IMPROVED (ratchet message); equal → as expected. No case is skipped or double-counted.

**Build gate results:**
- `:krapper_gen:nativeTest` UP-TO-DATE (313/0)
- `:featuregen:nativeTest` UP-TO-DATE (188/0)
- `:featuregen:kplusplusSync` ran (17 instantiations, clean)
- `:krapper_gen:ktlintCheck` UP-TO-DATE (green)
- `:krapper_model:build` UP-TO-DATE (green)
- `:cppfrontend:goldenCompare` green
- `:cppfrontend:handoffInstDiff` oracle: byte-identical across 9 files; functional: passed
- `:cppfrontend:featuregenParity` 0 fail / 18 diff / 0 identical, 0 regressions — all 18 units exactly at their ratcheted bounds

**Purity:** zero changes outside `cppfrontend/`. No krapped/ drift.

**Quality:** clean — dense but appropriate comments (explaining libclang mirror mechanics), `attachIfNew` properly factored, `attach` parameter threading consistent, no new ktlint violations in the changed code. Cppfrontend's pre-existing ktlint issues are not introduced by this PR.

monkopedia-reviewer (approved):
Tier 1 auto-approve. Full build verified: krapper_gen nativeTest 313/0, featuregen nativeTest 188/0, kplusplusSync clean, ktlintCheck green, goldenCompare green, handoffInstDiff byte-identical 9 files, featuregenParity 0 fail/18 diff/0 regressions. D2a attach semantics correct (attachIfNew + parent==null guard prevents double-attach); D2b spot-check confirmed live (root___locale_struct.kt in only-cpp every unit proves the libclang elaborated-keyword drop). Ledger regression semantics correct. Zero scope creep, zero new ktlint violations.

PR #67, opened 2026-06-11
cppfrontend: D1a — concrete initializer_list materialization converges (+ the dependent-pointer typedef lossy-collapse mirror) (#46)

monkopedia-coder:
D1a brick of the Phase D parity convergence (#46): characterize the initializer_list divergence family (full decomposition posted on #46) and land its cheapest sub-cause — the CONCRETE-element materialization.

Two surgical TypeBuilder rules:

1. **Concrete initializer_list decodes faithfully.** `DefaultArgs::sumIlist(std::initializer_list<int>)` is an ordinary specialization on the libclang path — the member survives and INCLUDE_MISSING materializes `std_Initializer_list__Int` — so a NON-DEPENDENT element now passes through the normal template decode. Dependent elements (`initializer_list<_CharT>` / `<value_type>` inside template member signatures) keep the documented deterministic UNRESOLVABLE drop (the D1b/D1c residue).

2. **Lossy-collapse mirror for dependent pointer/reference typedefs.** The materialized class's `begin()/end()` return `const_iterator` (`typedef const _E*`), which the libclang path collapses to a bare-NAME spelling leaf — never a substitutable tref — so they FAIL resolution and drop (live oracle log: `fun begin(): const _E* — Couldn't resolve return`). The cpp structural decode produced `pointerTo(const(tref(_E)))`, which would substitute and surface methods libclang never binds; the fall-through now mirrors visit()+createForType's spelling-leaf collapse for typedefs over dependent pointer/reference shapes (concrete underlyings keep the structural path; chained aliases keep the recursive follow).

Ratchet (diff-line bounds, parity-expectations.txt):

| unit | before | after |
|---|---|---|
| Box/Pair2/std::pair units | 336 | 287 |
| std::map units | 368 | 319 |
| std::set\<int\> | 414 | 365 |
| std::unique_ptr\<int\> | 533 | 484 |
| std::unordered_map / _set | 376 / 419 | 327 / 370 |
| std::vector units | 405-522 | 356-473 |
| ALL | 1375 | 1326 |

Every unit -49 diff lines (zero regressions); `std_Initializer_list__Int.kt` converged byte-identically out of every only-libclang file-set note.

Verified: full gate (krapper_gen nativeTest, krapper_model build, featuregen kplusplusSync zero-drift + nativeTest, ktlintCheck) + gated clang suite (cppfrontend self-checks, goldenCompare, handoffGenerate, handoffInstDiff, featuregenParity ratcheted, zero regressions).

### DIFF
--- a/cppfrontend/parity-expectations.txt
+++ b/cppfrontend/parity-expectations.txt
@@ -5,13 +5,30 @@
 # than N diff lines; improvements are reported so this ledger ratchets toward
 # all-identical as Phase D converges.
 #
-# State after the D3 convergence brick. EVERY unit still generates + compiles end-to-end
+# State after the D1a convergence brick. EVERY unit still generates + compiles end-to-end
 # on --frontend=cpp (0 fail). Remaining divergence families (full breakdown on #46):
-#   D1 initializer_list materialization (libclang's order-dependent visit() collapse
-#      RESOLVES initializer_list at featuregen scale; the cpp front-end's documented
-#      deterministic UNRESOLVABLE rule drops the ilist ctors/assign/insert overloads
-#      and the std_Initializer_list__* bindings) — now the dominant share of the
-#      remaining diff lines (132 of Box<int>'s 174 featuregen.cc +/- lines; D2b 37, D4 5).
+#   D1 initializer_list materialization — decomposed + partially converged (D1a brick):
+#      D1a CONVERGED: a CONCRETE element (DefaultArgs::sumIlist's initializer_list<int>)
+#          decodes faithfully, the member survives, and INCLUDE_MISSING materializes
+#          std_Initializer_list__Int byte-identically (-49 diff lines in EVERY unit).
+#          Required the LOSSY-COLLAPSE mirror: a typedef over a DEPENDENT pointer/
+#          reference shape (`typedef const _E* const_iterator`) collapses to libclang's
+#          bare-NAME spelling leaf (never a substitutable tref), so the materialized
+#          class's begin()/end() drop exactly like the live oracle.
+#      D1b remaining: dependent-element member overloads whose libclang spelling is a
+#          NAME tref (basic_string's `std_initializer_list_template__CharT` ctor/op=/
+#          +=/append/assign, vector's `template__Tp` ctor/op=/assign, + the per-element
+#          std_Initializer_list__* materializations they imply) — the same name-tref
+#          convention this front-end uses, so mirrorable deterministically; now the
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
**Tier 1 — APPROVED** (auto-merge follows)

**Category:** `feature_internal` — two surgical TypeBuilder rules for the D1a parity convergence sub-brick; cppfrontend-only.

**Correctness review (axis A):**

*Rule 1 (concrete initializer_list):* The `isDependentType() != false` guard correctly gates materialization to concrete-element only (`initializer_list<int>` materializes; `initializer_list<_CharT>`/`<value_type>` inside template members stay UNRESOLVABLE). Null-safe pattern is idiomatic Kotlin.

*Rule 2 (LOSSY-COLLAPSE mirror):* Narrowly scoped and correctly ordered — `referenceTypedef`/`pointerTypedef`/`assocTypedef` have first opportunity for the special-reducer cases (unexposed dependent shapes), so LOSSY-COLLAPSE only fires for remaining dependent pointer/ref typedefs that libclang also collapses to spelling leaves. The `asTypedefTypeOrNull() == null` guard correctly passes chained aliases to the recursive path. The spelling-leaf result is a `WrappedTypeReference` (bare name, not a `WrappedTemplateRef`), which cannot be substituted — matching libclang's behavior exactly. Oracle-pinned: the featuregenParity report confirms `std_Initializer_list__Int.kt` is now in both file sets (not only-libclang) while `begin()`/`end()` drop on both sides, per the live run at Jun 11 10:52 today.

**Quality review (axis B):** TypeBuilder.kt is ktlint-clean (confirmed by ktlintFormat — file unmodified). Comments cite libclang equivalents at each branch, consistent with surrounding code density. No dead code, no inconsistencies.

**Ratchet:** ALL 1375→1326 (−49 every unit), zero regressions, confirmed by live featuregenParity report. Purity confirmed: only two files changed from main (`TypeBuilder.kt` + `parity-expectations.txt`), zero drift in featuregen/krapped after sync.

**Build results:**
- Full gate (krapper_gen:nativeTest, krapper_model:build, featuregen:kplusplusSync zero-drift, featuregen:nativeTest, krapper_gen:ktlintCheck): **GREEN** ✓
- TypeBuilder.kt ktlint: **CLEAN** ✓ (pre-existing violations in build.gradle.kts and CppFrontend.kt/GoldenCompare.kt/ModelBuilder.kt are unchanged from main — separate cleanup needed)
- featuregenParity ALL=1326 all-as-expected, goldenCompare artifacts Jun 11 10:50 on this commit ✓

No findings warranting request_changes.

Issue #47, opened 2026-06-10
[clang-campaign] Phase E: flip the reducer to the cpp front-end + crossover (stage-0 → mavenLocal) + delete libclang-C


monkopedia-coder:
The endgame, after Phase D (#46) parity:

1. **Crossover mechanics** (decided 2026-06-09, see docs/clang-bootstrap.md): tag a deterministic **stage-0** commit → build → publish to **mavenLocal** (immutable by tag, NOT a SNAPSHOT) → the self-hosted reducer builds against it. No public Maven Central until self-bootstrapping is done.
2. **Flip the default** `--frontend=cpp`; libclang path kept selectable for one validation cycle.
3. **Delete** the libclang-C cinterop, the token-scrape block, and the ~200-line typedef-reducer (proven replaceable by `getCanonicalType()` — see the type-decode probe). The `-PenableClang` gate comes off: Clang-C++/LLVM becomes a mandatory dependency (intended end-state).
4. Cut the REAL release → becomes the next stage-0; steady-state takes over (always build from the previous released version; features reach the generator's own bindings one release behind).

Blocked by #46. Pre-requisite quality gates: #11 (hermeticity/determinism) must be closed before the stage-0 tag.

monkopedia-coder:
## Phase E step 2 (reversible flip-safety proof) — landed PR #76 (squash 0493fbc)

Additive gated harness (`:cppfrontend:featuregenCppBindings` + `-Pkpp.frontend.featuregen=cpp`), default path untouched, nothing deleted. **Result:**
- Generation GREEN: cpp front-end emits featuregen's full surface (116 files, all 17 instantiations, wrapper compiles).
- **186/188 featuregen test sources compile-clean against cpp-front-end bindings** — spelling-safety proven (no spelling delta in the 345-line parity diff renames a symbol any of those 186 callers use).

**Two concrete flip prerequisites surfaced (the value of the reversible step):**
1. #77 — `std::pair<int,int>` drops under the FULL multi-instantiation sync (per-unit OK) → CpPairTest. Binding-side cpp forcing-merge bug.
2. #78 — the cpp parse against system libstdc++ (gcc-16) surfaces newer internals K/N's bundled gcc-8.3 link can't resolve. libclang's lossiness was shielding featuregen from this skew; cpp's correctness exposes it. **needs-decision** on sysroot-alignment approach.

[...]

PR #69, opened 2026-06-12
cppfrontend+krapper_gen: D1c USR-tref ledger-accept + O1 forcing-order convergence — Phase D endpoint (#46)

monkopedia-coder:
Phase D's final convergence brick + the endpoint assessment. After this, every remaining parity diff line is a CITED, justified divergence where the cpp front-end is equal-or-more-correct than libclang.

## D1c — USR-tref overloads: LEDGER-ACCEPTED (with verified offsets)
set/unordered_set's surviving overloads spell the element by the param's **USR**. libclang's `assocTypedefElement` keys the reconstructed `value_type` tref by `WrappedTemplateRef(param.usr)`, and for these params `clang_getCursorUSR` falls back to USRGeneration's printLoc form `c:<header basename>@<byte offset>`. Exemplars VERIFIED against the dumps + the installed libstdc++ (gcc 16.1.1):
- `template_c_stl_set_h_3743` = byte 3743 of `bits/stl_set.h` = the `_Key` param of `set`
- `template_c_unordered_set_h_3824` = byte 3824 of `bits/unordered_set.h` = unordered_set's `_Value`

**Mirroring REJECTED:** (a) not cheap — the cpp front-end keys identity on `Decl::getID()` and has NO SourceLocation/SourceManager binding surface; mirroring needs that whole surface PLUS a byte-exact reimplementation of USRGeneration's printLoc fallback; (b) maximally version-fragile — the offsets are header BYTES, so any upstream edit above the template renames every generated symbol. Phase E makes cpp the stable reference, and a name-keyed tref (`template<_Key>`) is exactly the spelling the reference should keep. D1d (the `__const_template_..._and` disambiguation renames) collapses into this accept.

## O1 — forcing-round std-function ORDER: CONVERGED (the last uncited family)
The endpoint sweep found one uncited family. `ForcingHeader` prepends the target's std includes before the user header, so in a container-spec forcing TU `<new>` materializes lexically before `<cwchar>`. libclang's session-scoped USR memo (one Index across base + forcing parses) collapses the std namespace onto the base parse's first-seen order; the cpp per-payload forcing model kept the forcing TU's own order. **Fix:** `loadForcingModel` (Parsing.kt) reorders a forcing model's namespace-level children into the base model's first-seen order (unseen children append in TU order), mirroring the memo. This is a **cpp-forcing-load-path-only** seam — the libclang path is untouched (proven by zero featuregen drift).

## Ratchet — ALL 697 → 345 (−50%)
Every flipped unit −32 (O1). Final per-unit bounds: the vector/map/pair/Box family at **83**, set **129**, unordered_set **134**, unordered_map **91**, unique_ptr **248**, **ALL 345**. featuregenParity reports all 18 units **[as expected]** (0 improved / 0 regression / 0 fail).

## Phase D endpoint — every remaining line is cited
The endpoint sweep bucketed ALL remaining diff lines; **zero uncited residue**. The full ledger families (prose in `parity-expectations.txt`):
- **D2b** (libclang spelling-defect drops where cpp resolves correctly: `__locale_struct` every unit ~40 lines; `std::__exception_ptr` in unique_ptr/ALL) — ledger-accept, cpp more correct
- **D3** ledger-accept (parse-context-dependent libclang spellings, N5) + D1c
- **D4** (libclang reparents+loses Drawable's out-of-line virtual dtor; cpp keeps it) — ledger-accept, cpp more correct
The recurring theme: every "divergence" is either converged or a libclang defect we get to DELETE at the flip, not emulate.

## Verify (JDK 17)
- Full gate green: `:krapper_gen:nativeTest` 313/0, `:krapper_model:build`, `:featuregen:kplusplusSync` **zero krapped/ drift** (default-path purity — the Parsing.kt change is cpp-forcing-load only), `:featuregen:nativeTest` 188/0, `:krapper_gen:ktlintCheck`.
- Gated (`-PenableClang`): `goldenCompare`, `handoffGenerate`, `handoffInstDiff` (oracle byte-identical 9 files + functional), `featuregenParity` 18/18 [as expected], ALL=345.

Closes the Phase D convergence work; the parity oracle is satisfied for the flip decision (Phase E #47).

### DIFF
--- a/cppfrontend/parity-expectations.txt
+++ b/cppfrontend/parity-expectations.txt
@@ -99,8 +99,8 @@
 #   D4 out-of-line virtual destructor: the libclang cursor walk REPARENTS
 #      Drawable's dtor to the TU and loses it ("Can't find parent type"); the cpp
 #      path keeps it and emits dispose()/owned() (correct-by-construction).
-#   O1 forcing-round std-function ORDER (identified by the D1c endpoint sweep — the
-#      last family, the ONLY one that was uncited): in each instantiation ROUND's
+#   O1 forcing-round std-function ORDER — CONVERGED (D1c brick, 79474d0): the last
+#      family the endpoint sweep found, and the ONLY one that had been uncited. In each round's
 #      emitted "std Functions" block, the cpp path places set/get_new_handler before
 #      the wcs* group while libclang keeps base order. Mechanism: ForcingHeader
 #      prepends the target's std includes (e.g. `#include <vector>`) BEFORE the user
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
**Tier 1 — approved and merging.**

**Classification:** bug_fix (the O1 forcing-order correction is a correctness fix on the cpp path; D1c + ratchet are prose/ledger).

---

### O1 production-path purity — CONFIRMED

`loadForcingModel` is called exclusively from the `cppForcingModelPaths[target]?.let { ... }` branch in `IndexedServiceImpl.requestInstantiation`. `cppForcingModelPaths` is only populated in `App.kt` inside `if (frontend == Frontend.CPP)` — the libclang path is the `?: run { scope.parseHeader(...) }` branch which is structurally unreachable from this change. `cppBaseModelTu` is only ever set in `parseHeader()` when `cppParsedModelPath != null` (also the cpp path), so even hypothetically, a null base means `loadForcingModel` reduces to `loadParsedModel` (the `?.let` short-circuits). **Zero featuregen drift** — confirmed by running `kplusplusSync` fresh (17 instantiations regenerated, no git diff on featuregen/src/).

### O1 reorder correctness and determinism — CONFIRMED

`sortedBy` is Kotlin's stable sort — lossless (never drops or duplicates), deterministic (same input → same output). Elements not in `baseOrder` (unseen or null-keyed via `else -> null`) receive `Int.MAX_VALUE` and land after all base-seen children in their original relative order, exactly mirroring the libclang memo's "append new decls" semantic. `clearChildren()` + `addAllChildren(reordered)` is an established pattern in the codebase (identical pattern in `WrappedElement.kt:83-84`, `ModelResolution.kt:109`). The `addAllChildren` guard (`!children.contains(it)`) is satisfied since `children` was just cleared.

### D1c offset spot-check — VERIFIED

- `dd if=/usr/include/c++/16.1.1/bits/stl_set.h bs=1 skip=3743 count=60` → `typename _Key, typename _Compare = std::less<_Key>,` — byte 3743 IS `_Key` ✓  
- `dd if=/usr/include/c++/16.1.1/bits/unordered_set.h bs=1 skip=3800 count=80` → `template<typename _Value,` — byte 3824 IS `_Value` ✓  

The "mirroring rejected" reasoning is also correct: the cpp front-end has no SourceLocation/SourceManager surface, and enshrining header byte offsets in generated symbol names would make generated bindings version-fragile on every libstdc++ patch.

### Ratchet honesty — CONFIRMED

`featuregenParity` reported exact matches: ALL=345, set=129, uset=134, umap=91, map+vector variants=83, unique_ptr=248. Every bound is tight to the live run; no bound was loosened.

### Build results

- `:krapper_gen:nativeTest` 313/0 ✓
- `:krapper_model:build` ✓
- `:featuregen:kplusplusSync` zero krapped/ drift ✓
- `:featuregen:nativeTest` 188/0 ✓
- `:krapper_gen:ktlintCheck` ✓
- `goldenCompare` PASS ✓
- `handoffInstDiff` byte-identical (9 files) + functional ✓
- `featuregenParity` 18/18 [as expected], ALL=345, 0 regression ✓

Quality is clean: new code follows established patterns, comments match the file's density, `firstSeenKey`/`reorderToFirstSeen` naming is apt, `cppBaseModelTu` follows the CLI-scoped-global convention of `cppParsedModelPath`/`cppForcingModelPaths`.

PR #70, opened 2026-06-12
krapper_gen: hermeticity globals audit + reset the cursor->element parse memo per run (#11a)

monkopedia-coder:
Part **a** of #11 (self-bootstrap hermeticity): the residual **process-global state audit** for flip-gating determinism. A self-hosted stage-0 must generate identical output every run, and in service mode (`:slice`, the v2 plugin's `kplusplusSync`) one krapper_gen process is reused across runs — so any process-global that carries data across runs or makes output depend on run history is a determinism leak.

Scope is **#11a only** (globals). `#11b` (llvm-config) and `#11c` (runbook) remain.

## Audit — every mutable process-global in `krapper_gen` + `krapper_model` nativeMain

| Global | Where | Across-run carrier? | Reset/scoped? | Disposition |
|---|---|---|---|---|
| `GenerationContext` (`internedTypes` map, `rootPackage`) | krapper_model | yes | `reset()` in `IndexedServiceImpl.init` (PR #34) | already correct |
| `DropLedger` (`records`) | krapper_gen | yes | `reset()` in `IndexedServiceImpl.init` | already correct |
| `elementLookup` (cursor→element memo) | `ModelFactories.kt` | **yes — never cleared** | **NO** | **LEAK → fixed** |
| `droppedUsrs` | `ModelFactories.kt` | overwritten before each `mapAll` via `setDroppedUsrs` | set-before-use | not a leak (also reset defensively) |
| `Log.loggerImpl` | `Logging.kt` | logger sink, set at startup | config, not per-run data | not a leak |
| `dumpParsedModelPath`, `cppParsedModelPath`, `cppForcingModelPaths`, `dumpModelsDir`, `dumpModelCounter`, `cppBaseModelTu` | `Parsing.kt` | CLI/`--frontend=cpp`-only | null/default in service mode; service never reads them; ksrpc schema untouched | not a leak (debug/handoff path; one process = one CLI invocation) |
| `roundTripParsedModel` | `Parsing.kt` | env-derived (`KRAPPER_ROUNDTRIP_MODEL`) | deterministic constant, identical every run | not a leak (it's an input) |
| `Utils`, `FixupApplier`, `ForcingHeader` objects | both | — | stateless (no `var`/mutable fields) | not a leak |
| `WrappedTemplate.templateArgCounter` etc. | krapper_model | per-instance field, not global | — | not a leak |
| Service-impl fields (`forcingHeaders`, `requested`, `forcedContainerKeys`, `tempDir`, …) | `IndexedServiceImpl` | per-`IndexedServiceImpl` instance = per run | new instance each run | not a leak |

## The leak — `elementLookup`

`ModelFactories.kt` interns parsed `WrappedElement`s by USR-tag so a declaration re-encountered across the base parse and every forcing re-parse **within one run** collapses onto one instance (resolution then mutates that instance in place — metadata, synthetic children, dedup). The map was **process-global and never cleared**: in service mode a later run re-parsing the same header finds the prior run's already-resolved, mutated instances and the map grows unboundedly. The intended scope is one run's parses, not the process.

**Fix:** new `WrappedElement.resetElementMemo()` (clears `elementLookup`, and resets `droppedUsrs` defensively), wired into `IndexedServiceImpl.init` right next to the existing `DropLedger.reset()` / `GenerationContext.reset()`.

Transparency note: I could not exhibit an *output* divergence from this leak — resolution is idempotent over a reused element in every scenario I constructed (std::vector<int> instantiation; a fixed-path header with const-overload dedup + INCLUDE_MISSING synthetic pull-in). So this is a hermeticity hardening (removes cross-run element identity + unbounded growth in the reused service process), not a fix for an observed mis-generation.

## Determinism test

`DeterminismTest.twoInProcessRunsAreByteIdentical` runs two **full** generations (`filterAndResolve` + `writeTo`) in one process over a fixed-path header (USRs collide across runs, so the memo is genuinely re-hit) and asserts byte-identical output across every generated file. The per-invocation output dir is normalized out of the `.def` comparison (libraryPaths / `-I` are an input, not leaked state). This is the determinism regression guard the stage-0 flip depends on; any future global that leaks **and** affects output is caught here.

## Verify (JDK 17, foreground)

`:krapper_gen:nativeTest` **314/0** (+the new test), `:krapper_model:build`, `:featuregen:kplusplusSync` **zero `krapped/` drift**, `:featuregen:nativeTest` **188/0**, `:krapper_gen:ktlintCheck` green. Plugin IPC schema (KrapperConfig / IndexRequest / service interfaces) untouched, so no plugin recompile needed.

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -42,6 +42,7 @@ import com.monkopedia.krapper.generator.codegen.KotlinWriter
 import com.monkopedia.krapper.generator.codegen.NameHandler
 import com.monkopedia.krapper.generator.model.ForcingHeader
 import com.monkopedia.krapper.generator.model.WrappedClass
+import com.monkopedia.krapper.generator.model.WrappedElement
 import com.monkopedia.krapper.generator.model.kotlinType
 import com.monkopedia.krapper.generator.model.type.WrappedType
 import com.monkopedia.krapper.generator.resolvedmodel.ResolvedClass
@@ -108,13 +109,18 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
             index.dispose()
         }
         // Start each generation run from clean process-scoped state so its output and
-        // report reflect only this run: the drop ledger (shared with any prior run) and
-        // the GenerationContext (the WrappedType intern cache + the rootPackage). Both
-        // must be reset BEFORE any resolution (requestInstantiation / filterAndResolve):
+        // report reflect only this run: the drop ledger (shared with any prior run), the
+        // GenerationContext (the WrappedType intern cache + the rootPackage), and the
+        // parse memo (the cursor->element interning map). All must be reset BEFORE any
+        // resolution (requestInstantiation / filterAndResolve):
         // a type's Kotlin package is baked into its ResolvedKotlinType at resolve time,
         // so rooting it later (e.g. in writeTo) would be too late.
         DropLedger.reset()
         GenerationContext.reset(config.rootPackage)
+        // The libclang-path cursor->element memo: cleared so a reused service process
+        // (kplusplusSync) never builds a later run on a prior run's resolved/mutated
+        // elements (#11 hermeticity).
+        WrappedElement.resetElementMemo()
     }
 
     private val includePaths = generateIncludes(config.compiler)
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelFactories.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ModelFactories.kt
@@ -75,6 +75,24 @@ fun WrappedElement.Companion.setDroppedUsrs(usrs: Set<String>) {
 }
 
 /**
+ * Clear the process-global parse memo before a fresh generation run (#11 hermeticity).
+ *
+ * [elementLookup] interns parsed elements by USR-tag so a declaration re-encountered
+ * across the base parse and every forcing re-parse WITHIN one run collapses onto a single
+ * instance — resolution then mutates that instance in place (metadata, synthetic children,
+ * dedup). That identity must NOT survive across runs: in service mode the v2 plugin's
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
Tier 1 — approve and merge.

**Reset placement:** `WrappedElement.resetElementMemo()` is called in `IndexedServiceImpl.init` (Kotlin `init` block — fires at construction time). Confirmed via `KrapperServiceImpl.kt:42`: `override suspend fun index(request: IndexRequest): IndexedService = IndexedServiceImpl(getConfig(Unit), request)` — a fresh `IndexedServiceImpl` is created per service call, so the init fires once per run before any parsing. Covers service mode correctly.

**Leak analysis honesty:** the coder's transparency about not exhibiting an output divergence (resolution proved idempotent) is correct and appropriate. The fix is still justified on two independent grounds: (a) cross-run element identity is a semantic-correctness defect even if resolution today happens to be idempotent (this could change), and (b) unbounded `elementLookup` growth in a long-lived service process is a real operational defect regardless.

**Audit completeness (independently verified):** grepped all top-level `var` and `object` singletons across both modules. The table covers them all. Specifically: `WrappedTemplate.templateArgCounter` is a per-instance field reset to 0 in `mapAll` at line 128 (not a global); `NameHandler` is instantiated fresh in `writeTo` (not a global). The `Parsing.kt` CLI-only debug vars (`dumpParsedModelPath`, `roundTripParsedModel`, etc.) are constant in service mode — correct classification.

**DeterminismTest is meaningful:** uses a fixed-path header (`/tmp/krapper_determinism_input.h`) so USRs are identical across both runs and the `elementLookup` memo is genuinely re-hit; the class shape (const-overload dedup + `INCLUDE_MISSING` synthetic pull-in) maximizes the chance of non-idempotent mutation surfacing; the only expected per-run difference (output dir path in the `.def` file) is normalized out. This is a forward guard — it locks the hermeticity property even if resolution becomes non-idempotent in a future change.

**droppedUsrs reset:** defensive, harmless, and correctly documented (set-before-use by `setDroppedUsrs` but resetting it here removes the cross-run window).

**Build:** `krapper_gen:nativeTest` 314/0 (DeterminismTest passes), `krapper_model:build` green, `featuregen:kplusplusSync` 17 instantiations zero drift, `featuregen:nativeTest` 188/0, `krapper_gen:ktlintCheck` green.

PR #71, opened 2026-06-12
build: harden the llvm-config / -PenableClang clang-gate probe (#11b)

monkopedia-coder:
Part **b** of #11 (self-bootstrap build-infra): harden the `llvm-config` / `-PenableClang` probe that gates the clang modules (`:clangwalk`, `:cppfrontend`). Small, gate/probe-only scope. (#11a globals done; #11c runbook is a sibling's.)

## Probe audit — what was hardcoded/fragile
- `llvm-config` was **never invoked**. Everything was hardcoded: `-lLLVM-22` and `-L/usr/lib` literally, in **two** build files each (`clangwalk/build.gradle.kts`, `cppfrontend/build.gradle.kts`). An LLVM major bump (22→23) or a non-`/usr/lib` install meant editing 2 files in 2 places.
- `-PllvmConfig` was only used in a `canExecute()` **gate check** — its value was never threaded anywhere, so the documented "point at a specific install" override didn't actually point the link at anything.
- **No actionable guard:** `-PenableClang` with an absent/mis-pointed toolchain failed ~8 min into a release link with a cryptic `cannot find -lLLVM-22`. (The box even carries a stale `libLLVM-15.so` next to 22 — the wrong-version hazard is real.)

## What I hardened (settings.gradle.kts)
- **Probe once, thread discovered values:** when the modules are enabled, resolve `llvm-config` (`-PllvmConfig` override, else PATH) + `clang++`, then discover the LLVM major (`--version`) and libdir (`--libdir`) and thread them to both modules via `rootProject` extras. Both modules now emit `-L${llvmLibDir}` / `-lLLVM-${llvmMajor}` with `/usr/lib` + `22` fallbacks. An LLVM bump or relocated install needs **no per-module edit**.
- **Actionable failure at the gate:** enabled but no `llvm-config`/`clang++` → fail at **configure time** (not a late linker error) naming exactly what's missing and how to set `-PllvmConfig` (or drop the flag for the default build).
- **`-PllvmConfig` is now genuinely used** (threaded), the documented override — and is its own opt-in.
- **Soft WARNING** if the expected libs aren't under the discovered libdir (the "LLVM present but wrong version" hint, before the linker).

## Deliberately left
- Did **not** parse `llvm-config --libs` / emulate its full link line — the explicit `-lclang-cpp -lLLVM-<major>` shape is kept (minimal, matches the modules). Lib-existence is a soft warning, not a hard fail, to avoid false negatives across distros that name the monolithic lib differently (the linker stays the authority).
- Did **not** touch the default build, #11a globals, or the #11c runbook.

## Failure-message demo (`-PllvmConfig=/nonexistent`)
```
The clang modules (:clangwalk, :cppfrontend) are enabled (-PllvmConfig=/nonexistent) but the
LLVM/Clang toolchain they link against was not found:
  - llvm-config: '/nonexistent' is not an executable file
  - clang++:     /usr/bin/clang++

Fix one of:
  * install LLVM/Clang so `llvm-config` and `clang++` are on PATH; or
  * point the build at a specific install:
        ./gradlew <task> -PllvmConfig=/path/to/llvm-config
  * to build WITHOUT the clang modules, just drop -PenableClang/-PllvmConfig
    (the default build excludes them and needs no LLVM).
```
Fires in ~350 ms at configure time, not ~8 min into a link.

## Verify (JDK 17)
- **Default unaffected:** `./gradlew projects` → clang modules ABSENT; `:krapper_gen:nativeTest :krapper_gen:ktlintCheck` GREEN.
- **Gate works:** `-PenableClang` → both modules PRESENT (probe finds 22.1.6 / `/usr/lib`, no warning); valid `-PllvmConfig=/usr/bin/llvm-config` → PRESENT; `:clangwalk:tasks -PenableClang` configures cleanly with the threaded extras.
- **Failure path:** `-PllvmConfig=/nonexistent` → actionable error above (configure-time).
- **Link line unchanged on this host:** discovered values (major=22, libdir=/usr/lib) are byte-identical to the prior literals, so the emitted `-l`/`-L` flags are textually unchanged.
- Note: a full gated `:clangwalk:linkReleaseExecutableKlinker` currently fails **upstream of linking**, at `compileKotlinNative`, on regenerated bindings importing `clang.attr.Kind`/`.value` (a krapper_gen codegen gap vs the box's LLVM-22.1.6 headers). That is independent of this change (the Kotlin compile step never reads `compilerOpts`) and out of scope for the gate/probe brick.

### DIFF
--- a/clangwalk/build.gradle.kts
+++ b/clangwalk/build.gradle.kts
@@ -44,10 +44,14 @@ kotlin {
                 // The generated wrapper calls into Clang's C++ AST API, so the final
                 // executable links libclang-cpp (the monolithic LLVM/Clang shared lib)
                 // + LLVM core. -lstdc++ for the C++ wrapper runtime.
+                // libdir + LLVM major are DISCOVERED by settings.gradle.kts's #11(b) probe
+                // (llvm-config --libdir / --version) and threaded via rootProject extras,
+                // so an LLVM bump or non-/usr/lib install needs no edit here; the literal
+                // fallbacks match this host and only apply if the probe didn't run.
                 compilerOpts(
-                    "-L/usr/lib",
+                    "-L${rootProject.extra.properties["llvmLibDir"] ?: "/usr/lib"}",
                     "-lclang-cpp",
-                    "-lLLVM-22",
+                    "-lLLVM-${rootProject.extra.properties["llvmMajor"] ?: "22"}",
                     "-lstdc++",
                     "-lm",
                     "-lpthread"
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -36,10 +36,14 @@ kotlin {
                 // versions, --gc-sections so DEBUG dead-strips the stale platform.linux
                 // cinterop cache, and libclang-cpp + LLVM for the Clang C++ AST API.
                 linkerOpts("--gc-sections")
+                // libdir + LLVM major are DISCOVERED by settings.gradle.kts's #11(b) probe
+                // (llvm-config --libdir / --version) and threaded via rootProject extras;
+                // the literal fallbacks match this host and only apply if the probe didn't
+                // run. (See :clangwalk's build.gradle.kts for the full link rationale.)
                 compilerOpts(
-                    "-L/usr/lib",
+                    "-L${rootProject.extra.properties["llvmLibDir"] ?: "/usr/lib"}",
                     "-lclang-cpp",
-                    "-lLLVM-22",
+                    "-lLLVM-${rootProject.extra.properties["llvmMajor"] ?: "22"}",
                     "-lstdc++",
                     "-lm",
                     "-lpthread"
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -21,17 +21,96 @@ include(":featuregen")
 
 // :clangwalk is the stage1 self-bootstrap harness — it binds Clang's own C++ AST
 // (libclang-cpp) and walks a real AST on the generated bindings. It REQUIRES an LLVM
-// toolchain on the box (system clang++, libclang-cpp, LLVM-22) that an ordinary build
+// toolchain on the box (system clang++, libclang-cpp, libLLVM) that an ordinary build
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Tier-1 auto-review. Reviewed to merge standard.

**Default-build purity: CONFIRMED.** `./gradlew projects` (no flag) → `:clangwalk`/`:cppfrontend` absent; `:krapper_gen:nativeTest :krapper_gen:ktlintCheck` green. No llvm-config probe when flag is absent — the entire block is gated inside `if (clangEnabled)`.

**Value equivalence on host: CONFIRMED.** `llvm-config --version` → 22.1.6 (major=22), `llvm-config --libdir` → /usr/lib — byte-identical to the prior hardcoded literals, so the emitted link line is unchanged on this host. Fallbacks (`?: "22"`, `?: "/usr/lib"`) preserve behavior if probe returns empty.

**Actionable failure: CONFIRMED.** `-PllvmConfig=/nonexistent` → configure-time `GradleException` at line 66 in 362ms (not 8min into a link), naming the missing executable and how to fix it.

**Override threads: CONFIRMED.** `-PllvmConfig=/usr/bin/llvm-config` → both modules present, probe runs with the override.

**No MissingProperty: CONFIRMED.** `:clangwalk:tasks -PenableClang` → BUILD SUCCESSFUL, no extra property errors.

**clangwalk link caveat is INDEPENDENT: CONFIRMED.** This PR only touches link-flag plumbing (`-L`/`-lLLVM-` substitution) and the gate/probe in `settings.gradle.kts`. The `compileKotlinNative` failure on regenerated bindings importing `clang.attr.Kind`/`.value` is a krapper_gen codegen gap vs LLVM-22.1.6 — the Kotlin compile step never reads `compilerOpts`, so it is independent of this change and correctly out of scope. Tracking separately.

**Code quality:** Clean, idiomatic Kotlin/Gradle throughout. `resolveExecutable`/`probe` helpers are appropriately scoped, well-named, and the comment density matches the surrounding infra code. The semantic improvement to `clangEnabled` (bare presence of `-PllvmConfig` now enables rather than silently ignoring a bad path) is correct and documented.

PR #74, opened 2026-06-12
fix(krapper_gen): declaration name must ignore per-file import-alias remap (#73)

### LINKED ISSUE #73: [clang-campaign] clangwalk release generation broken on LLVM-22.1.6: generated bindings declare enum class AttrKind but import nonexistent clang.attr.Kind/.value (shared-remap leak)
## Summary
The gated flagship self-bootstrap walk `:clangwalk:runReleaseExecutableKlinker -PenableClang` (and `:clangwalk:linkReleaseExecutableKlinker`) is **RED on this box's LLVM-22.1.6 headers**. It fails at the `:clangwalk:compileKotlinNative` step (NOT linking): the freshly **regenerated** clangwalk bindings declare the `clang::attr::Kind` enum under one simple name but reference it under another, so the generated Kotlin does not compile.

Campaign: part of the clang self-bootstrap (#46/#47). The flagship `runReleaseExecutableKlinker` walk is currently broken.

## Exact error
```
> Task :clangwalk:compileKotlinNative FAILED
e: .../clangwalk/krapped/src/clang_ASTContext.kt:9:19  Unresolved reference 'Kind'.
e: .../clangwalk/krapped/src/clang_ASTContext.kt:645:91 Unresolved reference 'AttrKind'.
e: .../clangwalk/krapped/src/clang_ASTContext.kt:647:60 Unresolved reference 'value'.
```
(Reproduced with `GRADLE_OPTS=-Xmx12g JAVA_HOME=jdk17 ./gradlew :clangwalk:runReleaseExecutableKlinker -PenableClang`. `clangwalk/krapped/` is gitignored generated output, regenerated by `:clangwalk:kplusplusSync`.)

## Root cause — a long-latent shared-state / determinism codegen bug in krapper_gen
The generated `clang_ASTContext.kt` writes:
```kotlin
import clang.attr.Kind as AttrKind                                  // line 9
...
inline fun getAttributedType__clang_attr_Kind_clang_QualType_clang_QualType(
        attrKind: AttrKind, modifiedType: QualType, equivalentType: QualType): QualType {
    clang_ASTContext_get_attributed_type(ptr, attrKind.value, ...)  // line 647
}
```
i.e. it imports `clang.attr.Kind` (aliased locally to `AttrKind`) and uses `attrKind.value`.

But the freshly regenerated enum DECLARATION file `clang_attr_Kind.kt` declares:
```kotlin
package clang.attr
enum class AttrKind(val value: UInt) { ... }      // <-- declared name is AttrKind, NOT Kind
```
So the actual generated type is `clang.attr.AttrKind`; `clang.attr.Kind` does not exist. Hence the import (`Kind`), the param type alias (`AttrKind`, unresolved because the import failed), and `.value` (on an unresolved receiver) all fail.

**Why the two sides disagree** (the bug):
- `ImportBlock.build()` (`krapper_gen/.../builders/ImportBlock.kt`) is the PER-FILE import block. When a file references two types whose simple name is `Kind` (here `clang::attr::Kind` plus another `*::Kind` imported in `clang_ASTContext.kt`), `handleDuplicates`/`selectName` disambiguates by namespace-prefixing -> `AttrKind`, emits `import clang.attr.Kind as AttrKind`, and at lines 49-55 calls `setNameRemap(mappings)` on **every** `FqSymbol` in the file so use-sites render `AttrKind`. Correct *within that file*.
- BUT `remap` is **mutable per-instance state** on a **shared** `ResolvedKotlinType` (`resolvedmodel/ResolvedType.kt`: `private var remap`; `name`/`plainName` -> `mappedName` = `remap?.get(fullyQualified) ?: qualifyList.last()`). The mutation persists on the shared enum-type instance.
- `KotlinWriter.enumDeclarations` (KotlinWriter.kt:269) emits the enum's OWN declaration as `enum class ${enum.plainName}`. `plainName` honors the leaked `remap`, so the declaration becomes `AttrKind`. Meanwhile the FILE NAME and `package`/import-target are derived from `fullyQualified` (`clang.attr.Kind` -> `clang_attr_Kind.kt`, package `clang.attr`), which is NOT remapped. Declaration name (`AttrKind`) and canonical identity (`clang.attr.Kind`) diverge -> broken.

In short: a *file-local import alias* leaks, via shared mutable `remap`, into the *global declaration name* of a nested-namespace enum. A type's DECLARATION must always use its canonical simple name (`qualifyList.last()` = `Kind`), never a per-file import alias.

## Which bound binding pulls `clang::attr::Kind` in
- `clang::ASTContext::getAttributedType(attr::Kind, QualType, QualType)` — `clang::ASTContext` is in clangwalk's `only()` set (clangwalk/build.gradle.kts).
- `clang::Type::hasAttr(attr::Kind)` — pulls it into the `Type`/`FunctionType`/`FunctionProtoType` files too.
The `clang::attr::Kind` enum (475 entries on LLVM-22.1.6, `clang/Basic/AttrKinds.h`) is auto-materialized as a `WrappedEnumType` because it appears as a bound method's parameter type.

## When it regressed
The faulty machinery is OLD, not a recent codegen regression: the import-aliasing + `setNameRemap` dates to 2022 (086ff73); the `remap`/`mappedName` leak predates the campaign; #31 only refactored `name.trimEnd('?')` into `plainName` (behavior-preserving). What is NEW is the TRIGGER + visibility:
- The bug is **order/surface-dependent (a determinism hazard)** — it fires only when some file imports `clang.attr.Kind` alongside another `Kind` in the same file, poisoning the shared instance before/while the enum declaration is emitted. Evidence: cppfrontend's freshly regenerated `clang_attr_Kind.kt` (mtime within the same window) declares the **consistent** `enum class Kind` + `import clang.attr.Kind as AttrKind` and compiles; clangwalk's regen declares `enum class AttrKind` and does not. Same generator, different binding surface/emission order -> different declared name.
- `:clangwalk:runReleaseExecutableKlinker` is **NOT in the standard gate** (only cppfrontend's goldenCompare/handoffGenerate/handoffInstDiff/featuregenParity are gated, and those compare reducer/handoff C++ output — they do not K/N-compile the regenerated Kotlin bindings). So a regression in clangwalk's fresh-regen compile is invisible to CI and went unnoticed. PR #42 (methods/bases walk) is the last clangwalk change; it added nothing attr-related — `ASTContext` was already bound.

## Does it affect cppfrontend? (NO, currently — but latently at risk)
cppfrontend's `only()` set ALSO binds `clang::ASTContext`, `clang::Type`, and `clang::FunctionProtoType` — i.e. it binds the SAME `attr::Kind`-pulling methods. So this is **not** a clangwalk-only surface issue. cppfrontend is GREEN today only because (a) its current regen happened to emit the consistent `enum class Kind`, and (b) its gated suite does not K/N-compile the bindings. A different emission order would break cppfrontend identically. Classify: **general krapper_gen codegen/determinism bug**, surfaced first on clangwalk because clangwalk is the only path that compiles freshly-regenerated bindings binding `attr::Kind`.

## Fix recommendation
Prefer **(b) a general krapper_gen fix** over a clangwalk-local patch:
- **(b) General (recommended):** make a type's DECLARATION name independent of the per-file import remap. Either emit the enum/type declaration from the canonical `qualifyList.last()` (bypass `plainName`/`mappedName` for declarations), or stop `setNameRemap` from mutating shared model instances (scope the alias map to the single file build / clone-on-remap). This kills the determinism hazard for ANY nested `X::Kind` enum bound alongside another `Kind` and protects cppfrontend. Aligns with the determinism/hermeticity work (#11, #70).
- **(a) Immediate clangwalk unblock (optional, if a fast green flagship is wanted):** add a `fixup { removeMethod(...) }` for `clang_ASTContext_get_attributed_type__clang_attr_Kind_*` and `clang_Type_has_attr` (same pattern as cppfrontend's APSInt operator fixups) to drop the `attr::Kind`-pulling methods clangwalk does not need for the methods/bases walk. Leaves the general bug in place.

Recommend (b); (a) only if the flagship demo must go green before (b) lands.

monkopedia-coder:
Fixes #73.

## Confirmed mechanism
`ImportBlock.build` disambiguates a same-file import collision (two `Kind` imports — `clang.attr.Kind` next to another `Kind`) by aliasing one to `AttrKind` and calling `setNameRemap(...)`, which stamps the alias onto `remap` — **mutable state on a SHARED `ResolvedKotlinType`**. `plainName -> name -> mappedName` returns `remap[fullyQualified] ?: qualifyList.last()`.

The enum's own declaration file is rendered by `KotlinWriter.renderEnum` (a plain string builder with no ImportBlock), yet it read `enum.plainName` off that **same shared instance** — which a *different* file's collision had already mutated. Result: `enum class AttrKind` while the file name, package, and every reference kept `clang.attr.Kind` → unresolved reference, K/N `compileKotlinNative` failed. The class-declaration token (`named(type)` → `type.name`) carries the identical hazard.

This is also a hermeticity bug (#11): because `remap` is shared mutable state, the corruption is render-order dependent.

## Fix — declaration uses canonical, references keep alias
- Added canonical `declaredName` / `plainDeclaredName` on `ResolvedKotlinType` that resolve to `qualifyList.last()` and **ignore `remap`**.
- Declaration sites now use them: the `enum class X` token (`renderEnum`) and the `class X` token (`named`).
- References still go through `name` / `plainName`, so an aliased import legitimately resolves under its alias.
- Determinism: `ImportBlock.build` now applies its per-file alias map **unconditionally** (even when empty) so a collision-free file resets any stale alias a previously-rendered file left on the shared instance — reference output is now independent of file render order. Byte-identical when there are no collisions.

## Verification (JDK 17)
- `:krapper_gen:nativeTest` **315/0** (314 + new `DeclNameRemapTest` locking the invariant: a `remap` renames references but not the declaration), `:featuregen:nativeTest` **188/0**, `:krapper_gen:ktlintCheck` green.
- **featuregen output byte-identical** — zero `krapped/` drift (clean `git status`; featuregen's specs hit no attr collision).
- **Flagship clangwalk now GREEN** — `:clangwalk:runReleaseExecutableKlinker -PenableClang` previously failed at `compileKotlinNative` on `clang.attr.Kind`; it now compiles, links, and runs the real libclang-cpp AST walk:
```
clangwalk: top-level declarations of the parsed TU
  [CXXRecord] Circle
      <base> Shape
      <method> radius : double
  [Var] globalVar : int
clangwalk: walked 10 top-level declarations via real libclang-cpp
```
- `:cppfrontend:goldenCompare :cppfrontend:featuregenParity -PenableClang` still green (the pre-characterized `[as expected]` diffs only; 0 fail).

### DIFF
--- a/krapper_gen/src/commonMain/kotlin/resolvedmodel/ResolvedType.kt
+++ b/krapper_gen/src/commonMain/kotlin/resolvedmodel/ResolvedType.kt
@@ -168,6 +168,24 @@ data class ResolvedKotlinType(
         } else {
             "$mappedName${if (isNullable) "?" else ""}"
         }
+
+    // The CANONICAL declaration name — this type's own identity, INDEPENDENT of any
+    // per-file import alias (issue #73). `name`/`plainName` resolve through `mappedName`,
+    // which applies `remap`; `remap` is a per-FILE REFERENCE-site rename (how one file
+    // refers to a colliding import it had to alias, e.g. `clang.attr.Kind` -> `AttrKind`),
+    // set as SHARED MUTABLE STATE on this instance by `ImportBlock.build`. A DECLARATION
+    // (the `class X` / `enum class X` token, plus its file name and package) must NOT
+    // depend on that: `remap` set while emitting one file would otherwise leak into the
+    // declared name emitted for this same shared type in another file (a non-hermetic,
+    // order-dependent corruption). So declarations use this canonical name; only
+    // REFERENCES go through `name`/`plainName` and keep the alias.
+    val declaredName: String
+        get() = if (templates.isNotEmpty()) {
+            "${qualifyList.last()}<${templates.joinToString(", ") { it.name }}>" +
+                if (isNullable) "?" else ""
+        } else {
+            "${qualifyList.last()}${if (isNullable) "?" else ""}"
+        }
     val pkg: String
         get() = qualifyList.subList(0, qualifyList.size - 1).joinToString(".") {
             it.replaceFirstChar { it.lowercase() }
@@ -213,6 +231,12 @@ fun nullable(base: ResolvedKotlinType): ResolvedKotlinType = base.copy(isNullabl
 // a `<Name>Api`/`to<Name>` prefix needs the bare name without it.
 val ResolvedKotlinType.plainName: String
     get() = name.trimEnd('?')
+
+// [declaredName] with the use-position nullable `?` stripped — the bare declaration token
+// (`enum class X` / `class X` is illegal with a trailing `?`). Mirrors [plainName], but
+// canonical: a declaration name must never carry a per-file import alias (issue #73).
+val ResolvedKotlinType.plainDeclaredName: String
+    get() = declaredName.trimEnd('?')
 
 fun ResolvedKotlinType.typedWith(parseTypes: List<ResolvedKotlinType>): ResolvedKotlinType =
     copy(templates = parseTypes)
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/builders/ImportBlock.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/builders/ImportBlock.kt
@@ -48,10 +48,15 @@ class ImportBlock(pkg: String, private val target: CodeBuilder<KotlinFactory>) :
         }
         val mappings = fqNames.mapNotNull { item -> item.second?.let { item.first to it } }
             .toMap()
-        if (mappings.isNotEmpty()) {
-            fqSymbols.forEach {
-                it.setNameRemap(mappings)
-            }
+        // Apply this file's alias map to every symbol in the file UNCONDITIONALLY — even
+        // when empty. `remap` is shared mutable state on a `ResolvedKotlinType`, so a type
+        // referenced here may still carry an alias set while a PREVIOUS file (with a
+        // collision) was emitted. Re-applying this file's map (empty == no aliases) resets
+        // that stale alias, so a reference renders by the alias only in files that actually
+        // need it — making reference output independent of file render order (issue #73 /
+        // #11 hermeticity). Declarations are already canonical via `declaredName`.
+        fqSymbols.forEach {
+            it.setNameRemap(mappings)
         }
     }
 
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -102,6 +102,7 @@ import com.monkopedia.krapper.generator.resolvedmodel.type.ResolvedKotlinType
 import com.monkopedia.krapper.generator.resolvedmodel.type.ResolvedType
 import com.monkopedia.krapper.generator.resolvedmodel.type.fullyQualifiedType
 import com.monkopedia.krapper.generator.resolvedmodel.type.nullable
+import com.monkopedia.krapper.generator.resolvedmodel.type.plainDeclaredName
 import com.monkopedia.krapper.generator.resolvedmodel.type.plainName
 import com.monkopedia.krapper.generator.resolvedmodel.type.typedWith
 
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -264,9 +264,14 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
     // fixed boilerplate with no AST nodes to model.
     private fun renderEnum(enum: ResolvedKotlinType): String {
         val underlying = enum.enumUnderlying?.name ?: "Int"
-        // The enum's DECLARATION name must be the bare class name (`plainName`): an
-        // `enum class TranslationUnitKind?` is illegal. Type USAGES keep the `?`.
-        val name = enum.plainName
+        // The enum's DECLARATION name must be the bare CANONICAL class name
+        // (`plainDeclaredName`, not `plainName`): `plainName` routes through the shared,
+        // per-file import-alias `remap`, so another file aliasing this enum (e.g.
+        // `clang.attr.Kind` -> `AttrKind` to dodge a same-file `Kind` collision) would leak
+        // that alias into this declaration — `enum class AttrKind` while the file/package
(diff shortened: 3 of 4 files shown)


monkopedia-reviewer (approved):
## Review: fix(krapper_gen): declaration name must ignore per-file import-alias remap (#73)

**Tier 1 — auto-approve and merge** (kplusplus all-tier-1 override)

### Declaration-vs-reference audit

Every declaration site examined:

- **`renderEnum` (`enum class X` token)** — ✅ fixed via `plainDeclaredName` (ignores `remap`)
- **`named` (`class X` token)** — ✅ fixed via `declaredName`
- **`generateBaseInterfaces` (`interface ShapeApi`)** — ✅ safe: `interfaceName` = `kotlinName + "Api"` is computed in `computeBaseInterfaces()` BEFORE any file is rendered (remap=null → canonical), stored as a plain String in the `BaseInterface` data class, never re-read through `plainName`
- **Template facade declarations (`interface Box<T>`)** — ✅ safe: name comes from C++ type-string parsing (`facade.base`), never `plainName`
- **Constructor factory / `_Holder` names** — ✅ safe: both are emitted AFTER `importBlock` applies the declaring file's (empty) remap, so `plainName` returns canonical in that context

**One pre-existing latent (not introduced here):** `constructorMethod` at KotlinWriter.kt:2544 uses `type.plainName` when building a REFERENCE to `Kind_Holder` from a different file. If that file aliases `Kind` → `AliasedKind`, the call becomes `AliasedKind_Holder` (undeclared). This pre-existed before this PR, only triggers for CLASS (not enum) ARG_CAST returns with import collisions, and never applies to the enum case being fixed here. Non-blocking; worth a follow-up ticket.

### Determinism reset correctness

`ImportBlock.build` applying the (possibly empty) alias map unconditionally is correct. The render loop is sequential; each file's `importBlock` is the first symbol added, so its `build()` runs before the file body is rendered. `findFqSymbols` traverses the complete pre-populated symbol tree. The per-file reset writes `emptyMap()` when no collision exists, clearing any stale alias from a prior file — making reference rendering render-order independent. No race: files render one at a time.

### Test

`DeclNameRemapTest` exercises all three states: pre-remap identity, reference alias, declaration canonical. Would fail pre-fix. 1/1 passing.

### Build verification

1. **Gate** (`:krapper_gen:nativeTest :krapper_model:build :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck`) → **BUILD SUCCESSFUL** — 315/0 (includes new DeclNameRemapTest), 188/0 featuregen, ktlintCheck green, featuregen output byte-identical (clean git status)
2. **Flagship clangwalk** (`:clangwalk:runReleaseExecutableKlinker -PenableClang`) → **BUILD SUCCESSFUL** — `walked 10 top-level declarations via real libclang-cpp` (previously failed at compileKotlinNative on clang.attr.Kind)
3. **cppfrontend** (`:cppfrontend:goldenCompare :cppfrontend:featuregenParity -PenableClang`) → **BUILD SUCCESSFUL** — goldenCompare PASS, featuregenParity 0 fail (all 18 units [as expected])

PR #76, opened 2026-06-12
featuregen cpp-front-end flip-safety proof (#47, Phase E step 2)

monkopedia-coder:
## Phase E step 2 (#47): the REVERSIBLE behavioral flip-safety proof

Additive, gated harness that runs featuregen's behavioral golden gate (the 188 `nativeTest` tests, exercising the GENERATED bindings at runtime) against bindings produced by the **cpp front-end** instead of libclang — measuring whether the ~345-line Phase-D ALL parity diff renames anything a test calls.

**Reversible only — nothing flipped, nothing deleted:** no default changed, libclang / cinterop / token-scrape / typedef-reducer all untouched, krapped/ has ZERO drift. The actual flip remains a separate, human-approved step.

### Harness mechanics (all gated behind `-PenableClang` + `-Pkpp.frontend.featuregen=cpp`)
- **`:cppfrontend:featuregenCppBindings`** — materializes featuregen's FULL surface (all 17 instantiations at once, the real `kplusplusSync` shape) through `krapper_gen --frontend=cpp` over cppfrontend-emitted models (reusing the existing `parityEmit`), with the SAME CLI tail `kplusplusSync` uses (module name, `INCLUDE_MISSING`, `krapper.featuregen`, `--header`), into `featuregen/build/krapped-cpp` (NOT the committed `krapped/`).
- **plugin** (`compiler/gradle`): `krappedDirFor()` redirects the cinterop `.def` + generated sources to `build/krapped-cpp` under the property, and `kplusplusSync` skips its libclang body for that module (the external task supplies the dir). Both default-absent → standard path byte-for-byte unchanged.
- **featuregen**: under the property, wires `kplusplusSync` → `:cppfrontend:featuregenCppBindings`, and excludes the one test unit blocked by a dropped binding (below) so the rest can be measured.

### THE RESULT
- **Generation GREEN** — cpp front-end emits featuregen's full surface: **116 Kotlin files, all 17 instantiations**, wrapper `-c` compiles.
- **186/188 test sources COMPILE-clean against the cpp bindings** (`:featuregen:compileTestKotlinNative -PenableClang -Pkpp.frontend.featuregen=cpp`, GREEN). This is the **spelling-safety half** of flip-safety: no spelling delta in the ~345-line ALL diff renames a symbol any of those 186 callers use. (Most of the ALL diff is EXTRA / more-correct cpp bindings — `root___locale_struct`, `std___exception_ptr_Exception_ptr`, `std_Type_info` — which don't break callers.)

### Flip worklist (the valuable signal — categorized)
1. **`std::pair<int,int>` DROPPED under the full multi-instantiation cpp sync** → breaks `CpPairTest` (2 tests; `import std.Pair__Int__Int` + `first`/`second` unresolved). It generates fine **per-spec** (featuregenParity ledger `std::pair<int, int>=diff:83`), so this is a **multi-instantiation forcing cross-contamination drop** — *binding-side, NOT a spelling rename, NOT test-side*. (Other libclang-only drops under the full run — `Cpu_set_t`, `_pthread_cleanup_buffer`, `Sched_param`, `Timespec`, `_Bit_reference` — are incidental system types no test references; the compile confirms it.) `CpPairTest` is excluded from the cpp-harness compile (gated only) so the remaining 186 can be measured.
2. **Toolchain libstdc++ skew blocks the behavioral RUN** → `:featuregen:nativeTest -PenableClang -Pkpp.frontend.featuregen=cpp` reaches `linkDebugTestNative` and fails. The cpp front-end parses against the **system** libstdc++ (gcc 16), so its faithful model surfaces newer-libstdc++ internals (`std::__size_to_integer`, `__throw_bad_array_new_length`, `__glibcxx_assert_fail`) that K/N's bundled gcc-8.3 lld can't resolve. *Toolchain-side, not binding correctness* — aligning the cpp parse sysroot with the konan toolchain K/N links against is a separate flip prerequisite (front-end change, deliberately out of scope for this additive brick).

Because the runtime link is blocked by (2), I could not produce a runtime pass-count — so this lands the **generation + compile-resolution** half of flip-safety (honest smaller-scope per the brick's option 3), with the two prerequisites above as the concrete, categorized flip worklist.

### Verify
- Default unaffected: `./gradlew :featuregen:kplusplusSync :featuregen:nativeTest` (no flag) — GREEN, krapped/ ZERO drift.
- Gated harness: `GRADLE_OPTS="-Xmx12g" ./gradlew :featuregen:compileTestKotlinNative -PenableClang -Pkpp.frontend.featuregen=cpp` — GREEN (186/188 resolve); `:featuregen:nativeTest` with the same flags reaches link and fails on the documented toolchain skew.


monkopedia-reviewer (approved):
Tier 1 — approve and merge.

**Default-path purity (load-bearing check): CONFIRMED.** Every plugin hunk is strictly gated:
- `krappedDirFor()` returns `File(projectDir, "krapped")` by default; only redirects to `build/krapped-cpp` when `kpp.frontend.${target.name} == "cpp"`.
- The `kplusplusSync doLast` skip fires only under the same property condition.
- The featuregen wiring block is inside `if (providers.gradleProperty("kpp.frontend.featuregen").orNull == "cpp")`.
- `moduleName = target.name` (plugin line 66) — consistent with `krappedDirFor` property key.

**Gating completeness:** `:cppfrontend` is only included under `-PenableClang` (settings.gradle.kts); the featuregen→cppfrontend dep only wires under both flags. The `featuregenCppBindings` task writes only to `featuregen/build/krapped-cpp` (under `**/build` in root .gitignore) — never to committed `krapped/`.

**Honest reporting:** PR correctly scopes to generation + compile-resolution (186/188); the toolchain libstdc++ skew blocking the behavioral RUN is named with root cause; CpPairTest exclusion is documented as worklist item 1 (not silently dropped); worklist filed on #47.

**Build verification (local — no CI on PR branches):**
1. Default purity: `./gradlew :featuregen:kplusplusSync :featuregen:nativeTest` (no flags) → BUILD SUCCESSFUL 188/0, zero krapped/ drift. ✓
2. Plugin compile JDK 21: `:compiler:kplusplus-compiler-plugin:compileKotlin` → BUILD SUCCESSFUL. ✓  
3. Gated generation: `:cppfrontend:featuregenCppBindings -PenableClang -Pkpp.frontend.featuregen=cpp` → BUILD SUCCESSFUL (116 Kotlin files, 17 instantiations). ✓
4. Gated compile: `:featuregen:compileTestKotlinNative -PenableClang -Pkpp.frontend.featuregen=cpp` → BUILD SUCCESSFUL (186/188 with CpPairTest excluded per claim). ✓

Self-reduce (3rd commit: drop unused `kpp.krappedDir` escape hatch) is clean and correct.

Issue #78, opened 2026-06-12
[clang-campaign] cpp front-end parses against system libstdc++ (gcc-16); K/N bundled gcc-8.3 lld can't link the newer internals it faithfully surfaces


monkopedia-coder:
Found by the Phase E step-2 flip-safety harness (PR #76). The cpp front-end's faithful parse against the SYSTEM libstdc++ (gcc 16) surfaces newer libstdc++ internals (`std::__size_to_integer`, `__throw_bad_array_new_length`, `__glibcxx_assert_fail`) in the generated bindings. featuregen's `:nativeTest` then reaches `linkDebugTestNative` and FAILS — Kotlin/Native's bundled gcc-8.3 `lld`/sysroot can't resolve those symbols.

KEY INSIGHT: the libclang front-end's LOSSINESS was accidentally SHIELDING featuregen from this skew — it dropped those newer internals, so they never reached the link. The cpp front-end's correctness EXPOSES the long-standing parse-sysroot vs link-sysroot mismatch. This is toolchain-side, NOT a binding-correctness bug.

This is a genuine FLIP PREREQUISITE and a judgment call on approach — candidate fixes:
- Align the cpp parse sysroot with K/N's bundled toolchain (point buildASTWithArgs at a gcc-8.3-compatible sysroot / `-isystem`), so the cpp parse sees the same libstdc++ the link uses; OR
- a marshalling-boundary filter that drops/forwards the unlinkable newer internals (closer to what libclang did, but explicit); OR
- bump K/N's bundled toolchain / link sysroot to match system.
Needs a decision on which — recommend discussing before implementing. Blocks the behavioral half of flip-safety (the 186/188 compile-resolution half already passed).

PR #79, opened 2026-06-12
krapper_gen: a forced instantiation target may drop an unresolvable primary base — fixes std::pair<int,int> on the cpp front-end (#77)

### LINKED ISSUE #77: [clang-campaign] cpp front-end: std::pair<int,int> drops under the FULL multi-instantiation sync (per-unit OK) — breaks CpPairTest
Found by the Phase E step-2 flip-safety harness (PR #76). Under `--frontend=cpp` over featuregen's FULL 17-instantiation `kplusplusSync` (all specs forced together), the `std::pair<int,int>` binding is DROPPED, breaking CpPairTest's 2 tests. It generates fine PER-UNIT (parity ledger `std::pair<int, int>=diff:83`), so this is a multi-instantiation cross-contamination drop on the cpp forcing path — binding-side, NOT a spelling rename, NOT test-side. The libclang path keeps it under the full sync.

Likely in the cpp `--forcingModel` load + `requestInstantiation` merge (PR #63) when many specializations are loaded together: a per-payload identity/merge collision drops pair where it survives in isolation. Repro: the `:cppfrontend:featuregenCppBindings` task (PR #76) — pair absent from build/krapped-cpp under the full run. Flip-prerequisite (behavioral). Mirror the per-unit success into the full-sync path.

monkopedia-coder:
### Root cause (binding-side, NOT the merge / NOT multi-instantiation)
The cpp front-end models `std::pair` faithfully: newer libstdc++ declares
`struct pair : private __pair_base<_T1, _T2>`, and that private base is in the model
with **unsubstituted dependent params** (`__pair_base<_T1, _T2>`), so it can't resolve.
libclang's lossy AST walk omits `__pair_base` entirely, so its `pair` has no base.

When a forcing struct's `value` field (the requested specialization) is materialized
on-demand under `INCLUDE_MISSING`, the expansion uses `dropUnresolvablePrimaryBase = false`
(the legacy hard-fail that protects against resurrecting incidental transitively-referenced
subclasses). So pair's unresolvable `__pair_base` **hard-failed the whole pair class** and it
was dropped — `Warning: can't resolve forcing struct KrapperForce_std_pair_int_int`.

This is **not** a multi-instantiation cross-contamination: pair drops even when forced
ALONE on the cpp path. (The per-spec `featuregenParity` ledger reads `std::pair<int, int>=diff:83`
because the dropped-pair file-set gap was hidden as a `file sets differ` note inside that
unit's diff verdict, not a separate failure.)

### Fix
A forced instantiation target is an **explicitly requested binding**, so it deserves the
same listed/found-class treatment as a top-level allowlisted class: drop an unresolvable
PRIMARY base instead of hard-failing. Added `ResolveContext.forcingTargetKeys` (the request's
target type-strings); `requestInstantiation` seeds it with `setOf(target)`, and the
`INCLUDE_MISSING` typeMapper relaxes `dropUnresolvablePrimaryBase` to true **only** when the
class being expanded is exactly a forcing target. Everything else keeps the hard-fail.
`__pair_base` is an empty constraint base — dropping it loses nothing; pair keeps
`first`/`second`/`swap`, matching the libclang binding byte-for-byte.

No-op off the forcing path (the set is empty), and a no-op on the libclang forcing path
(its pair has no base), so `krapped/` is byte-identical.

### Before / after
- **Before:** `featuregen/build/krapped-cpp` had 116 files, **no `std_Pair__Int__Int.kt`**; CpPairTest excluded from the cpp harness.
- **After:** 117 files, `std_Pair__Int__Int.kt` present and **byte-identical** to the committed libclang `featuregen/krapped/src/std_Pair__Int__Int.kt`. CpPairTest re-included; `:featuregen:compileTestKotlinNative -PenableClang -Pkpp.frontend.featuregen=cpp` compiles all **188** tests (was 186/188). (The behavioral RUN is still blocked by the separate #78 toolchain skew — that gates the LINK, not the compile.)

### Verify
- `:krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` (no flag) — GREEN, **zero `krapped/` drift**.
- `:cppfrontend:goldenCompare :cppfrontend:featuregenParity -PenableClang` — GREEN, **18/18, 0 regressions** (the `std::pair<int, int>` unit's only-libclang file-set gap is now closed; the cpp surplus `root___locale_struct.kt` is the pre-existing D2b accepted residue).
- `:cppfrontend:featuregenCppBindings :featuregen:compileTestKotlinNative -PenableClang -Pkpp.frontend.featuregen=cpp` — GREEN (pair present, 188/188 compile).

### DIFF
--- a/featuregen/build.gradle.kts
+++ b/featuregen/build.gradle.kts
@@ -68,15 +68,9 @@ if (providers.gradleProperty("kpp.frontend.featuregen").orNull == "cpp") {
     tasks.named("kplusplusSync") {
         dependsOn(":cppfrontend:featuregenCppBindings")
     }
-    // FLIP-WORKLIST item (#47): CpPairTest references std::pair<int,int>'s binding, which
-    // the cpp front-end DROPS under the full multi-instantiation sync — even though it
-    // generates fine per-spec (featuregenParity: std::pair<int,int>=diff:83). So this is a
-    // multi-instantiation forcing cross-contamination drop (binding-side, not a spelling
-    // rename, not test-side), the ONE thing blocking a clean flip today. Exclude just this
-    // unit from the cpp-harness test compile so the remaining 186 behavioral tests can be
-    // measured against the cpp bindings. This filter applies ONLY under
-    // -Pkpp.frontend.featuregen=cpp; the default test path compiles ALL tests unchanged.
-    tasks.named("compileTestKotlinNative") {
-        (this as org.gradle.api.tasks.util.PatternFilterable).exclude("**/CpPairTest.kt")
-    }
+    // #77 RESOLVED: std::pair<int,int> now binds under the full multi-instantiation cpp
+    // sync (a forced instantiation target may drop an unresolvable primary base — pair's
+    // private __pair_base<_T1,_T2> the libclang walk omits — instead of hard-failing; see
+    // ResolveContext.forcingTargetKeys). So CpPairTest is no longer excluded: all 188
+    // behavioral tests compile against the cpp bindings.
 }
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -262,7 +262,13 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
             resolver,
             config.referencePolicy,
             alreadyBoundKeys,
-            forcedContainerKeys
+            forcedContainerKeys,
+            // The explicitly-requested specialization (this request's target) is a listed
+            // binding, so its INCLUDE_MISSING materialization may drop an unresolvable
+            // PRIMARY base instead of hard-failing — e.g. std::pair<int, int>, whose
+            // faithful cpp model carries the private __pair_base<_T1, _T2> the libclang
+            // walk omits. See ResolveContext.forcingTargetKeys.
+            forcingTargets = setOf(target)
         )
         // One entry per class type-key, last copy wins (#60): without this, every
         // writer received a fresh full copy of each bound class per instantiation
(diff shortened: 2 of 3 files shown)


monkopedia-reviewer (approved):
**Tier 1 — approve + merge.**

**Category:** `bug_fix` — targeted fix for `std::pair<int,int>` dropped on the cpp front-end; no API surface change. Tier 1 per kplusplus owner override (\* : tier 1).

---

**Shared-path purity — VERIFIED CLEAN (the hard scrutiny item).**

The relaxation is scoped to exactly the forcing target:
- `forcingTargetKeys = setOf(target)` seeds one type-string key — the single explicitly-requested specialization (e.g. `"std::pair<int, int>"`).
- The INCLUDE_MISSING typeMapper checks `it.toString() in context.forcingTargetKeys` — an exact match only when the class *being expanded* is the requested target, not any of its bases, members, or transitive references.
- Non-target classes retain `dropBase = false` → legacy hard-fail preserved. No sibling/transitive leakage: the set propagates via `context.copy()` into recursive resolution but only the literal target string matches.
- The legacy guard's purpose (preventing resurrection of incidentally-pulled subclasses with incomplete-at-emit bases) is intact for every non-target INCLUDE_MISSING expansion.

**Zero libclang drift — VERIFIED.** `featuregen/krapped/` has zero diff after sync. libclang's pair has no `__pair_base` in its model, so the `forcingTargetKeys` relaxation is a complete no-op on the libclang path.

**Pair byte-identical — VERIFIED.** `diff featuregen/build/krapped-cpp/src/std_Pair__Int__Int.kt featuregen/krapped/src/std_Pair__Int__Int.kt` → empty.

**CpPairTest re-included — VERIFIED.** FLIP-WORKLIST exclusion removed; 188 compile confirmed.

---

**Build results (all green):**
1. `:krapper_gen:nativeTest :featuregen:kplusplusSync :featuregen:nativeTest :krapper_gen:ktlintCheck` — BUILD SUCCESSFUL, zero `krapped/` drift.
2. `:cppfrontend:goldenCompare :cppfrontend:featuregenParity -PenableClang` — BUILD SUCCESSFUL, 18/18 [as expected], 0 regressions. `std::pair<int,int>` unit no longer shows a file-set gap.
3. `:cppfrontend:featuregenCppBindings :featuregen:compileTestKotlinNative -PenableClang -Pkpp.frontend.featuregen=cpp` — BUILD SUCCESSFUL, 117 files (pair present), 188/188 compile.

**Noted gap (non-blocking):** No dedicated `nativeTest` asserting "forcing target with unresolvable primary base survives." Coverage is via integration (parity + byte-identical pair + 188 compile). A follow-up unit test would lock this at the unit level.

Code quality: idiomatic, comments accurate and thorough, minimal abstraction, no dead code.

PR #80, opened 2026-06-13
fix(krapper_gen): holder-factory CALL must use canonical name, not per-file import alias (#75)

### LINKED ISSUE #75: [codegen] constructorMethod uses plainName for a cross-file _Holder call — sibling of #73 (class ARG_CAST + import collision)
Found during review of PR #74 (#73 fix), confirmed PRE-EXISTING + non-blocking. `constructorMethod` (KotlinWriter.kt ~2542-2545) builds a call to `<Type>_Holder` using `type.plainName`. If the calling file aliases that type's name via an import collision (the same per-file remap #73 fixed for DECLARATIONS), the emitted call becomes `<AliasedName>_Holder` — an undeclared symbol — because the `_Holder` factory is declared under the canonical name.

Scope: triggers ONLY for a CLASS (not enum) ARG_CAST return whose type is reached through an import collision in the calling file. No enum scenario hits it (so #73's clang.attr.Kind case is unaffected), and featuregen's specs don't trigger it (byte-identical holds). #73's fix corrected the DECLARATION side; this is the remaining REFERENCE-side case where a Holder-factory call should use the same per-file alias the import block establishes (or the canonical name, matching how the factory is declared) — get the declaration-vs-reference distinction right here too.

Likely a small fix mirroring #73: ensure the `_Holder` call site resolves the name through the same mechanism that declares the factory. A krapper_gen nativeTest reproducing a class ARG_CAST return under an import collision would lock it.

monkopedia-coder:
Fixes #75 — the REFERENCE-side sibling of #73 (declaration side, PR #74).

## Confirmed mechanism + trigger
A class with an `ARG_CAST` return is constructed via a generated `<Type>_Holder` factory.

- **Declaration** (`KotlinWriter.kt` ~750): the `fun MemScope.<Type>_Holder()` factory is declared once, in the class's *own* file. That file resets `remap` to its own import map and never aliases its own type, so the declared name is canonical (`Kind_Holder`).
- **Call** (`KotlinWriter.generateMethodBody` ~2084, the `returnStyle == ARG_CAST && kotlinType.isWrapper` path): the call is emitted in the file of the **method's owning class**, a *different* file. If that file has a same-file import collision and aliases the returned type (`clang.attr.Kind` -> `AttrKind`), reading `kotlinType.plainName` there returns the alias, so the emitted call became `AttrKind_Holder` — an **undeclared symbol** (the factory is `Kind_Holder`). K/N compile failed.

This is the inverse distinction from #73: there, declarations went canonical while imported *type* references legitimately kept the alias. Here the `_Holder` symbol is a generated **sibling factory name**, not an imported type reference — so the CALL must also be canonical, matching its canonical declaration.

## Fix
Both the `_Holder` **declaration** (~750) and its **call** (~2084) now use `plainDeclaredName + "_Holder"` (the canonical accessor added in #73, which ignores `remap`). Declaration-name == call-name by construction, independent of any calling file's alias. The receiver qualifier (`fullyQualified + ".Companion"`) was already canonical. Added comments stating the declaration-name-matches-call-name invariant at both sites.

The declaration change is byte-identical (it was already canonical in its own file); the call change is the actual fix.

## Test
`HolderCallRemapTest` builds the real holder-call symbol exactly as `generateMethodBody` does (via `extensionMethod` + `buildCode`), applies a reference-file remap (`clang.attr.Kind` -> `AttrKind`), and asserts the rendered call stays `Kind_Holder`. It also asserts the pre-fix `plainName` construction would have rendered `AttrKind_Holder`, pinning the exact regression. Mirrors #73's `DeclNameRemapTest`.

## Verify (JDK 17, foreground)
- `:krapper_gen:nativeTest` green — HolderCallRemapTest 1/0, suite green.
- `:krapper_model:build`, `:featuregen:kplusplusSync`, `:featuregen:nativeTest`, `:krapper_gen:ktlintCheck` all green.
- **Byte-identical**: zero `featuregen/krapped/` drift after sync (specs don't hit the collision).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -747,7 +747,14 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
                 cls.children.filterIsInstance<ResolvedDestructor>().firstOrNull()
             extensionFunction {
                 receiver = fqType(MEM_SCOPE)
-                name = cls.type.kotlinType.plainName + "_Holder"
+                // The `<Type>_Holder` factory is a generated SIBLING function, not an
+                // imported type reference: it is DECLARED here under the canonical name
+                // (`plainDeclaredName`, ignoring `remap`) and every CALL to it (see
+                // generateMethodBody's ARG_CAST path) must use that SAME canonical name.
+                // `plainName` would route through the per-file import alias (issue #75) —
+                // a remote calling file that aliased this type would then emit
+                // `<Alias>_Holder`, an undeclared symbol. Declaration-name == call-name.
+                name = cls.type.kotlinType.plainDeclaredName + "_Holder"
                 retType = type(cls.type)
                 body {
                     if (defaultConstructor != null) {
@@ -2081,7 +2088,13 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
                 initializer = memScope dot Call(
                     extensionMethod(
                         kotlinType.fullyQualified + ".Companion",
-                        kotlinType.plainName + "_Holder"
+                        // CALL the `<Type>_Holder` factory by its canonical DECLARED name
+                        // (`plainDeclaredName`), matching the declaration site above. This
+                        // body is emitted in the file of the method's owning class, which
+                        // may have aliased `kotlinType` via an import collision; `plainName`
+                        // would yield `<Alias>_Holder` (undeclared), since the factory is
+                        // declared canonical regardless of the calling file's alias (#75).
+                        kotlinType.plainDeclaredName + "_Holder"
                     )
                 )
             ).also {
--- a/krapper_gen/src/nativeTest/kotlin/HolderCallRemapTest.kt
+++ b/krapper_gen/src/nativeTest/kotlin/HolderCallRemapTest.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.krapper.generator
+
+import com.monkopedia.krapper.generator.builders.buildCode
+import com.monkopedia.krapper.generator.builders.extensionMethod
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
## Review: tier-1 auto-approve

**Category:** bug_fix — targeted fix for ARG_CAST _Holder factory call emitting an aliased (undeclared) name.
**Tier:** 1 (kplusplus all-tier-1 override; bug_fix is default tier 1).

### Correctness (axis A) ✓

**Canonical-call fix is correct and scoped.**
- Declaration site (~750): `cls.type.kotlinType.plainDeclaredName + "_Holder"` — byte-identical (the declaring file never aliases its own type, so was already canonical; comment now makes the invariant explicit).
- Call site (~2088): `kotlinType.plainDeclaredName + "_Holder"` — the real fix. The generating-method body is emitted in the **method owner's** file, which may have a same-file import collision that aliases the returned type (`clang.attr.Kind` → `AttrKind`). Pre-fix, `plainName` routed through that alias → `AttrKind_Holder` (undeclared, K/N compile failure). Post-fix, canonical → `Kind_Holder` (matches the declaration).
- Receiver (`fullyQualified + ".Companion"`): `fullyQualified` never routes through `mappedName`/`remap`, was already canonical. No change needed; confirmed.

**No over-reach.** The diff changes only the two `_Holder` token sites. Every other `plainName` call in KotlinWriter.kt (lines 336, 349, 374, 682, 1182, 1668, 1805, 2235) is legitimately a type *reference* (should follow the alias) and is left untouched. Arguments and casts in the ARG_CAST path are also unaffected.

**The invariant is clean:** a generated *sibling factory name* is canonical at both ends; only imported *type references* follow the per-file alias. This is the correct distinction from #73 (where the alias was right for type references but wrong for the declaration).

### Test (axis A) ✓

`HolderCallRemapTest.holderCallStaysCanonicalUnderAReferenceFileAlias` passes 1/0/0. It:
1. Builds the holder-call symbol exactly as `generateMethodBody` does (`extensionMethod` + `buildCode`).
2. Stamps a remap alias (`clang.attr.Kind` → `AttrKind`) via `setNameRemap`, matching `ImportBlock`'s mutation path.
3. Asserts `plainDeclaredName` path → `Kind_Holder` (canonical; the fix).
4. Asserts `plainName` path → `AttrKind_Holder` (regression; pinned).

Non-tautological, exercises the failure path directly, and would fail on the pre-fix code. Mirrors `DeclNameRemapTest` structure from #73.

### Build verification ✓

Full local gate (no remote CI configured on this repo):
- `:krapper_gen:nativeTest` — BUILD SUCCESSFUL, HolderCallRemapTest 1/0/0, suite 316/0.
- `:krapper_model:build` — UP-TO-DATE, green.
- `:featuregen:kplusplusSync` + `:featuregen:nativeTest` — 73 suites / 0 failures; zero `krapped/` drift (byte-identical as claimed).
- `:krapper_gen:ktlintCheck` — clean.

### Quality (axis B) ✓

Comments at both changed sites clearly explain the invariant and cite the issue number. Comment density and style match the surrounding code. Test structure mirrors the sibling `DeclNameRemapTest`. No dead code, no over-abstraction.

### Follow-up (noted, not blocking)

`constructorMethod` (~line 2555) uses `type.plainName` for the CONSTRUCTOR factory call — same hazard class (declaration canonical, call site in a cross-file context may have an alias active). Confirmed it's a distinct path (NATIVE_CAST / isWrapper return, not ARG_CAST) and correctly out of scope for #75. A follow-up issue should be filed.

---
Approving and squash-merging.

PR #82, opened 2026-06-13
featuregen: link the gated cpp-front-end test variant against system libstdc++ (#78)

monkopedia-coder:
## Fixes #78 — the last flip prerequisite

Completes the behavioral flip-safety proof: featuregen's **188 tests now RUN green** when built on cpp-front-end bindings.

### Diagnosis (confirmed)
PR #76's harness (`-Pkpp.frontend.featuregen=cpp`) regenerates featuregen's bindings via `--frontend=cpp`; they COMPILE (188/188 sources after #77/#79), but `:featuregen:nativeTest` failed at `linkDebugTestNative`. The cpp front-end parses against the **system libstdc++** (gcc-16), so the faithful generated wrapper references newer-libstdc++ internals — confirmed undefined at link:
- `std::__throw_bad_array_new_length()@GLIBCXX_3.4.29`
- `std::__glibcxx_assert_fail()@GLIBCXX_3.4.30`

K/N's default test link resolves libstdc++ from its bundled gcc-8.3 sysroot (caps at GLIBCXX_3.4.25) → undefined.

### The fix (Approach 1 — linker config, under the existing property guard)
On the cpp-variant test binary's link only:
- **Link the REAL system `libstdc++.so` by ABSOLUTE PATH**, discovered at configure time from `clang++ -print-file-name=libstdc++.so` (canonicalized to the real shared object; host-portable, not hardcoded). Absolute path is required because `-L/-lstdc++` is shadowed by konan's bundled-sysroot `-L` (searched first → old lib → symbols stay undefined).
- **`--gc-sections`** to dead-strip the stale `platform.linux` cinterop cache (the same trick the klinked clang consumers use — `clangwalk/build.gradle.kts`).
- **`--allow-shlib-undefined`** — the system `libstdc++.so.6` in turn references newer glibc symbols (`pthread_*@GLIBC_2.34` from the libpthread→libc merge, `strfromf128@2.26`, `__isoc23_*@2.38`) the old sysroot glibc stub lacks; these resolve at **runtime** against the host glibc. This relaxes only refs coming from **shared libraries** — the generated `libfeaturegen.a` archive's own symbols stay fully checked, so a genuinely missing binding symbol would still fail the link.
- **`-rpath`/`-rpath-link`** so the system `.so` (SONAME `libstdc++.so.6`) loads at runtime.

No parse-sysroot downgrade — the bindings stay maximally correct; this is purely the linker route.

### Result
[...]

PR #84, opened 2026-06-13
flip(B2): generic per-module cpp-front-end binding generation (#47)

monkopedia-coder:
## Flip brick B2 (#47, Phase E step 3) — generic per-module cpp-front-end binding generation

Generalizes the featuregen-**hardcoded** cpp binding path into a **generic per-module plugin path**, so ANY module can build its bindings via the cpp front-end from its own config. Prerequisite for B3 (flipping the default).

### The generic mechanism (how a module's own config drives cpp gen)
Under `-Pkpp.frontend.<module>=cpp`, the kplusplus gradle plugin's `kplusplusSync` (`compiler/gradle/.../KPlusPlusCompilerGradlePlugin.kt`) now **generates** the bindings itself instead of skipping and waiting on an external hardcoded task:

1. resolves the module's own `kplusplus { header(...) ; instantiate(...) }` config (the extension fields are `internal` — only the plugin can read them, which is why the orchestration lives here);
2. builds the worklist = the compiler-written manifest (`<projectDir>/krapped/requested.txt`) ∪ the seeded `instantiate()` calls, deduped + sorted — the **same** union the libclang path uses;
3. runs the gated `:cppfrontend` binary `--parity-emit` over **that module's** header (base model) + each forced instantiation (forcing models) → `:krapper_model` ModelIo JSON (one invocation per payload, so a parse crash on one spec is isolated + named);
4. invokes `krapper_gen --frontend=cpp --parsedModel/--forcingModel` with the **same CLI the libclang sync builds** (module name, reference policy, package, `--instantiate`/`--only`/`--header`/fixups/`--std`/`--root-package`), minus the libclang parse, into `build/krapped-cpp`;
5. auto-wires `:cppfrontend:linkReleaseExecutableKlinker` as a dependency (gated, only when `:cppfrontend` is in the build).

The featuregen-specific `:cppfrontend:featuregenCppBindings` task (header + seeds baked in) is **replaced** by this generic path; `featuregen/build.gradle.kts` loses its manual `dependsOn(":cppfrontend:featuregenCppBindings")` wiring (the plugin self-wires). `featuregenParity` keeps the diff/regression ledger.

### featuregen still green THROUGH the generic path
`:featuregen:nativeTest -PenableClang -Pkpp.frontend.featuregen=cpp` → **188/0/0**, with `build/krapped-cpp/cpp-models/` present (the new plugin-path marker; the removed task wrote elsewhere) and 117 generated Kotlin files. The system-libstdc++ run link from #82 is retained (under the property guard); only the generation wiring moved to the plugin.

### Generality demo — which modules are cpp-ready (B3 input)
New `:cppfixture` module: a minimal **std-free** single-header consumer (`point2d.h`) with its own `kplusplus { header(...) }`. `:cppfixture:nativeTest -PenableClang -Pkpp.frontend.cppfixture=cpp` → **2/0**, generating `root_Point2D.kt` from the module's own header (0 instantiations, base model only) and compiling + running against it. Proves the generic path drives a **different** module's own config.

cpp-readiness today:
- **cpp-ready:** `featuregen` (std header, full 17-spec worklist), `cppfixture` (std-free).
- **NOT yet cpp-ready:** `example` (v8), `cppfrontend`/`clangwalk` (clang/LLVM slice) — heavy headers cppfrontend can't parse cleanly yet; `slice` has no header (pure-instantiation). These are B3 inputs.
- Limitation: the cpp path parses ONE primary header (forcing headers `#include` it); multi-header cpp parsing is a later extension. Both cpp-ready modules are single-header.

### Fail-fast assertion (de-risks B3)
After generation, if `krapped-cpp` holds no ACTUAL binding source (only the always-emitted `CppBinding.kt` boilerplate, or no `.def`) — i.e. cppfrontend parsed the header into zero classes — `kplusplusSync` throws a clear message **naming the module** rather than silently compiling against nothing. Verified live: pointing `:cppfixture` at a comment-only header trips it with `EMPTY binding set ... binding kotlin files=0 ... refusing to compile cppfixture against nothing`.

### REVERSIBLE — nothing default changed, nothing deleted
- cpp stays **opt-in** via `-Pkpp.frontend.<module>=cpp`; B3 flips the default, not this PR.
- The libclang front-end (krapper_gen + the plugin's libclang body) is **untouched**.
- `:cppfixture` is included **only** under the LLVM gate (`-PenableClang`) in `settings.gradle.kts`, so the default `./gradlew` never sees it.
- **Default proof:** `:featuregen:kplusplusSync :featuregen:nativeTest` (no flags) → **188/0**, zero `krapped/` drift; plugin compiles under JDK 21.

Refs #47 (flip brick B2).


monkopedia-reviewer (changes requested):
## Review — flip(B2): generic per-module cpp-front-end binding generation

**Tier:** 1 (kplusplus override — every PR auto-merges on green review)
**Category:** feature_internal — generic plugin path, fully opt-in/gated
**Verdict:** request_changes (one quality issue)

---

### Build results (all four gates)

1. **Default purity** — `:featuregen:kplusplusSync :featuregen:nativeTest` (no flags, JDK 17) → **188/0**, `featuregen/krapped/` zero drift (clean `git status`). ✅
2. **Plugin compiles JDK 21** — `:compiler:kplusplus-compiler-plugin:compileKotlin` → green. ✅
3. **featuregen via generic cpp path** — `:featuregen:nativeTest -PenableClang -Pkpp.frontend.featuregen=cpp` → **188/0**. ✅
4. **cppfixture** — `:cppfixture:nativeTest -PenableClang -Pkpp.frontend.cppfixture=cpp` → **2/0** (`Point2DTest`). ✅

---

### What was reviewed (sound)

- **Default-path purity** (hunk-by-hunk): the new `dependsOn(":cppfrontend:linkReleaseExecutableKlinker")` block at configuration time is double-gated by both the property check AND `findProject(":cppfrontend") != null`; absent the property it never executes. The `doLast` cpp branch (`if (target.findProperty("kpp.frontend.$moduleName") == "cpp")`) is gated identically. When the property is absent, the kexe-not-found check and the full libclang body execute byte-for-byte unchanged. ✅
- **Worklist correctness**: `runCppFrontendSync` reads `<projectDir>/krapped/requested.txt` (the same path the FIR checker hardcodes in `applyToCompilation`) + seeded `instantiations`, deduped+sorted — identical union to the libclang sync. ✅
- **Removed hardcoded task**: `featuregenCppBindings` and featuregen's manual `dependsOn` are both gone; featuregen now generates through the generic path and verified green (188/0). ✅
- **cppfixture gating**: included only inside the `if (clangEnabled)` block in `settings.gradle.kts`; default `./gradlew` never sees it. ✅
- **Fail-fast**: checks for ≥1 kt file other than `CppBinding.kt` (the always-emitted boilerplate) AND the `.def` — correctly ignores the boilerplate and fires on a genuinely empty binding set. ✅
- **`internal` extension-field access**: plugin reads `KPlusPlusExtension` via `target.extensions.findByType(...)`, the same mechanism already used in the libclang path — no visibility hack. ✅
- **Reversibility**: libclang path entirely untouched; cpp stays opt-in via the property; no default changed. ✅

---

### Blocking quality issue — stale `krappedDirFor` docstring

**File:** `compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt`
**Function:** `krappedDirFor` (the doc comment, ~line 344)

The current docstring says:

```
* `-Pkpp.frontend.<module>=cpp` consumes cpp-front-end bindings under
* `<buildDir>/krapped-cpp` instead, with kplusplusSync's libclang body skipped (an
* external task — :cppfrontend:featuregenCppBindings — supplies that dir).
```

`:cppfrontend:featuregenCppBindings` is **deleted** in this PR. After merging, the docstring describes a mechanism that no longer exists. `kplusplusSync` now generates the bindings itself via `runCppFrontendSync`. The featuregen `build.gradle.kts` comment was correctly updated; this one was missed.

**Fix:** update those two lines to say `kplusplusSync` generates via `runCppFrontendSync`, e.g.:

```
* `<buildDir>/krapped-cpp` instead; kplusplusSync GENERATES the bindings
* itself via runCppFrontendSync (cppfrontend --parity-emit → ModelIo JSON →
* krapper_gen --frontend=cpp), with the libclang body skipped.
```

Everything else is sound. Fix the docstring and this merges.

monkopedia-reviewer (approved):
Re-review after changes-request cycle. The single blocking item from my prior review — the stale `krappedDirFor` docstring referencing the deleted `:cppfrontend:featuregenCppBindings` task as a current supplier — is fully addressed by commit e43c1fb. The delta is docstring-only (one file, one hunk): the new text accurately states that `kplusplusSync` generates `<buildDir>/krapped-cpp` itself via `runCppFrontendSync` (the :cppfrontend binary over the module's own config → krapper_gen --frontend=cpp, generically, no per-module task). No other stale present-tense references remain; the two surviving `featuregenCppBindings` mentions are correctly past-tense ("replaces the old ... wiring" / "generalization of the former ... task"). Since e43c1fb is comment-only, the prior four green build gates (default purity 188/0, plugin JDK21 compile, featuregen cpp via generic path 188/0, cppfixture 2/0) remain valid — no rebuild required. Approving. ✅

Issue #86, opened 2026-06-14
[determinism] forcing-header synthesis emits the consumer #include by ABSOLUTE path (non-reproducible across checkout locations)


monkopedia-coder:
Found during flip B1 review (PR #85). When krapper_gen forces a template instantiation, the synthesized `KrapperForce_*.h` emits the consumer header as an ABSOLUTE include, e.g. `#include "/home/jmonk/git/kplusplus/.claude/worktrees/wt-b1/cppfrontend/include/clang_slice.h"`. So generated forcing headers are NOT byte-reproducible across checkout locations / machines (the path embeds the absolute project dir) — a real determinism hole (cf. #11/#25). It's invisible normally because every module's `krapped/` is gitignored+regenerated, but it surfaced the moment B1 committed cppfrontend's seed (the absolute path broke portability; hand-relativized in #85).

Fix: `ForcingHeader` (in :krapper_model) should emit the consumer #include as a path RELATIVE to the krapped output dir (matching what the main `.cc` already does — `../include/<header>`), or otherwise stable/relocatable. Must stay correct across module layouts (the relative path differs per module — compute it from the krapped dir to the header). Verify featuregen still builds + 188/0 + the in-process DeterminismTest, and that a re-cut cppfrontend seed then byte-matches a fresh libclang regen. Low urgency (the B1 seed is hand-fixed + works) but it's the proper determinism fix and a prerequisite for cleanly RE-CUTTING seeds (e.g. on an LLVM bump).

PR #87, opened 2026-06-14
flip(B3): default consumers off the libclang front-end (#47)

monkopedia-coder:
## flip brick B3 — default every consumer off the libclang front-end (#47, Phase E step 3, Path A)

Sets **persistent per-module front-end defaults** (root `gradle.properties`, key `kpp.frontend.<module>`) so **no module's binding generation uses the in-tree libclang front-end by default**. This clears the way for **B5** to delete libclang cleanly. Reversible: nothing deleted, libclang stays present + selectable; revert a flip by deleting/emptying its line (`-Pkpp.frontend.<m>=` overrides back to libclang).

### Per-module disposition

| Module | New default | Mechanism | Verify (its new default) |
|---|---|---|---|
| featuregen | **cpp** | `--frontend=cpp` via gated `:cppfrontend` | `:featuregen:nativeTest -PenableClang` → **188/0** |
| cppfixture | **cpp** | `--frontend=cpp` via gated `:cppfrontend` | `:cppfixture:nativeTest -PenableClang` → green |
| cppfrontend | **seed** | committed stage-0 seed (B1) | `:cppfrontend:linkReleaseExecutableKlinker -PenableClang` → green |
| clangwalk | **seed** | committed stage-0 seed (new here) | `:clangwalk:linkReleaseExecutableKlinker -PenableClang` → green |
| slice | **seed** | committed stage-0 seed (new here) | `:slice:linkReleaseExecutableKlinker` → green |
| example/kotlin_project | **stays libclang** | — | **B5 BLOCKER** (see below) |

### LLVM-mandatory default (intended end-state)
`cpp` mode generates through the gated `:cppfrontend`, which is only configured under `-PenableClang`. So featuregen + cppfixture's **default** build is now **LLVM-mandatory** (requires `-PenableClang` / a real LLVM toolchain). This is the deliberate crossover end-state. Confirmed: the featuregen **#82 system-libstdc++ link** block (`if (providers.gradleProperty("kpp.frontend.featuregen") == "cpp")`) now fires in the **default** cpp build (the gradle.properties value satisfies it, no `-P` needed) — the 188-test link resolves the newer-GLIBCXX symbols against the system libstdc++, which it could not under K/N's bundled gcc-8.3.

### cppfrontend seed completeness
Verified the existing B1 cppfrontend seed is **complete**: a fresh libclang regen diffs **byte-identical modulo the #86 forcing-header absolute-path** (already hand-relativized in the commit). The stray `KrapperForce_std_vector_clang_CXXMethodDeclPtr.h` seen previously is **old cruft** — it is NOT part of the current 2-instantiation config (`std::vector<clang::Decl*>` + `std::vector<clang::CXXBaseSpecifier*>`); no missing file.

### New seeds (clangwalk, slice)
Generated once via the current libclang front-end and committed as **text** (`src/*.kt`, `*.cc`/`*.h`, `KrapperForce_*.h`, + slice `requested.txt`); the host-local `.a`/`.def`/`generated.txt` stay gitignored and are regenerated at build by `runSeedSync` (a plain C++ recompile, **no krapper_gen, no libclang**). clangwalk forcing-header includes hand-relativized to `../include/clang_slice.h` (#86). slice forcing headers are std-only (`<vector>`/`<map>`) — no absolute path. slice's seed was iterated to its **fixpoint**: it includes the second-pass `std::vector<std::vector<int>>` discovery that Main.kt requires.

### B5-readiness check
With the host-local products deleted, seed mode rebuilds purely from committed text: `kplusplusSync[cppfrontend]: stage-0 seed — 116 committed binding source(s); recompiling libcppfrontend.a (no krapper_gen, no libclang)` and the same for clangwalk (72) / slice (12). Re-link is **up-to-date** (byte-stable / deterministic regen).

### B5 BLOCKER — example/kotlin_project (v8)
example could **not** be seeded in this brick: its libclang generation **SIGABRTs on the v8 surface** —
```
> Task :kplusplusSync
kplusplusSync: headers=1 (no instantiations requested)
> krapper_gen failed: exit 134
```
It writes a 1 MB `kotlin_project.cc` + 351 KB `.h` but **zero Kotlin bindings** (`src/` empty) before aborting, so there is no usable seed to commit. example therefore **remains on libclang** and is filed as a B5 blocker on #47: B5 cannot delete libclang until example is either seeded (needs the v8 generation crash resolved) or migrated to cpp (needs multi-header v8 cpp-front-end support).


monkopedia-reviewer (approved):
**Tier 1 auto-review — APPROVED**

## Build verification

| Check | Result |
|---|---|
| No-flag build (`:featuregen:kplusplusSync`) | FAILS CLEARLY: `kplusplusSync[featuregen]: -Pkpp.frontend.featuregen=cpp but :cppfrontend is not in this build. The cpp front-end needs the LLVM-gated modules — build with -PenableClang (or -PllvmConfig=<path-to-llvm-config>).` — B2 fail-fast, excellent UX |
| `:featuregen:nativeTest -PenableClang` | **188/0 PASS** |
| `:slice:linkReleaseExecutableKlinker` (seed, no -PenableClang) | BUILD SUCCESSFUL |
| `:clangwalk:linkReleaseExecutableKlinker -PenableClang` (seed) | BUILD SUCCESSFUL |

## Correctness findings

**Absolute paths in committed seeds:** None. clangwalk forcing headers use `"../include/clang_slice.h"` (relative, per #86 fix); slice forcing headers use only `<vector>` / `<map>` (std).

**gradle.properties:** Only the 5 `kpp.frontend.*` defaults added; no heap overrides, no other unintended changes. Comments explain the mechanism, revert path, and intended end-state clearly.

**gitignore:** Both clangwalk and slice gitignores correctly protect committed text (`.cc/.h/KrapperForce_*.h/src/**`) and gitignore host-local products (`.a/.def/generated.txt`). Reviewed the negative-pattern chaining — correct.

**Example blocker:** Confirmed. `example/kotlin_project` has its own `settings.gradle.kts` and is NOT included in the root `settings.gradle.kts`. The B5 SIGABRT blocker is real but doesn't affect the main build or any CI. Correctly left on libclang and documented.

**Reversibility:** Confirmed — delete/empty a `kpp.frontend.<m>` line reverts to libclang; no front-end code removed in this brick.

## Non-blocking comment

The gitignore change from `/krapped/` to `/krapped/*` + `!/krapped/src/**` un-ignores previously-generated `.kt` files on any dev workspace that previously ran kplusplusSync for slice/clangwalk. On this machine those stale files appeared as untracked and caused a compile failure (unresolved C interop references) until cleaned with `git clean -fd slice/krapped/src/ clangwalk/krapped/src/`. On a fresh checkout this is a non-issue — only affects workspaces with prior libclang runs. Worth documenting in the B5 runbook or adding a `runSeedSync` guard that warns about unexpected `.kt` files in `src/`.

This is a non-blocker — does not affect CI or clean checkouts — but the dev machine will need the `git clean` on first checkout post-merge.

---

LGTM. Seeds complete + portable, no-flag fail-fast graceful, featuregen 188/0 on cpp default, both seed modules build from committed text. B5 BLOCKER correctly scoped and isolated.

PR #85, opened 2026-06-13
flip(B1): stage-0 seed anchor — laggards build from committed bindings post-libclang-delete (#47)

monkopedia-coder:
Phase E step 3, **flip brick B1, Path A** (owner-approved) of the self-hosting crossover (#47). Establishes the **stage-0 anchor**: the mechanism by which modules that are NOT yet cpp-parseable still produce their generated bindings AFTER the in-tree libclang reducer is deleted from krapper_gen (B5). **REVERSIBLE** — additive, no default flipped, nothing deleted.

## Anchor chosen: Option 1 (committed-seed bindings) — and why

Option 1 over Option 2 (mavenLocal stage-0 publish) because it is dramatically leaner: no maven-publish infra for a K/N kexe (non-standard), no stage-0 resolution path in the includeBuild plugin, no immutable-version juggling. The laggards' headers are PINNED (LLVM-22 clang slice, v8) and determinism is locked (#11/#25), so a committed seed is byte-stable — exactly the case Option 1 is built for. Cons (bindings frozen until the module becomes cpp-ready) are acceptable given the pinned headers.

## How it works

New sync mode `-Pkpp.frontend.<module>=seed` (parallel to B2's `=cpp`). In seed mode `kplusplusSync`:
- **drops its krapper_gen link-task dependency and kexe input** — krapper_gen is neither built nor run (post-B5 its libclang front-end is gone);
- builds the module's bindings from a **committed seed** under `<module>/krapped/`.

The committed seed is the **deterministic TEXT** output of a one-time libclang generation: the Kotlin bindings (`src/`), the C++ wrapper (`<module>.cc`/`<module>.h`), and the forcing headers (`KrapperForce_*.h`). `runSeedSync` regenerates the two **non-committed, host-local** build products from that text:
- the cinterop **`.def`** — deterministic from the module name + this checkout's krapped dir, so **no absolute path is ever committed**;
- the compiled wrapper object **`lib<module>.a`**, via the SAME plain C++ compile `CppCompiler` runs (`clang++ -std -c -fPIE -I<krapped> -DV8_COMPRESS_POINTERS <module>.cc`) — a compile with **no header parsing**, exactly the step that survives the B5 libclang delete.

This deliberately keeps the 2.9M host/LLVM-pinned `lib.a` **out of git** (it is rebuilt at sync) and commits only portable, deterministic text.

## B5 simulation — proof cppfrontend builds without in-tree libclang

[...]

PR #91, opened 2026-06-14
krapper_gen: emit forcing-header consumer #include relative to the krapped dir (#86)

### LINKED ISSUE #86: [determinism] forcing-header synthesis emits the consumer #include by ABSOLUTE path (non-reproducible across checkout locations)

monkopedia-coder:
## Fixes #86 — forcing-header consumer `#include` was an absolute path

`ForcingHeader` (`:krapper_model`) synthesizes a `KrapperForce_*.h` per template
instantiation: a `KrapperForce_*` struct whose `value` member IS the requested
specialization, `#include`-ing the consumer's own header so the element type
resolves. That consumer `#include` was emitted as the **absolute** path of the
consumer header (`request.headers`, which the gradle plugin passes as
`target.file(h).absolutePath`). So generated forcing headers were **not
byte-reproducible across checkout locations** — a determinism hole (cf. #11/#25).
It surfaced when flip B1 committed cppfrontend's seed and the absolute path broke
portability (hand-relativized in #85).

### The fix — per-module relative include
Defer the forcing-header *content* synthesis from `requestInstantiation` (where
the output dir is unknown) to `writeTo` (where it is), and relativize each
consumer `#include` against the forcing file's own location:

```kotlin
for ((forceName, target) in forcingHeaders) {
    val forcingFile = File(outputBase, "$forceName.h")
    val relHeaders = request.headers.map { File(it).relativeTo(forcingFile) }
    forcingFile.writeText(ForcingHeader.contentFor(target, relHeaders))
    syntheticHeaders.add(forcingFile.path)
}
```

This **mirrors the mechanism the main `.cc` already uses** for its consumer
includes (`CppWriter.onGenerate` → `File(header).relativeTo(cppFile)`). The
forcing header lands in `<krapped>/` and is compiled with `-I<krapped>` (the
`.def` `compilerOpts`), so a path relative to that dir resolves; and because it
carries no absolute host path, the relative form is computed *per module* from
the real `<module>/krapped/` → consumer-header offset:
- cppfrontend: `../include/clang_slice.h`
- featuregen: `../../src/cppMain/include/strings_feature.h`

`forcingHeaders` now stores `forceName → target spec` (not pre-baked content),
since the bytes can only be synthesized once the output dir is known.

### Portability proof
- **featuregen regen:** all 17 `KrapperForce_*.h` now carry the relative
  `../../src/cppMain/include/strings_feature.h`; `grep -rn /home/jmonk` over
  every regenerated `KrapperForce_*.h` is **empty**.
- **cppfrontend seed re-cut:** a fresh regen of cppfrontend's bindings via the
  cpp front-end path produces both committed forcing headers
  (`KrapperForce_std_vector_clang_DeclPtr.h`,
  `KrapperForce_std_vector_clang_CXXBaseSpecifierPtr.h`) **byte-identical** to the
  committed seed (`../include/clang_slice.h`). The generator now natively emits
  what #85 hand-relativized — **no seed update needed**.

### Verification (JDK 17)
- `:featuregen:nativeTest -PenableClang` → **188 / 0** (forcing headers now have
  relative includes and still compile).
- cppfrontend seed forcing headers regen byte-identically (diff = identical).
- `:cppfrontend:featuregenParity` is independently broken by flip B5 (#47): its
  libclang-baseline arm invokes krapper_gen with no `--parsedModel` and aborts
  (`krapper_gen requires --parsedModel`) because the libclang reducer was deleted.
  Pre-existing, unrelated to this change (which touches no libclang/parsedModel
  path).

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -67,9 +67,12 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
     private var classes: List<ResolvedElement> = emptyList()
     private val mappings = mutableListOf<MappingService>()
 
-    // Forcing headers recorded by requestInstantiation for the specializations the cpp
-    // front-end forced. Keyed forceName -> file contents. They are re-materialized into the
-    // output dir by writeTo, so the generated wrapper #includes them by a stable, relative,
+    // Forcing specializations recorded by requestInstantiation for the targets the cpp
+    // front-end forced. Keyed forceName -> the instantiation target spec. The header CONTENT
+    // is synthesized later, in writeTo: only then is the output dir known, and the consumer
+    // `#include` in the forcing header must be RELATIVE to that dir (so a regen from any
+    // checkout location produces byte-identical headers — #86). Materialized into the output
+    // dir by writeTo, so the generated wrapper #includes them by a stable, relative,
     // self-contained path (deterministic across runs).
     private val forcingHeaders = linkedMapOf<String, String>()
 
@@ -136,8 +139,9 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
         // bytes for its forcing parse, so both paths force the same specialization from
         // the same translation unit.
         val forceName = ForcingHeader.forceName(target)
-        val forcingContent = ForcingHeader.contentFor(target, request.headers)
-        forcingHeaders[forceName] = forcingContent
+        // Record only the target spec; writeTo synthesizes the bytes once it knows the output
+        // dir (the consumer #include must be relative to where the header lands — #86).
+        forcingHeaders[forceName] = target
         // The forcing-parse tree is produced by the cppfrontend binary (which synthesizes
         // the SAME forcing header and parses it with the Clang C++ AST); load it. The
         // forcing header content above is still recorded: writeTo re-materializes it so the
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
## Review: bug_fix — tier 1

### Correctness ✓

**Relativization math verified (custom `File.relativeTo`)**

The path-computation in `codegen/File.kt` is:
```kotlin
fun relativeTo(file: File): String {
    var maxCommon = pathSegments.zip(file.pathSegments).indexOfFirst { it.first != it.second }
    if (maxCommon < 0) return path
    return (List(file.pathSegments.size - maxCommon - 1) { ".." } +
            pathSegments.subList(maxCommon, pathSegments.size)).joinToString("/")
}
```
The `-1` in `file.pathSegments.size - maxCommon - 1` strips the filename component from the base path, so the resulting `..` count is calibrated for the C preprocessor, which resolves `#include "..."` relative to the *directory* of the including file — not the file itself. Traced for both cases:

- **featuregen**: `build/krapped-cpp/KrapperForce_std_vector_int.h` → consumer at `src/cppMain/include/strings_feature.h` → `../../src/cppMain/include/strings_feature.h` → resolves correctly from the `build/krapped-cpp/` directory ✓  
- **cppfrontend**: `krapped/KrapperForce_std_vector_clang_DeclPtr.h` → `include/clang_slice.h` → `../include/clang_slice.h` ✓ (matches committed seed from #85)

The mechanism mirrors `CppWriter.onGenerate`'s existing relativization exactly.

**Deferral (requestInstantiation → writeTo)**: sound. `forcingHeaders` storing `forceName → target spec` (not pre-baked content) is a minimal change. Which instantiations are forced is unchanged; only when the include-path bytes are baked moved (to `writeTo`, where the output dir is known).

### Build verification ✓

- `:featuregen:nativeTest -PenableClang` → **188 / 0** ✓
- Generated `featuregen/build/krapped-cpp/KrapperForce_*.h` all contain `../../src/cppMain/include/strings_feature.h`; `grep -rn "/home/jmonk" featuregen/build/krapped-cpp/KrapperForce_*.h` → **empty** ✓
- `cppfrontend/krapped/` committed seeds already carry `../include/clang_slice.h` (hand-relativized in #85; this generator now produces the same bytes natively) ✓
- `:cppfrontend:featuregenParity` 18-failure confirmed pre-existing and unrelated: root cause is `Uncaught Kotlin exception: kotlin.IllegalStateException: krapper_gen requires --parsedModel <path>` — the B5/#88 libclang-baseline arm no longer has a libclang reducer. Nothing in this PR touches that path. ✓

### Quality ✓

One-line diff in `requestInstantiation` (drop pre-baked content, store target spec) and a tidy synthesis loop in `writeTo` with a clear explanatory comment block. Comments accurately describe the "why" (output dir not known until writeTo). No dead code, no verbosity, idiomatic Kotlin (`map`, destructured `for`). Consistent with the surrounding file's style.

**Non-blocking note**: the committed seeds in `featuregen/krapped/*.h` still carry the old absolute paths (they predate this fix). They don't affect builds (the sync task regenerates them at build time), but a follow-up sync commit from any checkout will write the portable relative form and eliminate the stale absolute paths from the tree. Not a blocker for this PR.

Tier 1 — auto-merge.

PR #88, opened 2026-06-14
flip(B5): DELETE the libclang-C reducer — krapper_gen self-hosts on cppfrontend (#47)

monkopedia-coder:
## flip brick B5 — THE FLIP (Phase E step 3, Path A; #47)

The irreversible step: the in-tree **libclang-C reducer is DELETED** from krapper_gen. The shipping tool now self-hosts on its own generated libclang-cpp bindings via `:cppfrontend` (Clang C++ AST → ModelIo JSON → krapper_gen `--parsedModel`), with **NO** in-process libclang parse path.

### Deleted surface (files + line counts)
| File | Lines | What |
|---|---|---|
| `krapper_gen/clang.def` | 18 | the libclang-C cinterop def (+ `cinterops{}` wiring + cflags/ldflags in build.gradle.kts) |
| `ClangExtensions.kt` | 551 | the entire libclang-C accessor layer |
| `ModelFactories.kt` | 709 | cursor→`WrappedElement` construction + the USR-keyed parse memo |
| `TypeFactories.kt` | 439 | `createForType` + `*TypedefElement` libclang-C type reducer (cppfrontend's TypeBuilder mirror untouched) |
| `Utils.kt` (`CursorTreeInfo`) | ~30 | the libclang cursor dump (kept `Utils.printerrln`) |

Plus the libclang-coupled parts of the survivors:
- **App.kt** — the `CXCursor` visitor block, the `Frontend` enum + `--frontend` flag, the libclang-only oracle flags (`--dumpParsedModel`/`--roundTripModel`/`--dumpModels`). Hardwired to the single model-consume front-end; `--parsedModel` is now REQUIRED.
- **Parsing.kt** — the libclang `parseHeader` body + private `parseHeader` + the diagnostics layer (`ErrorDiagnostic`/`handleDiagnostics`/`collectErrorDiagnostics`/`topLevelDecl`) + `ResolverBuilderImpl` + the now-orphaned `generateIncludes`/`find`. `parseHeader()` now only loads the cpp ModelIo model; the resolve+codegen + ModelIo consume path is fully intact.
- **Resolver.kt** — the `ResolverBuilder` interface (the `CValue<CXType>` seam).
- **IndexedServiceImpl** — `createIndex`/`Arena`/`index`, the cursor→element memo reset, and the forcing temp-dir parse. `requestInstantiation` now requires a `--forcingModel` per `--instantiate` (no libclang fallback — errors loudly).

[...]

PR #93, opened 2026-06-14
test(krapper_gen): relocate skip-not-crash + determinism coverage to the self-hosted boundary (#89)

### LINKED ISSUE #89: [post-flip] Verify/relocate coverage of the 5 krapper_gen parse-tests deleted in B5 (#88)
The self-hosting flip (B5, #88) DELETED 5 krapper_gen nativeTests because they invoked the now-removed in-process libclang parse: `ParseTest` (17), `ReturnShapeTest` (15), `ForcingCharacterizationTest` (4), `SkipNotCrashTest` (9), `DeterminismTest` (1) — krapper nativeTest 313→272. The B5 reviewer judged their coverage "relocated cpp-side" (cppfrontend self-checks + featuregenParity + featuregen 188). This ticket is to VERIFY that, not assume it — and fill any real gap with explicit cpp-side regression tests:

- **SkipNotCrashTest** (unmodelable shapes drop+log, not crash): is the skip-not-crash behavior covered by a cppfrontend self-check / featuregenParity unit? Add an explicit cpp-side test if not (it guarded a core invariant).
- **ReturnShapeTest / ParseTest** (return-style + parse-shape decode): covered by cppfrontend's 100+ self-checks + the goldenCompare? Spot-check the specific shapes each asserted.
- **ForcingCharacterizationTest** (multi-instantiation forcing): covered by featuregenParity's forcing + the handoffInstDiff oracle? Confirm.
- **DeterminismTest** (the #11a hermeticity guard): it tested the cursor→element memo reset, which B5 DELETED — so its subject is gone. BUT determinism of the self-hosted output still matters: add a cpp-path determinism test (two cold cppfrontend→ModelIo→krapper_gen runs byte-identical, or featuregenParity twice-stable) to replace the guard at the new boundary.

Goal: no silent coverage hole from the flip. Each deleted test → confirmed-covered or a new cpp-side test filed.

monkopedia-coder:
Fixes #89.

The self-hosting flip (B5, #88) deleted 5 krapper_gen nativeTests because they drove their invariants through the in-process libclang parse it removed. This audits each against current coverage and re-establishes the two CORE invariants that had no explicit assertion left, on the surviving deserialize → resolve → codegen boundary (no parse front-end), built from hand-constructed / `ModelIo`-serialized models.

## Coverage audit

| Deleted test | Invariant | Status |
| --- | --- | --- |
| **ParseTest** (17) | parse-shape decode of C++ constructs | **Confirmed covered** — parse-shape decode is now cppfrontend's `TypeBuilder`/`ModelBuilder` job (its ~100+ construction self-checks + `:cppfrontend:goldenCompare`), and the resulting model→resolve→codegen shapes are asserted by the surviving krapper_gen tests (`CppCodeTests` 83, `CppHeaderTests` 80, `FullKotlinTests` 80, `KotlinCodeTests`) + `:featuregen:nativeTest` 188. Parse-dependent, not re-created. |
| **ReturnShapeTest** (15) | return-style determination (by-value/ptr/ref/STRING/range) | **Confirmed covered** — `ReturnStyle` is computed during resolve; the surviving model→resolve→codegen tests assert the emitted C++/Kotlin that encodes each style, and featuregen exercises by-value/ptr/ref/STRING/range end-to-end across 188 behavioral tests. |
| **ForcingCharacterizationTest** (4) | multi-instantiation forcing | **Confirmed covered** — featuregen forces 17 instantiations (incl. cumulative range-accessor recovery, the path #40 added) through the same `ForcingHeader` the cppfrontend binary synthesizes; 188/0 on the self-hosted front-end. |
| **SkipNotCrashTest** (9) | unmodelable shape DROPS + logs, doesn't crash (core invariant) | **NEW test** — `SkipNotCrashResolveTest` (below). cppfrontend also skips unmodelable shapes at construction and featuregen has dropped members, but there was no explicit *drop-not-crash* assertion on the krapper_gen side after the flip. |
| **DeterminismTest** (1) | two in-process generations byte-identical (#11a hermeticity) | **NEW test** — `DeterminismTest` re-added at the self-hosted boundary (below). The original guarded the (now-deleted) cursor→element memo plus two process-globals that survive the flip. |

## New tests

**`SkipNotCrashResolveTest`** (3) — drives the core skip-not-crash contract through the real `findClasses { defaultFilter() }.resolveAll(...)` path over hand-built models (resets `DropLedger` + `GenerationContext` first, as `IndexedServiceImpl.init` does):
- an unresolvable return type (under `IGNORE_MISSING`) **drops the offending method while its bindable siblings still resolve**, and the drop is ledgered through the central `ResolveContext.notifyFailed` choke point as a `RESOLVE`-phase record **with a reason** — it does not throw;
- a fully-bindable class drops nothing (the ledger does not false-positive);
- a const/non-const accessor pair de-dups to the const half (`DEDUP` phase), ledgered;
- `DropLedger.hasDrops()` — the exact predicate the opt-in `--fail-on-drop` gate reads (`IndexedServiceImpl.reportDrops`) — is asserted true on a drop and false on a clean run (closes the deleted L3 case).

**`DeterminismTest`** (1, the #11a guard relocated) — runs the real `ModelIo.decodeFromString → resolve → emit (HeaderWriter + CppWriter + KotlinWriter)` pipeline TWICE in one process over the same `ModelIo` JSON and asserts byte-identical output. Each run deserializes a FRESH model and resets the process-scoped state, so the only thing carried between runs is leaked global state. This guards the two process-globals the original protected that survive the flip — the `GenerationContext` `WrappedType` intern cache + `rootPackage`, and the `DropLedger` — i.e. reproducibility of the self-hosted output, which the stage-0 bootstrap depends on. (Omits the clang COMPILE + `.def`, which need the original headers on disk; determinism of the emitted SOURCE is the bootstrap property.) **Sanity-checked: temporarily leaking a process-global counter into one emitted file made it go RED, confirming it guards cross-run divergence.**

## Genuine gaps

None left open. The three confirmed-covered invariants are parse-shape/return-style/forcing decode that genuinely moved into cppfrontend + the surviving resolve/codegen + featuregen layers; the two that had no explicit post-flip assertion (skip-not-crash drop, determinism) now do.

## Verify (JDK 17)

- `:krapper_gen:nativeTest` → **276/0** (272 surviving + 4 new: `SkipNotCrashResolveTest` 3, `DeterminismTest` 1).
- `:featuregen:nativeTest -PenableClang` → **188/0** (self-hosted via cppfrontend, no libclang).
- `:krapper_gen:ktlintCheck` → green.
- Determinism test verified to FAIL on an injected cross-run leak, then reverted.

### DIFF
--- a/krapper_gen/src/nativeTest/kotlin/DeterminismTest.kt
+++ b/krapper_gen/src/nativeTest/kotlin/DeterminismTest.kt
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.krapper.generator
+
+import com.monkopedia.krapper.ReferencePolicy
+import com.monkopedia.krapper.generator.builders.CppCodeBuilder
+import com.monkopedia.krapper.generator.codegen.CppWriter
+import com.monkopedia.krapper.generator.codegen.File
+import com.monkopedia.krapper.generator.codegen.HeaderWriter
+import com.monkopedia.krapper.generator.codegen.KotlinWriter
+import com.monkopedia.krapper.generator.model.MethodType
+import com.monkopedia.krapper.generator.model.ModelIo
+import com.monkopedia.krapper.generator.model.WrappedArgument
+import com.monkopedia.krapper.generator.model.WrappedClass
+import com.monkopedia.krapper.generator.model.WrappedConstructor
+import com.monkopedia.krapper.generator.model.WrappedMethod
+import com.monkopedia.krapper.generator.model.WrappedNamespace
+import com.monkopedia.krapper.generator.model.WrappedTU
+import com.monkopedia.krapper.generator.model.type.WrappedType
+import com.monkopedia.krapper.generator.model.type.WrappedType.Companion.const
+import com.monkopedia.krapper.generator.model.type.WrappedType.Companion.referenceTo
+import com.monkopedia.krapper.generator.resolvedmodel.ResolvedElement
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Hermeticity lock at the SELF-HOSTED boundary (issue #89, the #11a determinism guard
+ * relocated).
+ *
+ * The deleted [DeterminismTest] proved that two sequential generations IN ONE PROCESS over
+ * the same input were byte-identical — its load-bearing job was catching a process-global
+ * that leaked across runs (the `WrappedType` intern cache + `rootPackage` in
+ * [GenerationContext], the [DropLedger], and the now-removed libclang cursor->element memo).
+ * It drove that through the in-process libclang parse the self-hosting flip (B5, #88)
+ * deleted, so it can't be ported as-is. Two of its three guarded process-globals survive
+ * the flip (GenerationContext + DropLedger), and determinism of the self-hosted output is
+ * exactly the property the stage-0 bootstrap depends on.
+ *
+ * This test relocates the guard onto the surviving boundary: it runs the real
+ * deserialize -> resolve -> codegen pipeline (the path the self-hosted CLI takes from a
+ * cppfrontend-produced ModelIo, minus the clang COMPILE) TWICE in one process and asserts
+ * byte-identical output. Each run DESERIALIZES A FRESH model from the same ModelIo JSON and
+ * resets the process-scoped state (as `IndexedServiceImpl.init` does) before resolving, so
+ * the only thing carried between runs is leaked global state — if the intern cache,
+ * rootPackage, or DropLedger leaked run #1 into run #2, the output would diverge and this
+ * fails. The model is shaped to MUTATE during resolution (the `value()` overload pair, the
+ * `const Dep&` accessor that pulls `Dep` in under INCLUDE_MISSING) so a non-idempotent reuse
+ * of carried-over state has the best chance to surface.
+ */
+class DeterminismTest {
+
+    private fun method(name: String, returnType: WrappedType, isConst: Boolean = false) =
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard — tier 1 auto-approve.

**Category:** `test_only` — two new nativeTest files, zero production changes.

**Scrutiny (all items checked):**

1. **Tests are MEANINGFUL, not tautological.** `SkipNotCrashResolveTest` drives the real `findClasses { defaultFilter() }.resolveAll(...)` pipeline on a hand-built `WrappedTU` with a genuinely unresolvable return type (`Skip::Missing`). The IGNORE_MISSING policy maps it to `RemoveElement`, `mapAndResolve` returns null, `notifyFailed` records a RESOLVE-phase drop with a non-blank reason, and the sibling `value()` still resolves — all three assertions guard the exact production path, not a stub. It would fail immediately if resolve threw instead of dropped (exception propagates through `runBlocking`). The `hasDrops()` assertion ties the test directly to `IndexedServiceImpl.reportDrops`'s `--fail-on-drop` gate. DEDUP test exercises the real `dropConstOverloadDuplicates` path in `modifyMethodsIfNeeded`.

   `DeterminismTest` drives the real `ModelIo.decodeFromString → ParsedResolver → resolveAll(INCLUDE_MISSING) → HeaderWriter + CppWriter + KotlinWriter` pipeline twice, with `DropLedger.reset() + GenerationContext.reset(null)` between runs mirroring `IndexedServiceImpl.init` exactly. The RED-on-injected-leak claim is structurally sound: any accumulation in `GenerationContext.internedTypes` or `DropLedger.records` that isn't cleared would produce divergent output and fail the per-file `assertEquals`.

2. **Audit is honest.** `ParseTest`/`ReturnShapeTest`/`ForcingCharacterizationTest` genuinely moved: parse-shape decode is now cppfrontend's domain (its self-checks + `:cppfrontend:goldenCompare`); `ReturnStyle` is asserted via the 83+80+80+ surviving codegen tests + featuregen 188/0; forcing is exercised by featuregen's 17 instantiation requests. No overclaim. The two gaps with no explicit post-flip assertion (drop-not-throw, determinism) now have explicit tests.

3. **Tests only.** Confirmed: two new files in `krapper_gen/src/nativeTest/kotlin/`, no production source touched.

4. **Determinism test guards the right globals.** Resets `GenerationContext` (intern cache + rootPackage) and `DropLedger` — the two process-globals that survived the B5 flip. The cursor→element memo guarded by the original test was deleted with the libclang reducer and is correctly omitted.

**Style (axis B):** Snake_case test method names are consistent with `FalseDefaultTest.kt`'s established pattern in this module; mixed camelCase/snake_case is pre-existing codebase convention. Comment quality is well above average — each test's docstring precisely describes the invariant and the deletion motivation.

**Build:** `:krapper_gen:nativeTest` → **276/0** (SkipNotCrashResolveTest: 3/0, DeterminismTest: 1/0, all 272 surviving passing). `:krapper_gen:ktlintCheck` → green.

PR #95, opened 2026-06-14
krapper_gen: constructor-factory call uses canonical plainDeclaredName (Fixes #81)

### LINKED ISSUE #81: [codegen] constructorMethod uses type.plainName for the CONSTRUCTOR factory call — same hazard class as #73/#75
Follow-up confirmed during review of PR #80 (#75). `constructorMethod` (KotlinWriter.kt ~2555) builds the CONSTRUCTOR factory call via `type.plainName` — the same per-file import-alias hazard #73 (declaration side) and #75 (the _Holder ARG_CAST call) addressed. If a constructor factory is ever called cross-file from a file that aliases the type via an import collision, the call resolves to `<Alias>...` instead of the canonically-declared factory → undeclared symbol. Distinct path from #75's ARG_CAST _Holder (so correctly out of #75 scope), not currently exercised by featuregen (byte-identical holds), but the same latent bug. Fix mirrors #75: use the canonical `plainDeclaredName` for the factory-call symbol so the call matches the canonical declaration; add a nativeTest. Low priority (latent, no current trigger).

monkopedia-coder:
Fixes #81 — the latent CONSTRUCTOR-factory-call sibling of #73 (declaration side) and #75 (the `_Holder` factory call). This closes the #73/#75/#81 declaration-name-vs-per-file-alias family.

## The bug
`KotlinWriter.constructorMethod` (~line 2555) builds the CALL to a class's generated `MemScope.<Type>(...)` CONSTRUCTOR factory via `extensionMethod(type.pkg, type.plainName)`. `plainName` routes through `mappedName`, which applies the per-file import-alias `remap` (shared mutable state stamped onto the `ResolvedKotlinType` by `ImportBlock.build`). The constructor factory is a generated SIBLING function, not an imported type reference — it is declared under the canonical class name. So a remote calling file that aliased the constructed type via an import collision (`clang.attr.Kind` -> `AttrKind`) would emit `AttrKind(...)`, an undeclared symbol, since the factory is declared as `Kind(...)`. Same bug class as #73/#75, latent today (no featuregen import collision triggers it).

## The fix
The single call-site change, mirroring #75's `_Holder` fix exactly:

```
-    private fun constructorMethod(type: ResolvedKotlinType) = extensionMethod(
-        type.pkg,
-        type.plainName
-    )
+    private fun constructorMethod(type: ResolvedKotlinType) = extensionMethod(
+        type.pkg,
+        type.plainDeclaredName
+    )
```

`plainDeclaredName` (= `qualifyList.last()`, ignoring `remap`) is the canonical declared name. Declaration-name == call-name; neither follows the per-file alias. A comment cites the invariant, consistent with #73/#75. Only the factory-call symbol changed — type references in the args/cast legitimately keep the alias (the #73 distinction).

## Test
`ConstructorCallRemapTest` mirrors `HolderCallRemapTest`: it builds the constructor-factory call symbol exactly as `constructorMethod` does, sets a `remap` aliasing the type, and asserts the call name. Under the alias, the fixed accessor (`plainDeclaredName`) stays canonical `Kind`, while the pre-fix accessor (`plainName`) follows the alias to `AttrKind` — locking the exact regression the fix removes (the accessor-contrast approach the issue accepts when a faithful e2e repro is impractical at unit level).

## Verify (JDK 17)
- `:krapper_gen:nativeTest` → 277/0 (276 + the new test)
- `:krapper_gen:ktlintCheck` → green
- `:featuregen:nativeTest -PenableClang` → 188/0 (no collision, so unchanged — confirms no regression)

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -2552,9 +2552,18 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
         memScope: Symbol
     ): Symbol = Call(constructorMethod(type), ptr, memScope)
 
+    // CALL the generated `MemScope.<Type>(...)` CONSTRUCTOR factory by its canonical
+    // DECLARED name (`plainDeclaredName`, ignoring the per-file import-alias `remap`),
+    // matching the factory's declaration. The factory is a generated SIBLING function,
+    // not an imported type reference: a remote calling file that aliased `type` via an
+    // import collision would, through `plainName`, emit `<Alias>(...)` — an undeclared
+    // symbol, since the factory is declared under the canonical name regardless of the
+    // caller's alias. Declaration-name == call-name; neither follows the alias. This is
+    // the constructor-factory sibling of the `_Holder` call fix (#75) and the enum/class
+    // declaration fixes (#73), completing the #73/#75/#81 remap-hazard family.
     private fun constructorMethod(type: ResolvedKotlinType) = extensionMethod(
         type.pkg,
-        type.plainName
+        type.plainDeclaredName
     )
 
     private fun KotlinCodeBuilder.generateStringReturn(call: Symbol, free: Boolean = true) {
--- a/krapper_gen/src/nativeTest/kotlin/ConstructorCallRemapTest.kt
+++ b/krapper_gen/src/nativeTest/kotlin/ConstructorCallRemapTest.kt
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Tier 1 — bug_fix — LGTM.

**Fix (call-side):** `constructorMethod` correctly switches from `type.plainName` (routes through the per-file import-alias `remap`) to `type.plainDeclaredName` (canonical, `qualifyList.last()`, ignores the alias). The constructor-factory is a generated sibling declared under the canonical name; the call must match the declaration regardless of the calling file's alias. Mirrors #75's `_Holder` fix exactly.

**Scope verified:** Only the call-symbol changed — args and cast type references retain the alias per the #73 distinction. The declaration sites (~1461/1524, using `.name`) are correctly left for the separate #96 follow-up; that's a declaration-side issue and genuinely separate from this call-side fix.

**Test (`ConstructorCallRemapTest`):** Structural mirror of `HolderCallRemapTest`. Builds the symbol exactly as `constructorMethod` does (`extensionMethod(type.pkg, type.factoryName())`), stamps a `remap` alias (`clang.attr.Kind` → `AttrKind`), and asserts:
- `plainDeclaredName` → `Kind` (canonical, fixed accessor)
- `plainName` → `AttrKind` (alias, the pre-fix bug)
Would fail pre-fix. Accessor-contrast pattern is the accepted approach where featuregen has no import collision for an e2e repro.

**Build (JDK 17):** `:krapper_gen:nativeTest` 277/0 (incl. `ConstructorCallRemapTest`), `:krapper_gen:ktlintCheck` green, `:featuregen:nativeTest -PenableClang` 188/0 (no regression).

PR #97, opened 2026-06-14
krapper_gen: dedup namespace free functions across forcing passes (#62)

### LINKED ISSUE #62: krapper_gen: namespace free functions still emit 18x sections per function (sibling of #60)
Noted as out-of-scope in #61 (PR description: "namespace free functions still duplicate 18x in .h/.cc and in the Kotlin facades").

`ResolvedMethod` elements (free functions / namespace-scoped methods) are not collapsed by `dedupClassesLastWins` — intentionally, since `KotlinWriter` also never deduped them, so fixing them would change the Kotlin baseline. They accumulate one copy per `requestInstantiation` round (18x for featuregen), producing 18 identical declarations in the .h and 18 `_`-uniquified definitions in the .cc, plus correspondingly-named Kotlin facades (`_default_handler`, `__default_handler`, ...).

Fix should mirror #60: dedup `ResolvedMethod` elements by their qualifying identity (e.g. `forcingIdentity` or namespace-qualified name) with last-wins semantics, after verifying that the Kotlin baseline is byte-identical post-dedup.

monkopedia-coder:
Fixes #62.

## Problem (still reproduces post-flip)

Namespace FREE FUNCTIONS still emitted one wrapper section per instantiation pass (18×) on the post-flip cpp front-end. #60 (PR #61)'s `dedupClassesLastWins` collapsed only `ResolvedClass` copies and let `ResolvedMethod` (namespace free functions) pass through untouched, so every `requestInstantiation` re-appended a re-resolved copy of each bound free function. The forcing + dedup machinery survived the flip, so the duplication reproduces via the cpp path (`featuregen:kplusplusSync -PenableClang`):

- `featuregen.h`: **18** `hh_default_handler` declarations
- `featuregen.cc`: **18** `default_handler` definitions (`_`-uniquified into dead `_default_handler`/`__default_handler`/... variants)
- `src/hh_Functions.kt`: **18** Kotlin facades — `default_handler`, `_default_handler`, `__default_handler`, ... 18 deep

## Fix

Extend `dedupClassesLastWins` to also collapse top-level `ResolvedMethod`s (namespace free functions), last-wins, mirroring the class semantics. Classes are keyed by type-string; free functions by their full `forcingIdentity` — `qualified::name(arg-type signature)#uniqueCName` — so genuine **OVERLOADS** (same name, different parameters) stay distinct and survive as separate sections. Any other top-level element still passes through untouched. (`forcingIdentity` was made `internal` so the resolver can reuse the exact identity `IndexedServiceImpl` already uses to decide keep-vs-drop.)

## Diff characterization (output shrinks, nothing else)

Regenerated via the cpp front-end and diffed against a parent-commit regen. Only the 6 free-function-related artifacts differ; **zero added lines**, all deletions:

| | baseline | fixed |
|---|---|---|
| `featuregen.h` `hh_default_handler` decls | 18 | **1** |
| `featuregen.cc` `default_handler` defs | 18 | **1** |
| `hh_Functions.kt` facades | 18 | **1** (clean `default_handler`) |
| `std_Functions.kt` `abs` sections | 90 (5 overloads × 18) | **5** (overloads preserved) |
| `std_Functions.kt` total funcs | 241 | 37 |

Overloads are NOT collapsed into each other (`abs` keeps its 5 distinct C++ signatures; the two that share a Kotlin spelling stay `_`-disambiguated by the writer as before). No class `.kt`, and no non-free-function section, changed.

## Stability

Two sequential `kplusplusSync -PenableClang` runs produce byte-identical generated sources.

## Tests

New `FreeFunctionDedupTest` (krapper_gen nativeTest) locks the invariant directly on `dedupClassesLastWins`:
- N (=18) forcing passes of a namespace free function collapse to ONE section
- genuine overloads (same name, distinct parameter types) survive as separate sections
- dedup is order-independent (forward vs reversed copy order → same survivors)

## Verify

- `:krapper_gen:nativeTest` → **280/0** (was 277; +3 new), `:krapper_gen:ktlintCheck` clean
- `:featuregen:nativeTest -PenableClang` → **188/0** on the new (smaller) baseline
- `:cppfrontend:handoffGenerate -PenableClang` → green

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -59,7 +59,7 @@ import com.monkopedia.krapper.generator.resolvedmodel.type.ResolvedType
 // re-parse was ALREADY main-bound (keep) or is incidental (drop). Two distinct functions
 // that happen to mangle to the same uniqueCName get different identities here, so a
 // collision can't false-keep an incidental function (which would emit non-compiling Kotlin).
-private val ResolvedMethod.forcingIdentity: String
+internal val ResolvedMethod.forcingIdentity: String
     get() = "$qualified::$name(${args.joinToString(",") { it.type.toString() }})#$uniqueCName"
 
 class IndexedServiceImpl(private val config: KrapperConfig, private val request: IndexRequest) :
@@ -221,10 +221,12 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
             // walk omits. See ResolveContext.forcingTargetKeys.
             forcingTargets = setOf(target)
         )
-        // One entry per class type-key, last copy wins (#60): without this, every
-        // writer received a fresh full copy of each bound class per instantiation
-        // request and the C++ writers emitted a complete wrapper section per copy
-        // (18 sections per class in featuregen). See dedupClassesLastWins.
+        // One entry per class type-key AND per free-function identity, last copy wins
+        // (#60 classes, #62 free functions): without this, every writer received a fresh
+        // full copy of each bound class and each bound namespace free function per
+        // instantiation request and emitted a complete wrapper section per copy (18
+        // sections each in featuregen). Overloads stay distinct (keyed on the full
+        // argument-type signature). See dedupClassesLastWins.
         classes = (
             classes + newClasses.filter {
                 it !is ResolvedMethod || it.forcingIdentity in boundMethodIds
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. Approving and merging.

**Fix correctness — overload-safe ✓**
`forcingIdentity` = `qualified::name(C++ arg-type sig)#uniqueCName`. Two `abs` overloads with different parameter types produce distinct keys; the 90→5 reduction (18 passes × 5 overloads → 5 distinct survivors) is consistent with this key design. Genuine overloads are not collapsed.

**Class dedup unchanged ✓**
Classes key on `"class:$type"` — semantically identical to the original `type.toString()` key (just namespace-prefixed to prevent any theoretical cross-type collision). Class dedup output is byte-identical.

**Only free-function sections changed ✓**
Non-class, non-method elements return `null` from `dedupKey()` and pass through via the `?: return@mapNotNull element` fast-path. Scoped cleanly.

**`internal` visibility for `forcingIdentity` ✓**
Correct visibility to share between `IndexedServiceImpl.kt` and `Resolver.kt` within the same module. Not exposed publicly.

**`toMap()` last-wins semantics ✓**
Matches the original `associateBy` behavior — later duplicate keys overwrite earlier ones in the `LinkedHashMap`.

**Tests meaningful ✓**
3 tests in `FreeFunctionDedupTest`: (1) 18 forcing copies → 1 section, (2) overloads stay distinct, (3) order-independent. All exercise the real resolver machinery. Would fail pre-fix.

**Build results:** `:krapper_gen:nativeTest` 280/0 (+3 new), `:krapper_gen:ktlintCheck` green, `:featuregen:nativeTest -PenableClang` 188/0, `:cppfrontend:handoffGenerate -PenableClang` green.

Issue #99, opened 2026-06-14
[example] make the v8 binding consumer cpp-ready post-flip


monkopedia-coder:
## Scope: make the `example/kotlin_project` v8 binding consumer cpp-ready post-flip

Investigation only (no fix implemented). Worktree `scope/v8-cpp-readiness`. Machine: LLVM/clang 22.1.6, prebuilt `:cppfrontend` + `:krapper_gen` release binaries reused for experiments.

### TL;DR
The "v8 SIGABRTs / zero Kotlin bindings" symptom is **NOT a parse crash on V8's templated surface**. Root-caused to ONE plumbing gap: `headerDirectory("../include/")` is never threaded as `-I` into either the cppfrontend parse OR the krapper_gen wrapper compile. With the include path supplied manually, the cpp front-end **parses v8 cleanly (13,770 names, all key classes, ~1s)** and krapper_gen **resolves 508 v8 classes and generates 249 Kotlin files + a compiling C++ wrapper (rc=0)**. The flip did not regress v8 in any V8-specific way; v8 just rides the same plumbing that single-header modules (featuregen) never exercised.

---

### 1. example/v8 current post-flip state — DEAD at sync, but not for the reason assumed
- Standalone build: `example/kotlin_project/settings.gradle.kts` does `includeBuild("../../")` and applies `id("com.monkopedia.kplusplus.compiler")` (the v2 compiler plugin = `compiler/gradle`), so it builds against the **in-tree** plugin/krapper_gen, not a published artifact.
- Module name is `target.name` = `kotlin_project` (root project name), so the frontend property it needs is `kpp.frontend.kotlin_project`.
- Post-flip `kplusplusSync` requires one of `=cpp` / `=seed`. The example sets **neither** (its `gradle.properties` is just cache/heap). So `kplusplusSync` hard-fails with the actionable: *"the in-tree libclang front-end was removed in flip B5 (#47) … Set kpp.frontend.kotlin_project=cpp or =seed."* There is **no committed `krapped/` seed** under `example/kotlin_project/`, so `=seed` also fails. The module is NOT in the root build/CI, so nothing caught this.
- The historical libclang `exit 134` is moot: libclang is deleted. v8 never reaches a parse today — it stops at task configuration/sync.

### 2. Can cppfrontend PARSE v8? — YES, cleanly (once `-I` is provided)
[...]

PR #100, opened 2026-06-14
plugin+tools: thread headerDirectory(...) as -I into the cpp front-end pipeline (#99)

monkopedia-coder:
## The gap

A module's `kplusplus { headerDirectory("../include/") }` include dirs were **dropped** on the cpp front-end path. Today only `-I<krappedDir>` reaches the cpp parse and the wrapper compile, so cross-directory quote-includes (e.g. v8's `cppgc/*.h` reaching `v8config.h`) failed and the wrapper compile aborted with **exit 134**.

## The generic fix — two `-I` seams

Mirrored across the parse and the compile so a model parsed against a header set and a wrapper compiled against it agree:

- **`cppfrontend --parity-emit`**: the trailing args now mix `-I<dir>` tokens with the instantiation specs. `parityEmit` partitions on the `-I` prefix (C++ type specs never start with `-I`) and folds the include flags into the `\n`-joined driver args, so cross-directory includes resolve in BOTH the base and the per-spec forcing parses.
- **`krapper_gen --include-dir`** (repeatable): carried on `IndexRequest.headerDirectories` into `CompileFlags.extraIncludeDirs`, so the generated C++ wrapper compile (`CppCompiler`) AND the cinterop `.def` `compilerOpts` (`DefWriter`) get the same `-I` set the front-end parsed under.

The kplusplus gradle plugin resolves `ext.headerDirectories` to absolute paths and threads them into both seams. `resolveCppFrontend()` now locates the `:cppfrontend` binary across **both** layouts — sibling project (featuregen/cppfixture) and **included build** (a standalone consumer pulling kplusplus in via `includeBuild`) — the same way `resolveKrapperGen()` already does; without it the standalone v8 example could not find cppfrontend (it lives in the included build).

This is a generic plugin/tool capability, not a v8 special-case.

## Example wiring

The standalone v8 example (`example/kotlin_project/`) is flipped to the cpp front-end: `gradle.properties` pins `kpp.frontend.kotlin_project=cpp`, `cppStandard = "c++17"`, and its v8 `fixup { }` is updated to the cpp front-end's `uniqueCName` spellings (the `Persistent<v8::Value>` copy ctor/assign — NonCopyable; and `Maybe<void>`'s phantom primary-template methods). Build with `-PenableClang`.

## Verification

- `:featuregen:nativeTest -PenableClang` → **188/0** (the self-hosted parity gate; the generic change does not regress the existing cpp modules).
- example `kplusplusSync -PenableClang` → cpp front-end generates **509 Kotlin binding files** AND the **C++ wrapper COMPILES (no exit 134)**. The full Kotlin/Native link against the 62 MB v8 monolith is **brick 3** and out of scope here.

### Note for reviewers

The `Persistent<v8::Value>` NonCopyable special members surface **parse-dependently** under the cpp front-end — one parse materializes the implicit copy ctor (`..._new__const_v8_Persistent_and`), another the copy-assignment (`..._op_assign`) — so the example drops **both** to stay robust across parses. Flagging this implicit-special-member non-determinism as a follow-up observation; it does not affect this brick's result.

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -72,10 +72,13 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
             // -PenableClang); the default path never references it. This replaces the old
             // per-module manual `dependsOn(":cppfrontend:featuregenCppBindings")` wiring —
             // ANY module flipped to cpp now self-wires the front-end build generically.
-            if (target.findProperty("kpp.frontend.${target.name}") == "cpp" &&
-                target.rootProject.findProject(":cppfrontend") != null
-            ) {
-                it.dependsOn(":cppfrontend:linkReleaseExecutableKlinker")
+            // resolveCppFrontend covers BOTH the sibling layout (featuregen/cppfixture) and
+            // the included-build layout (the standalone v8 example, where cppfrontend lives in
+            // the kplusplus build pulled in via includeBuild) — its link task is null when
+            // clang isn't enabled, in which case the doLast fails fast with the -PenableClang
+            // guidance.
+            if (target.findProperty("kpp.frontend.${target.name}") == "cpp") {
+                resolveCppFrontend(target).first?.let { ct -> it.dependsOn(ct) }
             }
             val krappedDir = krappedDirFor(target)
             val manifestFile = File(krappedDir, "requested.txt")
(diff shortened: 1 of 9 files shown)


monkopedia-reviewer (approved):
**APPROVE** — main-merge-quality review. Correct, generic, backward-compatible. Verified the diff against the surrounding code; no correctness blockers.

### Partition robustness (parityEmit `-I` vs spec) — sound
`trailing.partition { it.startsWith("-I") }` cannot misroute: a C++ type spec is an identifier/qualified-name (`v8::Persistent<v8::Value>`, `std::vector<int>`) and can never begin with `-`. Include flags are always `-I<abspath>` (plugin emits `includeDirs.map { "-I$it" }`). Backward-compat holds: with no `-I` tokens, `includeFlags` is empty and the new `listOf("-std=…","-resource-dir=…").joinToString("\n")` reproduces the prior `"-std=$std\n-resource-dir=…"` byte-for-byte, so existing featuregen callers are unaffected (consistent with the 188/0 gate).

### resolveCppFrontend correctness — mirrors resolveKrapperGen faithfully
Same two-layout shape (`Pair<Any?, File>`): direct sibling via `rootProject.findProject`, else iterate `gradle.includedBuilds` with a try/catch around `included.task(...)`, else hardcoded fallback. The dependency wiring is now actually sound: the old `it.dependsOn(":cppfrontend:linkReleaseExecutableKlinker")` was guarded on `findProject(":cppfrontend") != null`, which is null in the included-build (standalone example) topology — so the dependsOn was silently skipped there, and a string task-path wouldn't resolve cross-build anyway. The new `resolveCppFrontend(target).first?.let { ct -> it.dependsOn(ct) }` uses the included-build `TaskReference`, exactly mirroring the existing `linkTask?.let { lt -> it.dependsOn(lt) }` at line 68. Good. (Minor, non-blocking: the direct branch uses `direct.layout.buildDirectory.file(...)` while resolveKrapperGen uses `rootProject.file(...)`; the layout form is if anything more correct — fine.)

### End-to-end `extraIncludeDirs` threading — reaches both seams, no drops
App.kt `--include-dir` (repeatable) → `IndexRequest.copy(headerDirectories = (default + includeDir).distinct())` preserves the header-parent default and dedups → IndexedServiceImpl passes `request.headerDirectories` to BOTH `DefWriter.generateDef` (→ `.def` `compilerOpts`) AND `CppCompiler.compile` (→ wrapper compile). All three signatures (`CompileFlags`, `CppCompiler.compile`, `DefWriter.generateDef`) add the param with `= emptyList()` defaults, so all other callers are source-compatible (no positional breakage; no test callers). The CompileFlags refactor (`headerParents + extraIncludeDirs` over a `toSet()`) is dedup- and order-stable, so it cannot regress the default (empty) case or determinism.

### v8 special-casing not leaked into core — confirmed
The capability is generic (`headerDirectory(...)` → `-I`); nothing v8-specific entered the plugin/tools. The pre-existing hardcoded `-DV8_COMPRESS_POINTERS` in CompileFlags is untouched/not expanded (correct — that wart is out of scope).

### Example fixups — reasonable config, not bug-masking
The `Maybe<void>` and `Persistent<v8::Value>` `removeMethod` updates are example-config retargeting to the cpp front-end's `uniqueCName` spellings. A `removeMethod` that didn't match would be a no-op leaving an uncompilable wrapper, but the wrapper DID compile (509 files, no exit 134), so the load-bearing drops took effect.

### Determinism smell — non-blocking, follow-up filed
The `Persistent<v8::Value>` (NonCopyable) implicit copy-ctor-vs-copy-assignment materializing **parse-dependently** is a genuine cpp-front-end non-determinism. But: this PR does not introduce it, is correctly defensive (drops both members), is fully gated (`-PenableClang` + `kpp.frontend.*=cpp`), and does not regress the 188/0 parity gate. Tracking as a follow-up rather than blocking — agree with the lean.

Tier-1, green → squash-merging.

PR #102, opened 2026-06-15
example(v8): LINK + RUN the hello-world against the 62MB monolith (#99)

monkopedia-coder:
## Goal

Brick 3 of the v8 self-host campaign (#99): prove the cpp-front-end-generated v8 bindings (brick 1, #100) actually **LINK** against `../libv8_monolith.a` (62MB) and **RUN** the JS hello-world in `example/kotlin_project/src/nativeMain/kotlin/Main.kt`.

## Result — STRETCH/flagship reached: it RUNS

```
Got args: 0
Executing kotlin
Got: Hello, Worldy!
```

`./gradlew runDebugExecutableKlinker -PenableClang` links against the monolith and V8 executes the JS `'Hello' + ', Worldy!'`, returning the string end-to-end.

## The link recipe (PORT of featuregen's #78)

The cpp front-end parses v8 against the SYSTEM libstdc++ (gcc-16), so the generated wrapper references newer-libstdc++ internals K/N's bundled gcc-8.3 sysroot can't resolve. The example's `klinkedExecutable` now links the REAL system `libstdc++.so` by ABSOLUTE PATH (replacing `-lstdc++`, which shadows to the bundled old lib via search order) + `--gc-sections` + `--allow-shlib-undefined` + `-rpath`/`-rpath-link`. The `.so` is discovered at configure time via `clang++ -print-file-name=libstdc++.so`. Same mechanism clangwalk/cppfrontend/featuregen already use, applied to the EXECUTABLE binary.

## Main.kt adapted to the cpp-generated surface

Main.kt was written against the old libclang surface; the cpp front-end produces a slightly different shape. **No `instantiate()` specs were needed** — the methods were all present, only the surface shape differed:

- `NewSingleThreadedDefaultPlatform` / `NewFromUtf8`: the cpp front-end does not materialize ENUM-/CLASS-typed default arguments (only literal ones like `length = -1`), so the enum (`IdleTaskSupport.kDisabled`, `NewStringType.kNormal`) and `unique_ptr` defaults are now passed explicitly.
- `DeserializeInternalFieldsCallback()` → `DeserializeInternalFieldsCallback_Holder()` (and the `unique_ptr<TracingController>` Holder): a `_Holder()` only allocates storage (no ctor), and the wrapper consumes these by-value with `std::move`, so the storage is zeroed (empty `unique_ptr` / `{nullptr,nullptr}` callback) before use.
- `_reference()` → `reference()`; `InitializePlatform(platform.get())` → `InitializePlatform(platform)` (the generated factory already `.release()`s the `unique_ptr<Platform>`).

## Generator/codegen fixes the v8 surface exposed

Brick 1 only proved the C++ **wrapper** compiles; brick 3 is the first to compile + LINK the Kotlin bindings, surfacing four latent issues (all default-off / additive):

1. **bool-underlying enum** (`KotlinWriter.literalForUnderlying`): a C++ `enum class : bool` (v8's `MessageLoopBehavior`) reduces to Kotlin `Boolean`, but entry values were emitted as Int `0`/`1`. Now emits `false`/`true`.
2. **`removeMethod` now reaches TOP-LEVEL free functions** (`IndexedServiceImpl.applyResult`): a parentless element lives directly in `classes`, so `element.parent?.removeChild` silently no-op'd — `removeMethod` could not drop namespace free functions at all. Now drops from the top-level list when `parent == null`. This lets the example drop the uncompilable libstdc++ `<ios>`/`std::byte` bitmask `operator|=`/`&=`/`^=` wrappers (reference-return collapsed to value; the C wrapper is itself UB — returns a pointer to a by-value local).
3. **`noRtti` config knob** (`KrapperConfig`/`App --no-rtti`, `KPlusPlusExtension.noRtti`, threaded via `GenerationContext` to `CastTargets.downCastTargetsFor`): the v8 monolith is built `-fno-rtti` and exports no `typeinfo`, so generic `dynamic_cast<D*>` down-cast helpers fail to LINK (`undefined reference to typeinfo for v8::...`). Under `noRtti` those generic down-casts are skipped (RTTI-free LLVM `classof` down-casts unaffected). The example sets `noRtti = true`.

## `removeMethod` fixups added to the example (18, grouped + commented)

Methods that emit uncompilable Kotlin and are NOT on the hello-world path — two latent codegen-bug families filed as krapper_gen follow-ups:
- **Out-pointer param modeled as a value** (`Maybe::To`, `String::GetExternalStringResourceBase`, `Isolate::GetCodeRange`/`GetEmbeddedCodeRange`, `TraceBufferChunk::AddTraceEvent`, `Delegate::ReallocateBufferMemory`).
- **Reference-return of a value-reduced type** (the `std::byte` / `<ios>` `_Ios_*` bitmask `*=` operators — pure transitive-include noise).

## Verification

- `:krapper_gen:nativeTest` green (41s, no failures).
- `:krapper_gen:ktlintCheck` + `:krapper_model:ktlintCheck` green.
- `runDebugExecutableKlinker -PenableClang` links against the 62MB monolith and prints `Got: Hello, Worldy!`.
- All generator changes are default-off/additive, so the default build and other consumers (featuregen committed bindings) are unaffected.

## Files changed

- `example/kotlin_project/build.gradle.kts` (link recipe + `noRtti=true` + 18 removeMethod fixups)
- `example/kotlin_project/src/nativeMain/kotlin/Main.kt` (surface adaptation)
- `krapper_gen/.../codegen/KotlinWriter.kt` (bool enum)
- `krapper_gen/.../IndexedServiceImpl.kt` (free-function removal)
- `krapper_gen/.../codegen/CastTargets.kt` (noRtti down-cast skip)
- `krapper_gen/.../KrapperConfig.kt`, `App.kt` (`--no-rtti`)
- `krapper_model/.../GenerationContext.kt` (`noRtti` carrier)
- `compiler/gradle/.../KPlusPlusExtension.kt`, `KPlusPlusCompilerGradlePlugin.kt` (DSL `noRtti` + `--no-rtti` arg)

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -109,6 +109,7 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
                 it.inputs.property("instantiations", e.instantiations.sorted())
                 it.inputs.property("referencePolicy", e.referencePolicy ?: "")
                 it.inputs.property("cppStandard", e.cppStandard ?: "")
+                it.inputs.property("noRtti", e.noRtti)
                 it.inputs.property("rootPackage", e.rootPackage ?: "")
                 it.inputs.property("only", e.only.sorted())
                 it.inputs.property("onlyFile", e.onlyFile ?: "")
@@ -380,6 +381,9 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
         ext?.rootPackage?.let {
             args += "--root-package"
             args += it
+        }
+        if (ext?.noRtti == true) {
+            args += "--no-rtti"
         }
         for (spec in requested) {
             args += "--instantiate"
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusExtension.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusExtension.kt
@@ -64,6 +64,16 @@ open class KPlusPlusExtension {
     var cppStandard: String? = null
 
     /**
+     * The target C++ library is built `-fno-rtti` (e.g. v8's monolith) and exports no
+     * `typeinfo` symbols. Forwarded to krapper_gen as `--no-rtti`: the generated generic
+     * `dynamic_cast<D*>` down-cast helpers (`Base.asDerived()`) would reference a missing
+     * `typeinfo for Derived` and fail to LINK, so they are skipped (only RTTI-free LLVM
+     * `classof` down-casts survive). Default false (RTTI-on library keeps the generic
+     * checked down-casts).
+     */
+    var noRtti: Boolean = false
+
+    /**
      * Root Kotlin package for the generated bindings (e.g. "com.acme.app").
      * Forwarded to krapper_gen as `--root-package`; every binding's package
      * becomes `<rootPackage>.<C++ namespace path>` (so `std::vector<int>` lands
--- a/example/kotlin_project/build.gradle.kts
+++ b/example/kotlin_project/build.gradle.kts
@@ -35,6 +35,32 @@ repositories {
     mavenCentral()
 }
 
+// ---- v8 brick 3 (#99): LINK the executable against the system libstdc++ ----
+// PORT of featuregen's #78 recipe (see featuregen/build.gradle.kts). The cpp front-end
+// (kpp.frontend.kotlin_project=cpp) parses v8 against the SYSTEM libstdc++ (gcc-16 here),
+// so the faithful generated wrapper references newer-libstdc++ internals that K/N's
+// BUNDLED gcc-8.3 sysroot (caps at GLIBCXX_3.4.25) cannot resolve. The system libstdc++
+// HAS those symbols and is backward-ABI-compatible, so point K/N's own linker at the REAL
+// system libstdc++.so by ABSOLUTE PATH (NOT -lstdc++ — konan's bundled-sysroot -L is
+// searched first, so `-lstdc++` would shadow to the OLD lib and the newer symbols would
+// stay undefined; an absolute path can't be shadowed by search order). The .so is
+// DISCOVERED at configure time from `clang++ -print-file-name=libstdc++.so` (a GNU-ld
+// linker script/symlink), canonicalized to the real shared object — host-portable, not
+// hardcoded. This is the SAME mechanism clangwalk/cppfrontend/featuregen already use for
+// their klinked binaries; here it is applied to the example's EXECUTABLE binary.
+val systemStdcxxSo: File = run {
+    val proc = ProcessBuilder("clang++", "-print-file-name=libstdc++.so")
+        .redirectErrorStream(true).start()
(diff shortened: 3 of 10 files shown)


monkopedia-reviewer (approved):
## APPROVE — flagship reached, generator-core changes verified default-off/additive

Rigorous main-merge review (this PR touches GENERATOR CORE, not just the example). The mandatory regression gate was RUN, not skipped.

### ⛔ Mandatory regression gate — RAN, GREEN
- `:featuregen:nativeTest -PenableClang` (the 188/0 self-hosted cpp-front-end parity suite — featuregen generated via cppfrontend→ModelIo→krapper_gen, behavioral tests run): **BUILD SUCCESSFUL, 188 tests / 0 failures / 0 errors / 0 skipped.** The generator edits did NOT leak into the default cpp path. (Re-ran from clean after the daemon initially mispicked JVM 8; JDK 21 + committed 6g daemon heap.)
- `:krapper_gen:nativeTest --rerun-tasks`: **281 / 0** (determinism / free-fn-dedup / skip-not-crash suite intact).
- `:krapper_gen:ktlintCheck` + `:krapper_model:ktlintCheck`: green.

### Verdict on the 3 generator-core changes — all truly default-off/additive
1. **bool-underlying enum** (`KotlinWriter.literalForUnderlying`): adds only a `"Boolean" -> (value != 0L).toString()` branch. `underlying = enum.enumUnderlying?.name ?: "Int"`, so `"Boolean"` arises ONLY for a genuinely `: bool` C++ enum (matching the emitted `val value: Boolean` property). The Int/UInt/UShort/UByte paths are untouched; the normal EN-scoped/EN-unscoped Int path is the unchanged `else`. ✔
2. **free-function `removeMethod`** (`IndexedServiceImpl.applyResult`): for a `parent == null` element it now drops from the top-level `classes` var. `allElements` is snapshotted (line 397) before the mapping loop, so the mid-loop reassignment can't perturb iteration; the `!==` identity filter removes only the mapper-selected instance (no over-removal — distinct overloads are distinct instances). The sole default-path behavior change is that a `removeMethod` aimed at a top-level free function — which previously SILENTLY NO-OP'd — now works; the 188/0 parity gate confirms no committed featuregen fixup depended on that silent no-op. Correct bug fix. ✔
3. **`noRtti`** (`KrapperConfig.noRtti` default false → `GenerationContext.noRtti` default false, re-set every run by `reset(rootPackage, noRtti = false)` → `CastTargets`: `if (GenerationContext.noRtti && !useLlvmDynCast) continue`): every other `reset` caller uses the defaulted single-arg form. Default (`noRtti=false`) never takes the new branch, and the LLVM `classof` path is untouched even when set. The DSL knob mirrors the existing `cppStandard`/`referencePolicy` threading (extension KDoc → plugin `--no-rtti` arg + input property → `KrapperConfig` → `App` flag). ✔

### Example / consumer changes
- The 18 `removeMethod` fixups are example-config drops of two genuine latent codegen-bug families (Family 1 out-pointer-param-as-value; Family 2 reference-return of a value-reduced type — the `std::byte`/`<ios>` `_Ios_*` bitmask `*=` operators whose C wrapper is itself UB: returns a pointer to a by-value local). None are on the hello-world path; dropping them for a demo is the right call. Filed as follow-up #103 so they get fixed at the source and the workarounds can later be removed.
- Main.kt adaptations (explicit enum/unique_ptr defaults the cpp front-end doesn't materialize, `_Holder()` zeroed storage, `reference()` vs `_reference()`, `InitializePlatform(platform)` since the factory already `.release()`s) are reasonable cpp-surface consumer changes, not masked generator defects.
- The executable link recipe is a faithful port of featuregen's #78 system-libstdc++ mechanism (absolute `libstdc++.so`, `--gc-sections`, `--allow-shlib-undefined`, `-rpath`/`-rpath-link`, `-lstdc++` removed to avoid bundled-old shadowing).

Tier-1 auto-merge: green review + parity gate 188/0 → merging (squash).

PR #104, opened 2026-06-15
cppfrontend: lock NonCopyable special-member parse determinism (#101)

### LINKED ISSUE #101: cpp front-end: implicit special members of NonCopyable classes materialize parse-dependently (Persistent<v8::Value>)
### Summary
Under the cpp front-end, a `NonCopyable` class materializes its **implicit special members non-deterministically across parses**. Observed on `v8::Persistent<v8::Value>` (NonCopyablePersistentTraits::Copy is a `static_assert(sizeof(S) < 0)`): one parse surfaces the implicit copy ctor (`..._new__const_v8_Persistent_and`), another surfaces the copy-assignment (`..._op_assign`). The two special-member sets differ run-to-run for the same header.

### Why it matters
This violates the front-end determinism/hermeticity guarantee the bootstrap campaign depends on (#11, DeterminismTest standing guard). For NonCopyable types neither implicit special member is actually instantiable, so the divergence is between two equally-wrong member sets — but it is still parse-dependent output.

### Current mitigation (not a fix)
`example/kotlin_project` (PR #100) sidesteps it by dropping BOTH members via `removeMethod` fixups, which is robust regardless of which special member a given parse surfaces. That keeps the example green but masks the underlying non-determinism.

### Suspected area
Implicit special-member materialization in the cpp front-end (`CXXRecordDecl` special-member walk) appears order/parse-dependent rather than canonicalized. Likely needs the implicit-special-member emission to be deterministic (and, for deleted/`static_assert`-trapped copy ops on NonCopyable types, ideally suppressed).

### Acceptance
- Two parses of the same header yield identical special-member sets for NonCopyable types (extend DeterminismTest-style coverage).
- The v8 example no longer needs the `Persistent<v8::Value>` copy ctor/assign `removeMethod` workaround.

Filed as a follow-up from PR #100 (non-blocking).

monkopedia-coder:
## Root cause
`#101` reported that under the cpp front-end a NonCopyable type's special members materialize **parse-dependently** — `v8::Persistent<v8::Value>` surfacing its copy ctor (`..._new__const_v8_Persistent_and`) on one parse and its copy assignment (`..._op_assign`) on another. The hypothesised mechanism is **lazy template-member instantiation**: Clang instantiates a specialization's member functions on first odr-use, so an implicitly-instantiated specialization's `decls()` carries only what *that* parse happened to trigger — a parse-dependent set.

## Finding (verified empirically)
The cpp front-end is **deterministic by construction** and is *not* affected by lazy instantiation, because it never walks the specialization:

- `ModelBuilder.addContextDecls` **SKIPS** `ClassTemplateSpecialization`. A NonCopyable type's members ride the template **PATTERN** (fixed source-order `decls()`); the resolver materializes `Persistent<v8::Value>` by deterministic param substitution (`Parsing.typedAs`).
- Proven with the actual binary: the full v8 base model is **byte-identical across 10 parses** (19346 named elements, 1.26 MB), and **two separate `krapper_gen` processes** emit byte-identical `v8.h`/`v8.cc` with **both** copy ops present — never "one or the other".

## Why suppression (option a) is not feasible
The un-instantiability is a **body-only `static_assert`**, invisible to every copy-constructibility AST query: the copy ctor/assignment are *user-declared* and not deleted (`isDeleted()` is false, `is_copy_constructible` is true — the trap only fires if the body is instantiated). The front-end therefore cannot faithfully detect/suppress them; it can only emit them **deterministically**, and the consumer drops the un-compilable pair by `uniqueCName` (the example's two `removeMethod` fixups, now provably both-needed every run).

## Changes
- **`cppfrontend --noncopyable-determinism`** mode + gated **`noncopyableDeterminismCheck`** task: parses a self-contained v8::Persistent-shaped fixture (deleted-copy base + `static_assert` traits + user-declared copy ctor/assign + a `ForceNC` instantiation) **8×** and asserts the model is byte-identical *and* the NC pattern carries **both** special members on every parse. The standing regression lock for `#101` (a future change that started walking the lazy specialization fails here).
- **`ModelBuilder`**: cite `#101` at the specialization-skip — the skip **is** the determinism guarantee.
- **example**: correct the now-inaccurate Persistent fixup comment (both members are user-declared, emitted deterministically, both always present — not implicit/parse-dependent). The `removeMethod` workaround is **kept** (the members are genuinely un-compilable; suppression isn't AST-detectable).

## Verification
- `noncopyableDeterminismCheck`: **PASSED** (8 parses byte-identical; `(copyCtor, copyAssign)` = `(true, true)` every parse)
- `:featuregen:nativeTest -PenableClang`: **188/0**
- `:krapper_gen:nativeTest`: **281/0**
- `:cppfrontend:runReleaseExecutableKlinker`: ALL SELF-CHECKS PASSED

No model-output change (additive mode + comments), so the cppfrontend parity ledger is unaffected.

Closes #101.

### DIFF
--- a/cppfrontend/build.gradle.kts
+++ b/cppfrontend/build.gradle.kts
@@ -247,6 +247,17 @@ val goldenEmit = tasks.register<Exec>("goldenEmit") {
     commandLine(cppfrontendBinary.absolutePath, "--golden-emit", goldenDir.absolutePath)
 }
 
+// ---- #101: NONCOPYABLE special-member determinism guard ----
+// Runs the cppfrontend binary's `--noncopyable-determinism` mode, which parses a
+// self-contained reduction of v8::Persistent's NonCopyable shape N times (separate
+// ASTContexts) and asserts the special-member set is byte-identical and complete across
+// parses — the standing regression lock for the Persistent<v8::Value> parse-dependent
+// copy-member report (a non-zero exit fails the task). Gated with the module (-PenableClang).
+tasks.register<Exec>("noncopyableDeterminismCheck") {
+    dependsOn("linkReleaseExecutableKlinker")
+    commandLine(cppfrontendBinary.absolutePath, "--noncopyable-determinism")
+}
+
 // ---- #45 brick 2: THE HANDOFF (Phase C) ----
 // The first time generated-bindings-parsed C++ flows through krapper_gen's REAL pipeline:
 //   goldenEmit       — the cppfrontend binary parses the fixture with the Clang C++ AST
--- a/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
+++ b/cppfrontend/src/nativeMain/kotlin/CppFrontend.kt
@@ -24,6 +24,7 @@ import com.monkopedia.krapper.generator.model.WrappedElement
 import com.monkopedia.krapper.generator.model.WrappedField
 import com.monkopedia.krapper.generator.model.WrappedMethod
 import com.monkopedia.krapper.generator.model.WrappedNamespace
+import com.monkopedia.krapper.generator.model.WrappedTU
 import com.monkopedia.krapper.generator.model.WrappedTemplate
 import com.monkopedia.krapper.generator.model.WrappedTypedef
 import com.monkopedia.krapper.generator.model.serialized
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (approved):
## APPROVE — determinism finding is sound, guard is the right outcome (#101)

I scrutinized the central question: **does the new guard cover the same path brick-3 saw differing — the FORCING/instantiation path, not just a base parse?** It does, and "deterministic by construction" holds up.

### The load-bearing finding: the check exercises the production forcing mechanism
`ForcingHeader.contentFor` (krapper_model) synthesizes `struct KrapperForce_<target> { <target> value; };` to force an instantiation. The new fixture's `struct ForceNC { NC<Value> value; };` is the **byte-for-byte same mechanism** — an implicit instantiation of the NonCopyable specialization via a forcing-struct member. So the guard is not a "plain base parse"; it reproduces exactly how production forces a specialization.

Crucially, in **both** the base parse and the per-spec forcing parse, the model is built through the same seam — `parseToWrappedTU` → `buildWrappedTU` → `ModelBuilder.addContextDecls`, which unconditionally **SKIPS `ClassTemplateSpecialization`** (the skip predates this PR — added in #53/#64, i.e. it was already in place when brick-3 observed the symptom). The lazily-instantiated specialization's `decls()` — the only thing that could carry a parse-dependent "one-or-the-other" set — never enters the model on any path. The NC members ride the template **PATTERN**'s fixed source-order `decls()`, and the resolver substitutes deterministically. The fixture deliberately includes the lazy-prone implicit instantiation and proves 8 parses byte-identical with **both** copy ctor and copy assignment present every time.

### Reconciling the contradiction (it is NOT option-c untested-path)
The `static_assert(sizeof<0)` is a **body-only** trap, invisible to AST copy-constructibility queries. The front-end model always contains BOTH members. brick-3's "copy ctor one parse, copy-assign another" is most plausibly the **C++ wrapper-compile** error naming whichever trapped copy-member *body* the compiler instantiated first when compiling the generated wrapper — a downstream compile-time lazy-instantiation artifact, mis-attributed to front-end model non-determinism. It cannot be an untested model path because (1) the only mechanism yielding "one-or-the-other" is walking the specialization's lazy decls(), which is skipped in *every* parse mode; (2) production forcing uses the identical implicit-instantiation shape the guard tests; (3) v8's `Persistent<v8::Value>` needs no `instantiate()` spec (per #102: "the methods were all present"), so it rides the base parse the guard directly models.

### Closes #101 is justified
- Acceptance (a) "two parses yield identical special-member sets for NonCopyable types" — **PROVEN** + standing regression lock.
- Acceptance (b) "example no longer needs the removeMethod workaround" — correctly reclassified as infeasible: the un-instantiability is AST-invisible (body-only assert), so suppression cannot be faithful; dropping both by uniqueCName is the correct permanent fix, kept and re-documented. That bullet was premised on the symptom being real fixable non-determinism; the PR proves it is not. The issue's underlying worry is resolved.

### Verification (re-ran per the cppfrontend standing rule)
- `:cppfrontend:noncopyableDeterminismCheck -PenableClang`: **PASS** — 8 parses byte-identical, `(copyCtor, copyAssign) = (true,true)` all 8.
- `:featuregen:nativeTest -PenableClang --rerun-tasks`: **188 tests / 0 failures / 0 skipped** (aggregated from the XML reports).
- Diff is purely additive: new `--noncopyable-determinism` branch early-returns before any emit path; ModelBuilder change is comments-only (the `if` is unchanged); example change is comment + the kept removeMethods. No model-output change → parity ledger unaffected. krapper_gen is untouched.

Idiom matches existing patterns (arg-dispatch like `--golden-emit`/`--handoff-emit`, gated task like `goldenEmit`, ModelIo byte-compare like the round-trip oracle, cite-the-behavior comments). Approving and merging.

Issue #107, opened 2026-06-15
[krapper_gen] Local<PolymorphicBase>.reference() returns an empty marker Api interface — base instance methods unreachable


monkopedia-coder:
Surfaced by v8 brick 2 (#99) while exercising a richer V8 program (read a numeric JS result via `Value::Int32Value`).

## Symptom
`Local__Value.reference()` returns `ValueApi?`, but `ValueApi` is an **empty marker interface**:
```kotlin
interface ValueApi { val ptr: COpaquePointer }
```
All of `Value`'s real methods (`Int32Value`, `BooleanValue`, `NumberValue`, `IsString`, `ToString`, …) live on the concrete `Value` class, which `implements ValueApi`. So a dereferenced `Local<Value>` exposes **none** of them — `numResult.reference()?.Int32Value(ctx)` fails with `Unresolved reference 'Int32Value'`.

By contrast a NON-polymorphic type works: `Local__Script.reference()` returns the concrete `Script`, so `script.reference()?.Run(ctx)` resolves (this is why the brick-3 hello-world compiled). The divergence is that a polymorphic base (a class WITH subclasses, like `v8::Value`) gets the `<Base>Api` marker-interface treatment and `reference()` returns the bare marker.

## Why it matters
For v8 essentially every value handed back from JS is a `v8::Value` (or a subclass up-cast to it). If you cannot call any instance method on a dereferenced `Local<Value>`, the bindings are barely usable past the hello-world. This is a general polymorphic-base usability gap, not v8-specific.

## Note: the object IS already the right type
`reference()` constructs a real `Value` internally and merely up-casts it to the marker:
```kotlin
inline fun reference(): ValueApi? =
    Value((v8_Local_v8_Value_op_reference(ptr) ?: return null), memScope)
```
So a consumer-side `as? Value` downcast succeeds at runtime — that's the current workaround (used in the example, brick 2). The fix is to make this unnecessary.

## Suggested fix directions (design call)
1. Have `Local<Base>.reference()` return the **concrete** `Base` class (not the marker) when a same-named concrete class exists — matching the non-polymorphic case. Cleanest for consumers.
2. OR promote the base class's instance methods onto the `<Base>Api` interface so the marker is no longer empty.
Either is a non-trivial generator change with featuregen-parity implications, so filing for a scoped fix rather than tacking onto the brick.

## Acceptance
[...]

PR #108, opened 2026-06-15
example(v8): richer demo — read a numeric JS result via Value::Int32Value (#99 brick 2)

monkopedia-coder:
## v8 brick 2 (#99) — grow the proven v8 surface

The brick-3 hello-world proved the self-hosted v8 bindings compile+link+run a string concat. This brick extends the demo to a **richer operation**: evaluate `"40 + 2"` and read the result back as a native 32-bit `Int`.

```
Got: Hello, Worldy!
Computed: 42
```

### What changed (small, bounded)
- **`build.gradle.kts`:** `instantiate("v8::Maybe<int>")` — forces the `Maybe<int>` model so `v8::Value::Int32Value(context)` (returns `v8::Maybe<int32_t>`, `int32_t == int`) survives the `IGNORE_MISSING` drop. This is the only generation delta. Verified the method now binds: `v8_Value.kt: Int32Value(context): Maybe__Int`.
- **`Main.kt`:** the richer block. `Int32Value(context).FromJust()` yields the unwrapped `Int`.

### The `as? Value` downcast + a filed gap (#107)
`Local<Value>.reference()` returns the **empty marker interface** `ValueApi` (`val ptr` only) — a polymorphic base's real methods (`Int32Value`, `IsString`, `ToString`, …) live on the concrete `Value` class. So a consumer-side `as? Value` downcast is needed to reach them. It always succeeds (reference() builds a real `Value` under the hood). The underlying usability gap is filed as **#107** for a scoped generator fix; this brick routes around it rather than tacking on a non-trivial generator change.

### Verification
End-to-end green under `-PenableClang`: `kplusplusSync` + `compileKotlinNative` + link vs the 62MB `libv8_monolith.a` + `runDebugExecutableKlinker` — program prints both lines. No krapper_gen change (only one forcing spec + example consumer code), so featuregen parity is unaffected.

refs #99

### DIFF
--- a/example/kotlin_project/build.gradle.kts
+++ b/example/kotlin_project/build.gradle.kts
@@ -142,6 +142,14 @@ kplusplus {
     // (kpp.frontend.kotlin_project=cpp), built with -PenableClang.
     cppStandard = "c++17"
 
+    // v8 brick 2 (#99): force the v8::Maybe<int> model so Value::Int32Value(context) — whose
+    // return is v8::Maybe<int32_t> (int32_t == int) — survives codegen instead of being
+    // dropped (IGNORE_MISSING) for want of a forcing model. Unblocks the richer number→int
+    // demo in Main.kt (`numResult.Int32Value(context).FromJust()`). The Maybe<int> member
+    // set (FromJust/ToChecked/IsJust/…) is the primary Maybe<T> projection — no void-spec
+    // workarounds apply here.
+    instantiate("v8::Maybe<int>")
+
     // v8 brick 3 (#99): the v8 monolith is built `-fno-rtti`, so it exports no `typeinfo`
     // symbols. Without this, the generated generic `dynamic_cast<D*>` down-cast helpers
     // (e.g. TracingController.asTracingController(), ExternalStringResourceBase.as*()) emit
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
APPROVE — tier-1 auto-merge.

Verified example-only: the diff touches exactly two files, `example/kotlin_project/build.gradle.kts` (+8) and `example/kotlin_project/src/nativeMain/kotlin/Main.kt` (+27). Nothing under `krapper_gen/`, `cppfrontend/`, `compiler/`, or `plugin/` — so the generator surface is unchanged and the featuregen 188/0 gate is not implicated. The `instantiate("v8::Maybe<int>")` forcing spec affects only this example's own generation.

Re-built and RAN the example from a PR-108 worktree (not just compiled):
`JAVA_HOME=/usr/lib/jvm/java-17-openjdk ./gradlew runDebugExecutableKlinker -PenableClang` (cpp front-end). BUILD SUCCESSFUL; program output:
```
Got: Hello, Worldy!
Computed: 42
```
The post-build local-extraction refactor (numValue + computed) compiles and runs correctly — the `Maybe<int>` forcing spec lets `Value::Int32Value(context) -> Maybe<int> -> FromJust()` survive codegen.

`as? Value` workaround: reasonable and traceable. #107 is filed and accurately describes the polymorphic-base gap (`Local<Value>.reference()` returns the empty `ValueApi` marker; `reference()` constructs a real `Value` boxed as the marker, so the downcast always succeeds at runtime). The Main.kt comment cites #107, so the clean fix is correctly deferred, not lost.

Forcing spec is minimal (single `v8::Maybe<int>`, well-commented, no over-seeding). Idiom matches the file's heavily-commented style; max line length 98 (≤100). Minor non-blocking nit: imports in this file are not alphabetically sorted pre-existing (`std.*` follows all `v8.*`), so the `v8.Value` placement is consistent with the file's un-enforced ordering — the example module isn't ktlint-gated.

refs #99 stays open (v8 roadmap umbrella).

PR #109, opened 2026-06-15
clangwalk: self-checking assertion test for the AST walk (#8 part 2)

### LINKED ISSUE #8: [clang-campaign] Make the stage1 walk representative (methods/bases at runtime) + a real clangwalk test

monkopedia-coder:
## #8 part 2 — give clangwalk a real assertion test

**Part 1 was already done (PR #42):** the walk already descends into a `CXXRecordDecl`'s `bases()` AND `methods()` at runtime. The only remaining work was part 2: replace the `println`-only walk with a REAL test that ASSERTS the expected decls/kinds/names.

### What changed (pure consumer Kotlin, `clangwalk/src/nativeMain/kotlin/AstWalk.kt` only)
- `main()` now COLLECTS the walk into a `List<WalkedDecl(kind, name, typeStr, bases, methods)>` instead of only printing.
- A local `check(name, pass, detail)` helper mirrors `cppfrontend`'s self-check pattern: records each assertion, prints `PASS`/`FAIL`, and `main()` calls `exitProcess(1)` if any failed — so `:clangwalk:runReleaseExecutableKlinker` is now self-verifying (exit 0 = pass, non-zero = build fails).
- Seed bindings unchanged — `methods()`/`bases()`/`getReturnType()`/`getAsString()` already bind in the committed seed (the walk already called them). No regeneration.

### Assertions added (16), all against the EXACT Clang oracle captured by running the walk
- **count == 10** top-level decls, and the ordered `(kind,name)` sequence matches — including the **5 implicit compiler-builtin typedefs** Clang injects: `__int128_t`, `__uint128_t`, `__NSConstantString`, `__builtin_ms_va_list`, `__builtin_va_list`.
- **Point** — kind `CXXRecord`, **0 bases**, **0 methods** (fields-only; fields aren't walked).
- **Shape** — kind `CXXRecord`, 0 bases, methods == `{area:int, name:const char *, scale:double}` (exact return-type spellings).
- **Circle** — kind `CXXRecord`, bases == `{Shape}`, methods == `{radius:double}`.
- **freeFunction** — kind `Function`, type spelling == `void (int)`.
- **globalVar** — kind `Var`, type spelling == `int`.

### Exact Clang `getAsString()` / `getDeclKindName()` spellings observed (the oracle)
- bases: `Shape`
- return types: `int`, `const char *`, `double`
- function type: `void (int)`; var type: `int`
- builtin typedef kind: `Typedef`; record kind: `CXXRecord`; function kind: `Function`; var kind: `Var`

### Verification
- `JAVA_HOME=/usr/lib/jvm/java-21-openjdk ./gradlew :clangwalk:runReleaseExecutableKlinker -PenableClang` → all 16 self-checks PASS, **BUILD SUCCESSFUL, exit 0**.
- **Teeth-check (the test has teeth):** temporarily flipped the `globalVar` expected spelling `int`→`long`, rebuilt+ran → `[FAIL] globalVar type spelling`, `clangwalk: 1 self-check(s) FAILED`, process exited **non-zero 1**, **BUILD FAILED**. Then reverted and re-verified green.

Only `clangwalk/` changed (`git diff --name-only` = one file).

Closes #8.

### DIFF
--- a/clangwalk/src/nativeMain/kotlin/AstWalk.kt
+++ b/clangwalk/src/nativeMain/kotlin/AstWalk.kt
@@ -15,6 +15,29 @@
  */
 import clang.tooling.buildASTFromCode
 import kotlinx.cinterop.memScoped
+import kotlin.system.exitProcess
+
+// One walked top-level declaration, collected from the AST so the walk can be ASSERTED as a
+// whole rather than printed. methods is (name -> return-type-spelling); bases is the list of
+// base-class spellings (CXXBaseSpecifier::getType().getAsString()). For non-records both are
+// empty; for a record both come from the SAME owning class (#40's cumulative-forcing fix).
+private data class WalkedDecl(
+    val kind: String,
+    val name: String,
+    val typeStr: String,
+    val bases: List<String>,
+    val methods: List<Pair<String, String>>,
+)
+
+// Mirror of cppfrontend's self-check helper (CppFrontend.kt:check): record every assertion,
+// print PASS/FAIL with a message, and let main() exit non-zero if ANY failed — so
+// `:clangwalk:runReleaseExecutableKlinker` is self-verifying instead of a println demo.
+private var failures = 0
+
+private fun check(name: String, pass: Boolean, detail: String = "") {
+    if (!pass) failures++
+    println("  [${if (pass) "PASS" else "FAIL"}] $name${if (detail.isEmpty()) "" else " — $detail"}")
+}
 
 // Stage1 self-bootstrap: parse C++ into a real Clang AST and walk it — entirely on the
 // kplusplus-generated bindings of Clang's own C++ AST API (no libclang). This is the proof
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
APPROVED (tier-1 auto-merge).

Consumer-Kotlin-only verified: `gh pr diff 109 --name-only` shows ONLY clangwalk/src/nativeMain/kotlin/AstWalk.kt — nothing under krapper_gen/cppfrontend/compiler/seed bindings, so no featuregen gate needed.

Re-ran the self-checking executable on the PR head (wt-8 worktree, JDK 21):
  ./gradlew :clangwalk:runReleaseExecutableKlinker -PenableClang
Result: all 15 self-checks PASS, BUILD SUCCESSFUL, exit 0. The runReleaseExecutableKlinker task re-executed and printed the live assertion lines that exist only in this PR (e.g. 'Shape methods == {area:int, name:const char *, scale:double}', 'top-level (kind,name) sequence matches the parsed TU' over the 5 builtin typedefs + Point/Shape/Circle/freeFunction/globalVar). Spellings match Clang's oracle exactly: const char *, void (int), base Shape.

Teeth confirmed by inspection: module-level `failures` var, `check` increments it on any false assertion, and `exitProcess(1)` fires unconditionally when failures>0 before the success println — failures genuinely reach the non-zero exit and are not swallowed. The int->long flip the coder reported is consistent with this wiring.

Idiom: keeps the per-line campaign-primitive commentary ethos. Minor non-blocking nit: lines 39 (101) and 192 (110) exceed the 100-char guideline — clangwalk has no ktlint gate so not enforced; tidy opportunistically.

Closes #8 part 2.

PR #112, opened 2026-06-18
cleanup: unify samples under samples/ + add a minimal newcomer sample

monkopedia-coder:
Part of the repo-cleanup effort. Unifies the examples under `samples/` and adds a tiny, monolith-free first-contact sample so a newcomer isn't dropped straight into the 62 MB V8 demo.

## Part 1 — move the V8 example to `samples/v8` (flattened)
`example/kotlin_project` + `example/include` + `example/libv8_monolith.a` → `samples/v8/{…, include/, libv8_monolith.a}` (all via `git mv`, tracked as renames; `example/` removed).

Path/identity fixes for the flatten:
- `build.gradle.kts`: `../include/v8-combined.h` → `include/v8-combined.h`, `../include/` → `include/`, `../libv8_monolith.a` → `libv8_monolith.a`.
- `settings.gradle.kts`: `rootProject.name = "kotlin_project"` → `"v8"`; `includeBuild("../../")` unchanged (`samples/v8` is two levels below the repo root, so `../../` is still the root).
- `gradle.properties`: `kpp.frontend.kotlin_project=cpp` → `kpp.frontend.v8=cpp` (the key is the rootProject.name); comments in `build.gradle.kts` updated to match.
- Updated `README.md`, `docs/ARCHITECTURE.md`, and a stale path comment in `krapper_gen/.../FixupApplier.kt`.

**Verified:** `cd samples/v8 && ./gradlew kplusplusSync -PenableClang` → **BUILD SUCCESSFUL**, cpp front-end generated 508 Kotlin files (reads the flattened `samples/v8/include/` correctly). The move + path fixes are sound.

## Part 2 — new `samples/minimal`
A standalone k++ sample binding a hand-written C++ geometry library — no STL/template surface, no monolith:
- `cpp/geometry.h` / `cpp/geometry.cc`: `geo::Point` (x, y, `distanceTo(const Point&)`) and `geo::Rect` (origin, width, height, `area()`, `center()` by value). Plain fields + by-value/const-ref methods.
- A `compileGeometryLib` Gradle `Exec` task compiles `geometry.cc` → `build/cpplib/libgeometry.a` (`clang++ -std=c++17 -c -fPIC` then `ar rcs`); `kplusplusSync` is wired `dependsOn(compileGeometryLib)` so the archive exists before generation.
[...]

PR #110, opened 2026-06-17
cleanup: delete dead/superseded modules (v1 plugin, testlib trio, slice, classes.py)

monkopedia-coder:
## Repo cleanup — part 1: deletions

Owner-approved aggressive removal of modules with no future/users (git history is the backstop). First of a multi-PR cleanup; **renames/restructure + docs land in follow-ups**.

### Removed
| Path | Why |
|------|-----|
| `plugin/` | LEGACY **v1** Gradle plugin (`com.monkopedia.kplusplus.plugin`) — superseded by the v2 `compiler/` plugin; nothing internal applied it |
| `.github/workflows/publish.yaml` | the only CI workflow, and it only published the v1 plugin — re-established against v2 at a real release |
| `testlib/`, `testlib_kotlin/`, `testlib_kotlin_manual/` | orphaned manual-vs-generated fixtures (settings-only, no test aggregation, no CI) — superseded by featuregen/feature-tests/cppfixture |
| `slice/` | redundant — its only purpose (configurable `rootPackage`) is covered by `RootPackageTest` + `DeterminismTest` |
| `classes.py` | dead pre-flip libclang Python tool (`import clang.cindex`) |

### Wiring
Dropped the 4 includes (+ the gated `slice`) from `settings.gradle.kts`, removed the `kpp.frontend.slice` property, updated comments that named removed modules. No renames, no logic changes.

### Verification
- `settings.gradle.kts` configures cleanly (no dangling project refs).
- `:krapper_gen:nativeTest` unaffected.
- `:featuregen:nativeTest -PenableClang` → **188/0** (the self-hosted suite — removing the modules didn't disturb the v2 path).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #113, opened 2026-06-18
cleanup: drop redundant v8 binary + fix stale v1-plugin comment (part 4)

monkopedia-coder:
## Repo cleanup — part 4: small leftovers

- **`samples/v8/include/libv8_monolith.a`** — a 55 MB **unreferenced** duplicate of the 62 MB `samples/v8/libv8_monolith.a` that the build actually links (`library("libv8_monolith.a")`). Carried along when the v8 example moved in #112; nothing reads the `include/` copy.
- **`KPlusPlusExtension.kt`** — corrected a now-false doc comment that pointed users to "drop back to the legacy `com.monkopedia.kplusplus.plugin` (still maintained for backwards compatibility)". The v1 plugin was **deleted** in #110, so that escape hatch no longer exists; the comment now just points at extending the `Fixup` sealed type.

Comment + dead-binary removal only; no logic changes.

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusExtension.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusExtension.kt
@@ -31,11 +31,10 @@ import org.gradle.api.Action
  *      forwarded to krapper_gen via --fixup-file.
  *   3. (future) Per-spec compiler options, package overrides, etc.
  *
- * This block intentionally does not re-introduce the v1 mapping DSL. If
- * you need arbitrary FIR introspection or per-method closures, that's a
- * signal to either (a) extend the [Fixup] sealed type with a new narrow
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
APPROVED (tier-1). Verified:
- include/libv8_monolith.a is genuinely unreferenced — the only live link is `library("libv8_monolith.a")` (build.gradle.kts:139), which resolves project-root-relative to the intact root copy samples/v8/libv8_monolith.a (62 MB, still present). The include/v8.def `staticLibraries` mention is in a def file nothing in the build consumes (no cinterop/definitionFile reference), and cinterop staticLibraries resolves against libraryPaths, not the def dir — dead either way.
- KPlusPlusExtension.kt change is comment-only (inside the doc block; no code/logic) and accurate: the v1 com.monkopedia.kplusplus.plugin was removed in #110, so the new wording correctly points at extending the Fixup sealed type.
- Diff is exactly those two items, nothing else. No build gate needed (comment + dead-asset deletion).

PR #114, opened 2026-06-20
krapper_gen: Local<PolymorphicBase>.reference() returns the concrete base, not the empty Api marker (#107)

### LINKED ISSUE #107: [krapper_gen] Local<PolymorphicBase>.reference() returns an empty marker Api interface — base instance methods unreachable

monkopedia-coder:
## Root cause
A class used as a base gets a generated `interface <Name>Api`. Its method surface (`baseInterfaceMethods` in `KotlinWriter.kt`) includes **only VIRTUAL instance methods** — the interface exists for polymorphic dispatch. When a polymorphic base's instance methods are all **non-virtual** (e.g. `v8::Value`'s `Int32Value`/`IsString`), `<Name>Api` collapses to an **empty marker**: `interface ValueApi { val ptr: COpaquePointer }`.

Separately, the dereference accessors `reference()` / `pointer_reference()` (C++ `operator*` / `operator->` on a `Local<T>` / pointer wrapper) had their **declared return type widened to that marker** via the base→interface remap (`remapReturnType`). So a dereferenced `Local<Value>` was typed `ValueApi?` and exposed none of the concrete methods — a manual `as? Value` downcast was required.

## Fix
Keep the deref-accessor return **concrete**, mirroring the existing index-get carve-out in `generateBasicMethod`. The body already constructs the concrete `Value(ptr, memScope)`, so the declared type just matches what is actually returned. Genuine polymorphic **method positions** (a method that returns/takes a base, so a subclass can substitute) still remap — only the `reference()`/`pointer_reference()` deref of a wrapper changed.

`KotlinWriter.generateBasicMethod` now keeps `method.returnType` when `isDerefAccessor(method)` (operator is `REFERENCE` or `POINTER_REFERENCE`), alongside the pre-existing `IND`/`keepConcreteElement` cases.

## Test
New `PolymorphicReferenceTest` (krapper_gen nativeTest): a polymorphic base with a non-virtual instance method + a subclass + an `operator*` wrapper. Asserts both halves — `BaseApi` is an empty marker (root cause) **and** the wrapper's `reference()` returns the concrete `Base` (not `BaseApi`), so the non-virtual method is reachable.

## Verification
- **featuregen cpp-parity gate: 188/0** (`:featuregen:nativeTest -PenableClang`). The `Ih*` polymorphic/cast tests are unchanged — no abstract-base refinement was needed (`IhAbstractTest` still green), because the change only touches the deref-accessor declared return, not method positions or the cast machinery.
- **krapper_gen nativeTest: 288/0** (incl. the new test).
- **samples/v8** drops the `as? Value` downcast (and the #107 comment): `numResult?.reference()?.Int32Value(context)?.FromJust()` now resolves directly. The sample still runs and prints `Got: Hello, Worldy!` and `Computed: 42`.

Closes #107

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/codegen/KotlinWriter.kt
@@ -1714,6 +1714,14 @@ class KotlinWriter(private val pkg: String, policy: CodeGenerationPolicy = Throw
         // `fun in(...)` (won't parse). Back-tick-escape it (mirrors the param/field path).
         .escapeKotlinKeyword()
 
+    // The C++ dereference operators `operator*` ([ResolvedOperator.REFERENCE]) and
+    // `operator->` ([ResolvedOperator.POINTER_REFERENCE]) on a `Local<T>`/pointer wrapper.
+    // Their declared return must stay CONCRETE (not the base-interface remap) — see #107 and
+    // the retType comment in [generateBasicMethod].
+    private fun isDerefAccessor(method: ResolvedMethod): Boolean =
+        method.operator == ResolvedOperator.REFERENCE ||
+            method.operator == ResolvedOperator.POINTER_REFERENCE
+
     private fun KotlinCodeBuilder.generateBasicMethod(
         method: ResolvedMethod,
         uniqueCName: Symbol,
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
APPROVE — #107 fix verified, both mandatory gates green, test has teeth.

**Fix narrowness (verified against the diff + surrounding code).** `isDerefAccessor` matches ONLY `ResolvedOperator.REFERENCE`/`POINTER_REFERENCE` — i.e. C++ `operator*`/`operator->`, both unary no-arg, mapping to `BasicMethod("reference")`/`BasicMethod("pointer_reference")`. It is defined once and used in exactly ONE place: the `retType` ternary in `generateBasicMethod`. The arg path (`remapArgType`), the base-interface method surface (`baseInterfaceMethods`/`renderInterfaceType`), and `remapReturnType` for every non-deref method position are all UNTOUCHED — so genuine polymorphic method positions (a method returning/taking a base, for subclass substitution) still remap to `<Base>Api`. Only the deref-accessor *declared* return changed. Binary `*` (multiply) is a different operator (`TIMES`, arity 1), so there is no collision.

**Correctness.** The body always builds the concrete wrapper: `defineArgsAndBody(returnType = method.returnType, ...)` is passed the concrete type independent of `retType`, so the now-concrete declared type matches what is actually constructed — no lie. The abstract-base case is safe too: the generated Kotlin wrapper just holds a `ptr` (it is never an abstract Kotlin class), so wrapping a constructed pointer is fine even when the C++ base is abstract — and `IhAbstractTest` is green without refinement because the change touches no method positions.

**Test has teeth — PROVEN.** Reverted just the `isDerefAccessor(method) ||` line and re-ran `PolymorphicReferenceTest` → it FAILS with AssertionError (the `assertFalse("BaseApi" in localText)` half is the discriminating assertion; on the unfixed path `reference()` widens to `BaseApi`). Restored to PR HEAD, clean. The test drives the real resolve+codegen path and pins both halves (empty `BaseApi` marker precondition + concrete `Base` return).

**Mandatory regression gates (RAN):**
- `:featuregen:nativeTest -PenableClang` → **188/0**, all 7 Ih* suites green (Abstract 3/0, Cast 4/0, DownCast 2/0, Flatten 4/0, MultiLevel 3/0, NvDtor 2/0, Upcast 4/0). The cpp-front-end sync regenerated 117 Kotlin files this run and they were byte-identical to the prior build (downstream compile/link/test UP-TO-DATE) — confirming the fix is a no-op on featuregen's generated surface = zero regression surface.
- `:krapper_gen:nativeTest` → **288/0**, incl. the new PolymorphicReferenceTest.

**Sample.** `samples/v8/Main.kt` drops the `import v8.Value` and the `as? Value` downcast; the call is now `numResult?.reference()?.Int32Value(context)?.FromJust()`. Verified by inspection (coder ran it: "Computed: 42").

Closes #107.

PR #115, opened 2026-06-20
cleanup: drop the dead v1-era v8.def from samples/v8

monkopedia-coder:
`samples/v8/include/v8.def` is a leftover hand-written cinterop `.def` from the v1 plugin era (it lists the v8 headers + `staticLibraries = libv8_monolith.a`). The v2 compiler plugin generates its own `.def` during `kplusplusSync`; **nothing** in the build (or anywhere outside `build/`) references this file — confirmed via `grep`. Flagged during the #113 review; this is the last sliver of the samples cleanup.

Dead-file removal only.


monkopedia-reviewer (approved):
Verified: diff is solely the deletion of `samples/v8/include/v8.def` (the v1-era hand-written cinterop def). Confirmed genuinely dead — zero references to `v8.def` anywhere in-tree (kt/kts and full grep, excluding build/krapped), and `samples/v8/build.gradle.kts` uses the v2 `kplusplus { }` compiler subplugin with no `cinterops`/`definitionFile`/`.def`. The v8 sample binds via the v2 plugin (kpp.frontend.v8=cpp), so removing this unreferenced file cannot affect the build. APPROVE + squash.

Issue #117, opened 2026-07-08
Duplication: fir/KPlusPlusDiagnostics.kt is a byte-identical copy in compiler/plugin and compiler/native


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** duplication
**Where:** `compiler/plugin/src/main/kotlin/com/monkopedia/kplusplus/compiler/fir/KPlusPlusDiagnostics.kt` and `compiler/native/src/main/kotlin/com/monkopedia/kplusplus/compiler/fir/KPlusPlusDiagnostics.kt`
**Current:** The Kotlin compiler plugin ships two build targets (`compiler/plugin` = embeddable/JVM, `compiler/native` = K/N host). Most of the paired FIR/registrar files legitimately DIVERGE per target (FirKPlusPlusRegistrar, KPlusPlusComponentRegistrar, KPlusPlusFirCheckersComponent all differ, and several files exist in only one tree — that split is real and correct). But `fir/KPlusPlusDiagnostics.kt` (30 lines: the `SYNC_REQUIRED` diagnostic factory + its renderer map) is **byte-for-byte identical** between the two modules — verified with `diff -q`, no differences.
**Proposed:** Make the two modules share this one source rather than keep two copies. Point one modules Kotlin `sourceSets` at the others directory via `kotlin.srcDir(...)`, or extract the shared diagnostic into a tiny common source dir both `include`. (The other paired files stay per-target — only the identical one is shared.)
**Why:** A byte-identical copy across two modules drifts silently: edit the diagnostic key/renderer in one tree and the other keeps the old text with no compile error, so the plugin and native hosts would emit mismatched diagnostics. Sharing the single identical file removes that drift risk without touching the genuinely target-specific files.
**Effort:** small refactor
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage→work). To decline: close it — it won't be suggested again._

monkopedia-coder:
Closing on implementation — this is a **false positive**, no change warranted.

[...]

PR #119, opened 2026-07-08
code-health: dedup InstantiationRequest spec (#116) + looping-template guard idiom (#118)

### LINKED ISSUE #116: Duplication: InstantiationRequest canonical spec string is reconstructed by hand in 3 places (must stay in lockstep)
## Code-health suggestion (proactive — staged for your review)
**Category:** duplication
**Where:** `krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt:205` and `:208`, `krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt:135`; the type is `InstantiationRequest` in `krapper_gen/src/commonMain/kotlin/KrapperService.kt:33`.
**Current:** The canonical instantiation-spec string is rebuilt inline from an `InstantiationRequest` in three spots, all with the exact same interpolation:
```kotlin
// App.kt:205 — building the forcing-model key map
val req = parseInstantiation(entry.substring(0, idx))
"${req.base}<${req.args.joinToString(", ")}>" to entry.substring(idx + 1)
// App.kt:208 — checking every --instantiate has a matching --forcingModel
.map { "${it.base}<${it.args.joinToString(", ")}>" }
// IndexedServiceImpl.kt:135 — the target string requestInstantiation forces on
val target = "${req.base}<${req.args.joinToString(", ")}>"
```
The comment at App.kt:198-199 spells out the coupling explicitly: the forcing-model key "matches requestInstantiations target string regardless of arg spacing" — i.e. these three literals MUST format `args` identically (same `", "` separator) or the `--forcingModel` lookup silently misses and the instantiation is rejected / unforced.
**Proposed:** Give `InstantiationRequest` a single canonical rendering and call it everywhere:
```kotlin
data class InstantiationRequest(val base: String, val args: List<String>) {
    val spec: String get() = "$base<${args.joinToString(", ")}>"
}
```
Then App.kt / IndexedServiceImpl.kt use `req.spec` instead of re-interpolating.
**Why:** Removes a hand-maintained correctness coupling across two files — today a change to how args are joined (spacing, ordering) in one place silently breaks forcing-model matching, with no compile error to catch it. One source of truth on the data class makes the canonical form authoritative and self-documenting.
**Effort:** mechanical
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage→work). To decline: close it — it won't be suggested again._

### LINKED ISSUE #118: Idiom: WrappedTemplateType.init constructs an exception only to printStackTrace() it, then continues with the degenerate value
## Code-health suggestion (proactive — staged for your review)
**Category:** idiom
**Where:** `krapper_model/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/type/WrappedTemplateType.kt:27-31`
**Current:**
```kotlin
init {
    if (baseType.toString().endsWith("<${templateArgs.joinToString(", ")}>")) {
        IllegalArgumentException("Looping templates $baseType $templateArgs").printStackTrace()
    }
}
```
This is a "should never happen" guard for a self-referential (looping) template, but it: (1) constructs an `IllegalArgumentException` purely to use it as a logging vehicle — it is never thrown; (2) writes a full stack trace to stderr during an otherwise-normal generator run; and (3) then silently continues, constructing and returning the known-degenerate looping `WrappedTemplateType`. It is the only `printStackTrace()` in the human-authored source, and krapper_gen already has a proper logging seam (`generator/Logging.kt` / `Logger`).
**Proposed:** Pick one honest behavior for the guard: either genuinely fail fast (`throw IllegalArgumentException(...)` if a looping template is truly invalid), or if continuing is intended, route the warning through the project logger (or a plain diagnostic string) rather than a thrown-away exception + stack trace. If the guard is dead/never-hit in practice, consider removing it. (Note: `krapper_model` is a lower-level module — confirm whether it can depend on the `Logger` seam or should take a minimal diagnostic hook.)
**Why:** An exception constructed only to `printStackTrace()` is a code smell — it clutters generator output with a stack trace for a non-fatal condition and hides whether the loop is actually a bug (should throw) or a tolerated case (should be a clean warning). Making the intent explicit removes the ambiguity.
**Effort:** mechanical
_Filed by the nightly code-health pass. To act on it: relabel `agent-workable` (it'll flow into triage→work). To decline: close it — it won't be suggested again._

monkopedia-coder:
Two small cleanups from the nightly code-health pass (both approved by the owner to `agent-workable`).

## #116 — duplication: `InstantiationRequest` spec string
The canonical `base<arg, arg>` specialization spelling was reconstructed by hand at three sites that must stay byte-for-byte in lockstep (the forcing-model keys in `App.kt` ×2, and `IndexedServiceImpl.requestInstantiation`) — a silent coupling flagged by the code's own comment. Added a single `spec` computed property on `InstantiationRequest` and routed the three sites through it. It's a getter, not a constructor parameter, so the `@Serializable` shape is unchanged.

## #118 — idiom: `WrappedTemplateType.init` looping guard
The init block constructed an `IllegalArgumentException` only to `printStackTrace()` it (the **only** `printStackTrace` in the tree) and then continued, building a degenerate self-referential type. Replaced with `require(...)`, matching the three sibling `error(...)` invariant guards already in the same class — fail loudly rather than fake a looping template. `krapper_model` has no logger seam, so `require`/`error` is the file-idiomatic choice.

## Verification
- `:krapper_gen:nativeTest` → **288/0**
- `:featuregen:nativeTest -PenableClang` → **188/0** (the mandatory codegen gate)
- The new `require(...)` guard **did not fire** anywhere in the featuregen corpus (grep count 0), so #118 is behavior-neutral in practice — it only removes the java-ism and hardens the impossible-state path.

Closes #116
Closes #118

(Sibling #117 from the same pass was closed as a false positive — the "duplicate" native file is gitignored `syncSource` output, not a committed copy.)

### DIFF
--- a/krapper_gen/src/commonMain/kotlin/KrapperService.kt
+++ b/krapper_gen/src/commonMain/kotlin/KrapperService.kt
@@ -30,7 +30,14 @@ data class IndexRequest(
 )
 
 @Serializable
-data class InstantiationRequest(val base: String, val args: List<String>)
+data class InstantiationRequest(val base: String, val args: List<String>) {
+    /**
+     * The canonical `base<arg, arg>` specialization spelling. This is the string every
+     * consumer keys instantiations on (forcing-model paths, the requested set), so it lives
+     * here to keep those sites in lockstep instead of being hand-rebuilt at each one.
+     */
+    val spec: String get() = "$base<${args.joinToString(", ")}>"
+}
 
 @KsService
 interface KrapperService : RpcBidiService {
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/App.kt
@@ -202,10 +202,10 @@ class KrapperGen : CliktCommand() {
             val idx = entry.indexOf('=')
             require(idx > 0) { "--forcingModel must be '<spec>=<path>', got: $entry" }
             val req = parseInstantiation(entry.substring(0, idx))
-            "${req.base}<${req.args.joinToString(", ")}>" to entry.substring(idx + 1)
+            req.spec to entry.substring(idx + 1)
         }
         val missing = instantiate.map { parseInstantiation(it) }
-            .map { "${it.base}<${it.args.joinToString(", ")}>" }
+            .map { it.spec }
             .filter { it !in cppForcingModelPaths }
         require(missing.isEmpty()) {
             "Every --instantiate spec requires a matching --forcingModel '<spec>=<path>' " +
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -132,7 +132,7 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
     }
 
     override suspend fun requestInstantiation(req: InstantiationRequest) {
-        val target = "${req.base}<${req.args.joinToString(", ")}>"
+        val target = req.spec
         requested.add(target)
         // The forcing struct's name + header content are SHARED with the cpp front-end
         // (ForcingHeader, :krapper_model): the cppfrontend binary synthesizes the same
--- a/krapper_model/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/type/WrappedTemplateType.kt
+++ b/krapper_model/src/nativeMain/kotlin/com/monkopedia/krapper/generator/model/type/WrappedTemplateType.kt
@@ -25,8 +25,10 @@ class WrappedTemplateType(val baseType: WrappedType, val templateArgs: List<Wrap
         }
 
     init {
-        if (baseType.toString().endsWith("<${templateArgs.joinToString(", ")}>")) {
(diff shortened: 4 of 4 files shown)


monkopedia-reviewer (approved):
Clean two-item cleanup. Both changes are sound.

**#116 (spec dedup):** The `spec` property is a computed getter (`val spec: String get() = ...`), not a constructor parameter — the `@Serializable` shape is genuinely unchanged. The expression is identical to what each of the three hand-rebuilt sites produced (`"$base<${args.joinToString(", ")}>"`), so the substitution is byte-for-byte faithful. Verified that all three `InstantiationRequest` call sites that reconstructed the spec string are routed through the new property; no other sites in the tree hand-build the spec from `req.base`/`req.args`.

**#118 (require guard):** Replacing `printStackTrace()`-then-continue with `require(!...)` is the right hardening. The old code built a degenerate self-referential `WrappedTemplateType` and silently continued; `require` matches the three sibling `error(...)` invariant guards already in the same class and is the file-idiomatic choice. The PR confirms the guard did not fire anywhere on the featuregen corpus (grep 0), so the behavior change only affects an impossible-or-malformed-input path. Callers cannot sensibly recover from a looping template — failing loudly is strictly preferable to constructing a broken type that propagates errors downstream.

Verification gate: nativeTest 288/0 + featuregen -PenableClang 188/0 green. No comment-quality or naming issues. Approve.

Issue #123, opened 2026-07-08
krapper_gen: skip methods whose signature references an inaccessible (protected/private) type


monkopedia-coder:
Surfaced by #122 step 1.1 (cppfrontend self-generation). LLVM-22 added `clang::ASTContext::getPredefinedSugarType(clang::Type::PredefinedSugarKind)`; the enum param is a **protected** member of `clang::Type`. Because `ASTContext` is allowlisted, krapper_gen bound the method and emitted a wrapper cast to the protected enum, which the C++ wrapper compile rejects (`'PredefinedSugarKind' is a protected member`).

Worked around in cppfrontend with `removeMethod("clang_ASTContext_get_predefined_sugar_type")`, but that's a manual, LLVM-version-specific patch that will recur every time a pinned LLVM adds protected/private surface on an allowlisted class.

**Proposed:** in the generator, detect when a bound method's parameter/return type is inaccessible from the wrapper's context (protected/private nested type, etc.) and skip that method automatically (log it), rather than emitting an uncompilable cast. This makes the allowlist robust across LLVM bumps and removes a class of manual fixups.

Not a flip/bootstrap blocker — the removeMethod workaround suffices — but it's the durable fix.

Issue #122, opened 2026-07-08
cppfrontend self-generation + stage-0 anchor → per-LLVM-major binary distribution


monkopedia-coder:
## Goal
Enable kplusplus to be **distributed** as: the plugin jar + **N prebuilt `cppfrontend` binaries, one per supported LLVM major**, with the plugin selecting the matching one at runtime. Consumer contract = "install LLVM major N (dev pkg) + GCC-12-era libstdc++ + glibc" — a normal dev prerequisite, no consumer-side build of our tooling.

This unifies three previously-separate threads: (a) retire the checked-in `cppfrontend/krapped/` seed, (b) support multiple LLVM majors, (c) the stage-0 → mavenLocal crossover.

## Why this shape (from the portability probes, branch `ci/portability-probe-1`)
- **Load** is gated on an exact `libLLVM.so.<major>` SONAME match → one binary per major (no cross-major load). Two extra floors: libstdc++ **GLIBCXX_3.4.30** (GCC 12+) and **glibc-only** (no musl).
- **Build/source drift** across majors is **97/98 incidental** — the generator over-binds new LLVM surface `cppfrontend` never calls; only 1 used symbol (`SuppressInlineNamespaceMode`) differs, and only at LLVM 19.
- ⇒ We do **not** need a maintained seed per major. **Generate the bindings on the fly per build against the local LLVM**, from ONE mostly-shared source set. The N binaries become a CI build-matrix over one source, not N regenerated seeds.

## Design
- **On-the-fly generation, not checked in.** `:cppfrontend` generates its own bindings during the build (against whatever LLVM is present) instead of compiling a committed seed. Drift never bites because we never cross-compile.
- **One stage-0 generator bootstraps all versions.** `cppfrontend` parses arbitrary C++ headers, so a single stage-0 (built against, say, 22) can parse LLVM 20/21/22's Clang headers to emit each version's bindings.
- **Source sets:** `common` (hand-written, version-agnostic — ~99%, since we call essentially only stable API) + thin per-version sets (`llvm22`, `llvm20`, …) holding the on-the-fly generated bindings + the tiny version shims (so far: 1 symbol).

## Plan — build for LLVM 22 first, architected for N
[...]

PR #125, opened 2026-07-08
cppfrontend self-generation seam: stage-0 override + PredefinedSugarType fixup (#122 step 1.1)

monkopedia-coder:
Foundation for #122 (step 1.1 keystone). Additive + gated — **no-op for the default build.**

## What
Two small changes that unlock `:cppfrontend` self-generation (the compiler generating its own Clang-AST bindings instead of compiling the committed seed):

1. **`resolveCppFrontend` bootstrap override** — when `-Pkpp.stageZeroCppFrontend=<abs-path>` is set, use that external stage-0 cppfrontend binary and declare **no** task dependency. This breaks the Gradle self-cycle: building `:cppfrontend` via `-Pkpp.frontend.cppfrontend=cpp` would otherwise resolve the parser to `:cppfrontend`'s own link task → a dependency cycle. Absent property = today's behavior byte-for-byte.
2. **`removeMethod("clang_ASTContext_get_predefined_sugar_type")`** in cppfrontend's `fixup {}` — LLVM-22 added `ASTContext::getPredefinedSugarType`, whose param is the *protected* enum `clang::Type::PredefinedSugarKind`; the generated wrapper cast doesn't compile. Nothing in the bridge uses it. (Generalized fix tracked in #123.)

## Why it's safe to land now
- The default build has `kpp.frontend.cppfrontend=seed` and never sets `kpp.stageZeroCppFrontend`, so neither change is exercised — the override returns early only when the property is present, and `removeMethod` only runs on the `=cpp` generation path.
- It's the seam step 1.2 (stage-0 packaging) builds on.

## Proof (spike, not in this PR)
With these changes, the full two-stage self-rebuild works end to end: seed → stage-0 → stage-0 parses `clang_slice.h` (+ `kppbridge` + both `std::vector` forcings) → 124 generated binding files → stage-1 builds (3m14s) + runs + can re-parse its own header. The self-generated bindings are a **superset** of the committed seed (the 20 differing files are the *seed* being stale — missing #114/#62/LLVM-22 surface; self-gen is more current). Bootstrap closed and reproducible.

Refs #122. Does not close it (epic).

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -546,6 +546,16 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
      */
     private fun resolveCppFrontend(target: Project): Pair<Any?, File> {
         val relBinary = "cppfrontend/build/bin/klinker/cppfrontendRelease/cppfrontend"
+        // Bootstrap override (#122, step 1.1): when an EXTERNAL stage-0 cppfrontend binary is
+        // supplied via -Pkpp.stageZeroCppFrontend=<abs-path>, use it and declare NO task
+        // dependency (null link task). This is what breaks the self-hosting cycle: building
+        // :cppfrontend via -Pkpp.frontend.cppfrontend=cpp would otherwise resolve the parser
+        // to :cppfrontend's own link task -> a Gradle dependency cycle. With a null task the
+        // caller (dependsOn only when non-null) wires no self-edge, and runCppFrontendSync
+        // uses this .second path as the parser. Additive + gated: absent property = today's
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Approve (tier-1, infra). Verified both changes against the safe-to-land claim:

- `resolveCppFrontend` override is a genuine no-op when `kpp.stageZeroCppFrontend` is absent/blank — the early return is guarded on `takeIf { it.isNotBlank() }`, so the default build takes the existing `:cppfrontend` project branch unchanged. Returning a null task is consistent with the only `.first` consumer (line 81, `?.let { dependsOn }` → no self-edge when null) and `.second` supplies the parser binary path (line 282) via the override's `File(path)`.
- `removeMethod("clang_ASTContext_get_predefined_sugar_type")` matches the surrounding fixup pattern exactly and is inert in `=seed` mode (krapper_gen is neither built nor run there, per the seed-mode comment at lines 64-66), so it only fires on the `=cpp` generation path.

Comments accurate, idiom fits the file, no scope creep. Default build configures cleanly (`:krapper_gen:help` green; `:cppfrontend` is correctly absent without -PenableClang). LGTM.

Issue #128, opened 2026-07-09
First shippable v2 release: consume-from-published path + acceptance test (LLVM 22)


monkopedia-coder:
## Goal
Reach the **first shippable release** of kplusplus (v2, self-hosted) — a coherent, tagged release a real consumer can use from **published artifacts**, targeting **LLVM 22** as the single supported major (multi-version breadth follows in a separate pass, see #122).

## The core gap
**No one has ever consumed kplusplus from published artifacts.** Both samples `includeBuild("../../")` (pull source). The plugin locates `krapper_gen`/`cppfrontend` only as sibling projects or `gradle.includedBuilds` — a real consumer (declares the plugin id + a repo, no includeBuild) would have neither, so the plugin can't find its tools. There is also no publish infra (except #126's stage-0 cppfrontend) and versions drift (root `0.2.3` vs `compiler/` `0.2.2`).

## Strategy
Prove shippability **to mavenLocal first**; defer the public channel (Central vs GitHub Packages) to the last mile. The consumer-resolution + acceptance-test work is channel-independent and is the real substance. Release version target: **0.3.0**.

## Consumer artifact set (what a from-published consumer needs — to confirm/refine in brick 1)
- The **Gradle plugin** (`com.monkopedia.kplusplus.compiler`, from `compiler/gradle`) — resolvable via `id(...) version "0.3.0"` from a repo/plugin portal.
- The **FIR compiler plugin** jar (`compiler/plugin` [+ `compiler/native`]) — added to the kotlinc plugin classpath by the Gradle plugin.
- `krapper_gen` native binary — invoked to resolve+generate.
- `cppfrontend` native binary (LLVM-22) — invoked to parse.
- (`krapper_model` if it's a runtime dep of the above.)

## Bricks
- [ ] **R1 (this brick):** map the full consumer artifact set precisely; unify versions (`compiler/` 0.2.2 → the release version); make the plugin (Gradle + FIR) publishable to **mavenLocal**. Report the exact remaining publish/resolution gaps.
- [ ] **R2 — consumer-resolution path (keystone):** teach `resolveKrapperGen`/`resolveCppFrontend` (and the FIR-plugin classpath wiring) to resolve from **published Maven artifacts** when no in-tree sibling/includedBuild exists — extending the #126 stage-0 pattern (classified artifact → staged +x → run). Publish the native tools to mavenLocal.
[...]

PR #126, opened 2026-07-08
cppfrontend: stage-0 mavenLocal bootstrap anchor (#122 step 1.2)

monkopedia-coder:
## What

Step 1.2 of epic #122. Publishes the **cppfrontend stage-0 binary** to **mavenLocal** as an immutable, non-SNAPSHOT bootstrap anchor and wires a clean consume seam so the `=cpp` self-generating build can resolve its stage-0 parser from a Maven coordinate instead of a hand-typed filesystem path.

This is a **one-time transitional anchor** (steady-state will later be "last release as stage-0"), so it favors the simplest thing that works — an artifact-only publication of the raw K/N executable.

## Coordinates

`com.monkopedia.kplusplus:cppfrontend-stage0:0.2.3-stage0` classifier `linuxX64`, empty extension (bare ELF, not a jar).

## Changes (single file: `cppfrontend/build.gradle.kts`)

- Apply `maven-publish` + an **artifact-only** publication (a K/N linuxX64 executable has no software-component to wire).
- **`publishStage0ToMavenLocal`** — depends on `linkReleaseExecutableKlinker`, enforces the **from-seed** invariant (refuses a `-Pkpp.frontend.cppfrontend=cpp` build, which would be circular). The gradle.properties default `kpp.frontend.cppfrontend=seed` means the default link already builds from the committed seed = the bootstrap source.
- A resolvable **`stage0`** configuration + **`resolveStage0Path`** task that pulls the artifact by coordinate, stages an executable copy under `build/stage0/` (Maven strips the +x bit), and prints its path for `-Pkpp.stageZeroCppFrontend` (the step 1.1 override seam, PR #125).

Additive + gated with the module (`-PenableClang`). The **default `=seed` build is completely untouched**: nothing sets `kpp.stageZeroCppFrontend`, and publish/consume are opt-in.

## Verified end-to-end (LLVM 22, JDK 21)

1. `publishStage0ToMavenLocal` → `~/.m2/repository/com/monkopedia/kplusplus/cppfrontend-stage0/0.2.3-stage0/` contains the 17MB ELF + POM.
2. `=cpp` build consuming the **mavenLocal** stage-0 (resolved path, NOT a /tmp path) → self-generated **124** Kotlin bindings + the stage-1 binary linked.
3. Default `=seed` build with no stage-0 property → still builds from the committed seed (no krapper_gen, no libclang).

## Consume flow (opt-in, two-step)

```
./gradlew :cppfrontend:publishStage0ToMavenLocal -PenableClang
[...]

PR #131, opened 2026-07-09
consumer(R2): from-published tool resolution + samples/minimal as a real external consumer (#128)

monkopedia-coder:
## R2 (epic #128) — prove a REAL from-published consumer works

Starts from R1/#130 (version unified to 0.3.0, the full consumer set published to mavenLocal). Today the plugin only finds the `krapper_gen`/`krapper_parse` tool binaries as in-tree sibling projects or `includeBuild`s. A true external consumer (declares the plugin + a repo, **NO `includeBuild`**) has neither. This makes that work end-to-end against mavenLocal and proves it by converting `samples/minimal` into a genuine from-published consumer that BUILDS AND RUNS — the same path k++ will use to self-host as its own first consumer.

### (A) Plugin — resolve the tool binaries from published Maven artifacts
`KPlusPlusCompilerGradlePlugin.kt`:
- New `resolvePublishedTool()`: a resolvable `Configuration` depending on `com.monkopedia.kplusplus:<tool>:0.3.0` with the `linuxX64` classifier + empty extension + `isTransitive=false`, resolved to a `File`, copied to `build/kplusplus/tools/<tool>` and `setExecutable(true)` (Maven drops the exec bit). Modeled directly on the #126 stage-0 consume seam (`resolveStage0Path`).
- `resolveKrapperGen` / `resolveKrapperParse` append this fallback **LAST** → resolution order is **in-tree sibling → includedBuild → published**. Existing dev/self-build flows are byte-for-byte unchanged; the fallback only fires when there is no in-tree tool.
- The FIR plugin jar keeps resolving via `getPluginArtifact()`'s `SubpluginArtifact` (Kotlin pulls it from the consumer's declared repos) — confirmed a `mavenLocal()` consumer picks it up.
- Robustness: `krapperGenKexe` becomes an `inputs.files` (plural — tolerates a staged path wiped by a same-invocation `clean`) and the cpp `doLast` re-resolves/re-stages the tool at execution time, so `clean run` in one invocation works.

### (B) `samples/minimal` → pure from-published consumer
- `settings.gradle.kts`: drop `includeBuild("../../")`; `pluginManagement { repositories { mavenLocal(); gradlePluginPortal(); mavenCentral() } }` resolves the plugin marker.
- `build.gradle.kts`: apply `id("com.monkopedia.kplusplus.compiler") version "0.3.0"`; `repositories` already carry `mavenLocal()`+`mavenCentral()` for the tool config + FIR plugin + ksrpc-jni.
- `gradle.properties`/`README`: **`-PenableClang` is NOT needed** for a standalone consumer (it only gates modules inside the kplusplus root build, which this consumer does not include); the published `krapper_parse` is LLVM-linked and needs `clang++` on `PATH` at runtime.

## Verification
- **From-published build + RUN** (no `includeBuild`, no `-PenableClang`), from a wiped `build/`:
  ```
  PARITY_BASE .../build/krapped-cpp/cpp-models/base_model.json
  kplusplusSync[minimal]: krapper_gen (cpp model) over 0 instantiation(s) ... -> .../build/krapped-cpp
  kplusplusSync[minimal]: cpp front-end generated 3 Kotlin file(s).
  Rect area: 6.0
  distance (0,0)->(3,4): 5.0
  center: (1.0, 1.5)
  BUILD SUCCESSFUL
  ```
- Tools resolved from `~/.m2`: staged `build/kplusplus/tools/{krapper_gen,krapper_parse}` are **byte-identical (md5)** to `~/.m2/repository/com/monkopedia/kplusplus/{krapper_gen,krapper_parse}/0.3.0/...-linuxX64`.
- `clean runReleaseExecutableKlinker` in a single invocation also passes (re-stage-on-execute robustness).
- **In-tree dev flow unbroken**: `:featuregen:kplusplusSync -PenableClang -m` (dry-run) still resolves via the sibling `:krapper_gen:linkDebugExecutableNative` / `:krapper_parse:linkReleaseExecutableKlinker` link tasks — no mavenLocal fallback, no tool staging.

`samples/v8` still uses `includeBuild` — left as-is.

Refs #128 (R2). mavenLocal only; nothing outward. Do not merge without review.

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -99,7 +99,15 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
             val ext = target.extensions.findByType(KPlusPlusExtension::class.java)
             val hasHeadersAtConfig = ext?.headers?.isNotEmpty() == true
             if (!seedMode) {
-                it.inputs.file(kexe).withPropertyName("krapperGenKexe")
+                // inputs.files (plural), NOT inputs.file (singular): for a from-published
+                // consumer (R2, #128) the kexe is a mavenLocal artifact staged under
+                // build/ (resolvePublishedTool). If a `clean` in the SAME invocation wipes
+                // build/ AFTER configuration staged it, singular inputs.file would abort at
+                // validation on the missing path; the plural form tolerates a non-existent
+                // entry, lets the task run, and the doLast re-stages it via resolveKrapperGen
+                // (idempotent) or fails fast on a genuinely-missing tool. When the file is
+                // present it still drives the up-to-date check identically.
+                it.inputs.files(kexe).withPropertyName("krapperGenKexe")
                     .withPathSensitivity(org.gradle.api.tasks.PathSensitivity.NONE)
             }
             it.inputs.files(manifestFile).withPropertyName("requestedManifest")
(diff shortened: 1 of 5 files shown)


monkopedia-reviewer (approved):
Approve (tier-1, infra) — reviewed both axes.

**Ordering / additivity (the load-bearing property): correct.** In both `resolveKrapperGen` and `resolveKrapperParse` the new `resolvePublishedTool` call sits AFTER the sibling-project check and AFTER the `includedBuilds` loop — it is the LAST resolution attempt before the legacy hardcoded-path fallback. Dev / includeBuild flows early-`return` before it, so `config.resolve()` never runs in those cases (no config-time resolution cost, no behavior change). featuregen (in-tree sibling link tasks) and samples/v8 (includeBuild) are byte-for-byte unchanged.

**Coordinate + selector match the published publications:** `com.monkopedia.kplusplus:{krapper_gen,krapper_parse}:0.3.0` with artifact `name=<tool>`, `classifier=linuxX64`, `extension=""`, `isTransitive=false` — matches krapper_gen/krapper_parse build.gradle.kts publications exactly (artifactId + linuxX64 + empty extension).

**+x/copy pattern faithfully mirrors resolveStage0Path (#126):** copy out of the immutable Maven store, `setExecutable(true)` (Maven drops the exec bit), staged under build/. Same-invocation `clean run` robustness (plural `inputs.files` + `doLast` re-resolve) is sound and the re-stage is idempotent.

**samples/minimal is a genuine from-published consumer:** includeBuild dropped, pluginManagement has mavenLocal(), plugin applied with `version "0.3.0"`. Docs/gradle.properties correctly explain -PenableClang isn't needed standalone.

Scope clean: diff is exactly the plugin resolver + samples/minimal wiring + docs (5 files) — no KotlinWriter/generator/API change. `./gradlew help` resolves clean (exit 0). Trusted the coder's end-to-end from-published run per policy (resolution-order change, mavenLocal only, verified byte-identical staged tools).

Non-blocking: `resolvePublishedTool` catch-all returns null, masking a genuine (vs absent) resolution failure behind the generic fail-fast message — acceptable for a best-effort fallback seam.

Judgment lever: mavenLocal only, no outward publish/signing/credentials — release/publish-automation lever does not fire. Tier 1.

PR #134, opened 2026-07-10
fix(release): build native release binary before Central publish (v0.3.0 hotfix)

monkopedia-coder:
**Hotfix for the failed v0.3.0 Central publish.**

The cut fired the release workflow: the Gradle plugin published live to the Plugin Portal, but the Central upload **failed** — `Invalid publication 'releaseBinary': artifact file does not exist: .../releaseExecutable/krapper_gen.kexe`.

**Root cause:** the 'build the release binary first' dependency (`dependsOn(linkRelease…)`) was attached only to the **mavenLocal** publish task, not the **Central** repository publish task. So local validation passed (a stale release binary was present) but clean CI failed. Same latent bug in `krapper_gen` and `krapper_parse`.

**Fix:** generalize the dependency via `tasks.matching { … }` to cover both `publishReleaseBinaryPublicationToMavenLocal` **and** `…ToMavenCentralRepository`, for both modules. Verified by dry-run — the Central publish task now depends on `linkReleaseExecutableNative`/`linkReleaseExecutableKlinker`.

**Also:** gate the Plugin Portal step to the `release` event only — the plugin 0.3.0 is already live on the Portal (only Central failed), and the Portal rejects re-publishing an existing version, so a `workflow_dispatch` re-run does **Central-only** to complete 0.3.0 without a version bump.

Refs #128.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,8 +40,12 @@ jobs:
           sudo apt-get install -y clang llvm llvm-dev libclang-cpp-dev gcc-multilib
 
       # --- Publish the Gradle plugin to the Gradle Plugin Portal ---
-      # publishPlugins uses gradle.publish.key / gradle.publish.secret.
+      # publishPlugins uses gradle.publish.key / gradle.publish.secret. Runs ONLY on an actual
+      # release event: the Plugin Portal rejects re-publishing an existing version, so a manual
+      # workflow_dispatch re-run (e.g. to retry a failed Central upload) skips this step and does
+      # Central only — the plugin is already published from the original release.
       - name: Publish Gradle plugin to the Plugin Portal
+        if: github.event_name == 'release'
         run: >-
           ./gradlew -p compiler :kplusplus-compiler-gradle:publishPlugins
           -Pgradle.publish.key=${{ secrets.GRADLE_PUBLISH_KEY }}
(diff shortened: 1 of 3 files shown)


monkopedia-reviewer (approved):
Approved (tier-1, infra / release-tooling).

Correct root-cause fix. The old `tasks.named("...ToMavenLocal")` wired the 'build the release binary first' dependency onto the mavenLocal publish task only, so a clean-CI Central publish had no link dependency and failed 'artifact file does not exist' (it passed locally only on a stale binary). Verified:

- Both modules' `tasks.matching {}` predicate matches exactly the two publish tasks (`...ToMavenLocal` OR `...ToMavenCentralRepository`) via exact-name equality — no wildcard leakage. Same `dependsOn` + existence `check` as before; krapper_parse keeps its `!stage0BuiltFromCpp` seed guard. mavenLocal path behavior unchanged.
- Correct link deps: krapper_gen -> linkReleaseExecutableNative, krapper_parse -> linkReleaseExecutableKlinker.
- Central task-name spelling is corroborated by the adjacent disable-block (`publish\${name}PublicationToMavenCentralRepository`); `releaseBinary` is intentionally kept out of that disable list, so its Central task exists.
- Dry-run confirmed: `:krapper_gen:linkReleaseExecutableNative` is in the plan ahead of `:krapper_gen:publishReleaseBinaryPublicationToMavenCentralRepository`.
- Workflow `if: github.event_name == 'release'` correctly limits the Plugin Portal step to release events (plugin 0.3.0 already live; Portal rejects re-publish), so a workflow_dispatch re-run does Central-only; the Central step stays unconditional.

Scope is the 2 build files + release.yml, no codegen. Unblocks the partial v0.3.0 Central publish.

PR #135, opened 2026-07-10
fix(release): builtBy on native artifacts so the sign task depends on the link task

monkopedia-coder:
**Second v0.3.0 Central-publish fix** (the first, #134, moved the failure).

The Central publish now failed at signing: `':krapper_gen:signReleaseBinaryPublication' uses this output of task ':krapper_gen:linkReleaseExecutableNative' without declaring an explicit or implicit dependency` — Gradle 9 rejects it. #134's per-publish-task `dependsOn` covered the *publish* tasks but not the *sign* task.

**Fix:** declare `builtBy(linkRelease…)` on the artifact itself, so **every** consumer of the .kexe — the sign task and both publish tasks (mavenLocal + Central) — depends on the producing link task. Applied to `krapper_gen` (`linkReleaseExecutableNative`) and **both** `krapper_parse` artifacts, releaseBinary + stage0 (`linkReleaseExecutableKlinker`).

**Validated locally** (not just dry-run): with a throwaway in-memory GPG key, `:krapper_gen:publishReleaseBinaryPublicationToMavenLocal` runs `linkReleaseExecutableNative → signReleaseBinaryPublication → publish` **BUILD SUCCESSFUL** with no implicit-dependency error — reproducing the exact failing CI path.

Sign runs before upload, so the failed run published nothing to Central (verified 404). Refs #128.

### DIFF
--- a/krapper_gen/build.gradle.kts
+++ b/krapper_gen/build.gradle.kts
@@ -143,6 +143,11 @@ publishing {
             artifact(krapperGenBinary) {
                 classifier = "linuxX64"
                 extension = ""
+                // Declare the producing task on the artifact so EVERY consumer of the .kexe —
+                // the sign task AND the publish tasks (mavenLocal + Central) — depends on it.
+                // (A per-publish-task dependsOn missed signReleaseBinaryPublication, which
+                // Gradle 9 rejects as an undeclared implicit dependency.)
+                builtBy(tasks.named("linkReleaseExecutableNative"))
             }
             artifact(emptySourcesJar)
             artifact(emptyJavadocJar)
--- a/krapper_parse/build.gradle.kts
+++ b/krapper_parse/build.gradle.kts
@@ -483,6 +483,7 @@ publishing {
             artifact(krapperParseBinary) {
                 classifier = stage0Classifier
                 extension = ""
+                builtBy(tasks.named("linkReleaseExecutableKlinker"))
             }
         }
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Approve — correct, complete, and idiomatic release hotfix.

Verified:
- **Correct link task per module.** krapper_gen's artifact points at bin/native/releaseExecutable/krapper_gen.kexe → builtBy(linkReleaseExecutableNative); krapper_parse's artifacts point at bin/klinker/krapper_parseRelease/krapper_parse → builtBy(linkReleaseExecutableKlinker). Both mappings corroborated by the pre-existing dependsOn on the Exec tasks that run those same binaries.
- **No native artifact missed.** All three .kexe artifacts get builtBy: krapper_gen releaseBinary; krapper_parse releaseBinary + stage0. The empty sources/javadoc jars (their own Jar tasks) are correctly left alone.
- **Behavior = task-edge only.** classifier/extension/coordinates unchanged → from-published consumers and self-host unaffected.
- **Idiomatic.** builtBy on the artifact attaches the producer to every consumer of the output (sign + both publish tasks), superseding #134's per-publish-task dependsOn that missed signReleaseBinaryPublication — exactly what Gradle 9 wants for the undeclared-implicit-dependency rejection.
- Config evaluates cleanly (./gradlew help on Gradle 9.4.1); coder validated the full linkRelease → sign → publish path locally with a throwaway GPG key.

Unblocks the live v0.3.0 Central publish. Merging tier-1.

PR #136, opened 2026-07-10
fix(release): install LLVM 22 (apt.llvm.org) + CPATH in the release workflow

monkopedia-coder:
**Third v0.3.0 Central-publish fix.** The prior fixes (#134/#135) unmasked this by letting the build proceed past krapper_gen.

`:krapper_parse:kplusplusSync` failed: *'stage-0 seed wrapper compile failed'* — the release workflow installed the **distro** clang (`apt-get install clang llvm`), but krapper_parse's committed seed (`krapper_parse.cc`, `#includes <clang/AST/...>`) was **generated against LLVM 22** and only compiles against 22 (the cross-major API drift the portability probe characterized). apt.llvm.org's headers also live under `/usr/lib/llvm-22/include`, not the distro default path.

**Fix:** mirror the probe's proven install — LLVM 22 from apt.llvm.org, unversioned `clang++`/`llvm-config` symlinks, and `CPATH`/`LD_LIBRARY_PATH` from `llvm-config` so the seed compile + klinker link + runtime all resolve. (This same install is what the portability-probe build job used to compile krapper_parse successfully.)

Workflow-only; merging publishes nothing. After merge I re-dispatch → the build should now complete and reach the actual Central upload. Refs #128.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,14 +30,28 @@ jobs:
       - name: Set up Gradle
         uses: gradle/actions/setup-gradle@v4
 
-      # krapper_parse links against the system LLVM/Clang toolchain (libclang-cpp + libLLVM);
-      # the parser binary is built FROM SEED (a plain C++ recompile) but still needs clang++
-      # and the LLVM libs present on the runner. -PenableClang brings :krapper_parse into the
-      # build; the settings probe discovers the LLVM major + libdir from llvm-config.
-      - name: Install LLVM/Clang toolchain
+      # krapper_parse is built FROM SEED (a plain C++ recompile of its committed krapped/
+      # wrapper). That wrapper #includes <clang/AST/...> and was GENERATED against LLVM 22 —
+      # it only compiles against LLVM 22 (the cross-major C++ API drift proven in the
+      # portability probe), and apt.llvm.org puts those headers under /usr/lib/llvm-22/include,
+      # NOT the distro default path. So install LLVM 22 from apt.llvm.org (not the distro's
+      # clang) and surface its include/lib dirs via CPATH/LD_LIBRARY_PATH so the seed compile,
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
APPROVE — release-blocking hotfix, workflow-only, correct.

Verified:
- The seed IS clang-major-coupled: krapper_parse/krapped/krapper_parse.cc + include/clang_slice.h #include <clang/AST/DeclCXX.h>/<clang/AST/ASTContext.h>/etc. Distro clang (wrong major, headers under distro default not /usr/lib/llvm-22/include) cannot compile it — the fix installs LLVM 22 from apt.llvm.org, which is the right toolchain.
- The settings probe (settings.gradle.kts:65-68) resolves 'llvm-config' from PATH under -PenableClang (no -PllvmConfig override in the release step), so symlinking llvm-config-22 -> /usr/local/bin/llvm-config makes the probe pick LLVM 22. Correct.
- software-properties-common installed before llvm.sh (add-apt-repository gotcha handled). set -euxo pipefail fails fast.
- CPATH/LD_LIBRARY_PATH exported via GITHUB_ENV -> inherited by the gradle seed-compile / klinker link / runtime steps. CPATH covers <clang/AST/...> resolution; LD_LIBRARY_PATH covers link+runtime for libclang-cpp/libLLVM.
- Blast radius: only the install step changed; publish steps, secrets, -PenableClang, automaticRelease all untouched. Single clean commit atop #134/#135.

Mirrors the portability-probe install that already compiled krapper_parse -> cppfrontend, so it's a known-good pattern, not a guess. Not locally reproducible (CI apt install) as the PR honestly notes; the remaining unknown (OSSRH token auth against the Central Portal) only surfaces at the upload this unblocks -> not a merge blocker. Tier-1 auto-merge.

PR #146, opened 2026-07-11
feat(codegen): migrate carried MemScope field to Kotlin context parameters

Monkopedia:
## What

Implements the MemScope-field → context-parameter migration described in `docs/design/memscope-context-params.md` (follows §8 exactly). The carried `memScope` field is **dropped from every generated wrapper**; the allocation/cleanup scope now flows through a `context(scope: MemScope)` parameter at each site that needs one.

### Codegen (`KotlinWriter.kt`)
- **Allocating / cleanup-registering sites** emit `context(scope: MemScope)` and run `alloc`/`_Holder`/`defer` on the ambient `scope`:
  - `_Holder` factory, direct & stack constructor factories, the string-based template-facade factory
  - by-value `ARG_CAST` returns (instance methods / allocating operators) — the enclosing member gains the context param via `defineArgsAndBody`
  - `owned()`
- **Borrowed / non-allocating sites** carry `ptr` only (no scope): up/down-cast views, container element access, by-`&`/`*` returns, `dispose()`, plain scalar/ptr methods.
- Wrappers become `class T(val ptr: COpaquePointer)`.
- A **by-value-wrapper field getter** can't be a context-carrying property accessor, so it is emitted as a `context(scope) fun name()` (with the setter, if any, a plain `fun`).
- **Template/base interface methods** whose concrete override allocates — including every `T`-returning method (a wrapper instantiation like `Box<Point>.get()` allocates) — carry the context param uniformly, and the concrete overrides are forced to match (a scalar instantiation like `Box<Int>.get(): Int` then carries an unused-but-consistent context param).

### Builder
`contextParams` moved onto the shared `FunctionBuilder` and is propagated to the signature in `FunctionSymbol.init()`, so both `function {}` and `extensionFunction {}` emit context params.

### Self-hosting consumers
- `krapper_parse`'s hand-written `ModelBuilder`/`TypeBuilder` thread the one ambient `MemScope` (every QualType-by-value accessor — `getType`/`getReturnType`/`getCanonicalType`/… — is now `context(scope)`), dropping the former `wrapper.memScope` field reads and constructor args.
- The in-tree consumers (`featuregen`, `krapper_parse`, `cppfixture`, `clangwalk`) add `-Xcontext-parameters` to their K/N compile options.
- Hand-written `featuregen`/`cppfixture` tests updated to call the context-param API (bare factory/allocating calls inside `memScoped { }` / `with(Companion) { }` instead of `ms.Factory(...)`).

## Proof
- **`:featuregen:nativeTest -PenableClang` → 188 tests, 0 failures** (aggregated `TEST-*.xml`).
- `:krapper_gen:nativeTest` → **288/0** (golden expectations in `FullKotlinTests`/`KotlinCodeTests` updated to the new shapes).
- `:krapper_parse:kplusplusSync -PenableClang` re-syncs clean through the changed generator (124 files), and all downstream K/N compiles are green.

## Notes / deviations
- Per §3 the **stack constructor** factory takes `context(scope: MemScope)` even though it doesn't heap-allocate (the stack object lives only for the callback) — this keeps `memScoped { Foo(...) { … } }` call-site parity with the former MemScope-receiver form. All other sites follow §8's allocate-vs-borrow split literally.
- The template-interface uniform-context rule (a `T`-returning method forces context on every override) is the one place a non-allocating concrete method (`Box<Int>.get()`) gains a context param; this is required for override-signature compatibility and is harmless (unused param).


monkopedia-reviewer (approved):
Gates: `:krapper_gen:nativeTest` **288/0**, `:featuregen:nativeTest -PenableClang` **188/0**, `:krapper_parse:kplusplusSync -PenableClang` 124 files clean. All three re-run locally, independently verified.

**Correctness**: Design §8 decision rule applied exactly across every emission site in `KotlinWriter.kt`. Verified generated output for all four shape categories:
- Plain value type (`geo_Vec`): `class Vec(val ptr: COpaquePointer)`, `context(scope: MemScope)` on factories and `Vec_Holder()`, plain scalar methods scope-free. ✓
- Owning/polymorphic (`Animal`): factory + `owned()` take context, `asDog()`/`asPuppy()` cast views carry only `ptr`. ✓
- Container (`std_Vector__Point`): element access (`get`/`at`/`front`/`back`/`data`) is scope-free; constructors/`_Holder`/`owned()` carry context. ✓
- Template scalar instantiation (`Box__Int`): `get(): Int` carries an unused `context(scope: MemScope)` matching the interface — correct, required for override-signature compatibility. ✓
- By-value field getter (`ConstRetHolder.origin_()`): emitted as `context(scope: MemScope) fun origin_()` via `generateByValueFieldAccessors`, setter as plain `fun`. ✓

**Kotlin idiom / call-site simplification confirmed**: `val ms = this` boilerplate gone uniformly; `with(Foo) { FactoryCall() }` inside `memScoped {}` resolves cleanly through the KEEP §7.5 receiver-satisfies-context-param rule. Call sites got simpler, not just relocated. The `with(Companion) {}` usage pattern for free functions is unaffected.

**Two intentional deviations from §8** (coordinator asks 3a/3b):
- Stack-ctor factory carrying `context(scope)` despite no heap alloc: right call — preserves `memScoped { Foo(...) { } }` call-site parity with the former `MemScope.Foo(...)` receiver form; the wrapper handed to the callback is `ptr`-only as required.
- `Box<Int>.get(): Int` carrying an unused context param: right call — the `context(scope: MemScope) fun get(): T` interface declaration is uniform (required because `Box<Point>.get()` allocates), and Kotlin generics can't parameterize on whether T is scalar; the unused param is harmless and consistent.

**Builder refactor**: `contextParams` correctly promoted to `FunctionBuilder` base, removing the duplicate field on `KotlinFunctionSymbol`. `scopeContext()` freshly allocates per call (appropriate, given FQ-import walker rejects shared Symbol instances). `allocatesByValueReturn()` and `generateByValueFieldAccessors()` are clean extractions at the right scope. No dead code, no scope creep, no style ledger violations.

**Line count**: net +157 non-test lines in KotlinWriter.kt — the added complexity sits in the generator (right altitude), not the generated output. Call sites in tests uniformly shorter.

PR #152, opened 2026-07-12
krapper_gen: skip methods whose signature references an inaccessible (protected/private) nested type (#123)

### LINKED ISSUE #123: krapper_gen: skip methods whose signature references an inaccessible (protected/private) type

monkopedia-coder:
Fixes #123.

## Problem

A public method whose parameter/return type is a **protected or private nested tag type** (enum/record) can't be bound: the generated wrapper is a free function, so it can't name the inaccessible type to cast to it. The LLVM-22 repro is `clang::ASTContext::getPredefinedSugarType`, whose param is the protected enum `clang::Type::PredefinedSugarKind` — the wrapper compile rejects the emitted cast (`'PredefinedSugarKind' is a protected member of 'clang::Type'`). Until now the only fix was a manual, per-symbol, LLVM-version-specific `removeMethod` fixup that recurs on every LLVM bump.

## Detection + where it hooks into the drop machinery

The signal is the referenced type's clang access specifier. `krapper_parse`'s `TypeBuilder.buildWrappedType` decodes a `QualType` to a `WrappedType` and has the `Decl` in hand at both tag-type leaves (record and enum). New helper `Decl.isAccessibleTagType()` returns false for `AS_protected`/`AS_private` — exactly the accessor + values `ModelBuilder` already uses to filter private/protected members (`Decl::getAccess()`). When a leaf is inaccessible, the builder returns `UNRESOLVABLE`.

That's the whole hook: an `UNRESOLVABLE` signature type flows into the resolver's **existing** "couldn't resolve type" path — `ModelResolution.resolvePlainMethod`/`resolveArgument` → `ResolveContext.notifyFailed` → the method is dropped and recorded in the `DropLedger` as a `DropPhase.RESOLVE` drop, identically to any other unbindable type. No new drop seam was needed. `AS_none` (top-level, non-member tags) and `AS_public` members are untouched, so accessible-signature siblings still bind.

## The `removeMethod` fixup

The durable fix makes the `clang_ASTContext_get_predefined_sugar_type` fixup unnecessary for a **newly built** parser — but it is **retained** as a bootstrap crutch, because `:krapper_parse:kplusplusSync` self-generates krapper_parse's own bindings with the **pinned stage-0 tool binary**, which predates this fix. Removing the fixup makes that stale tool re-emit the broken cast and the self-sync aborts (exit 134, reproduced). The comment now notes the fixup can be dropped once the stage-0 tool is bumped past #123.

## Tests

featuregen `AccessProbe` fixture (`strings_feature.h`): a protected nested enum param (`useHiddenEnum`) and a private nested struct return (`makeHiddenTag`) both **drop**; a method over the accessible top-level `Color` enum (`useColor`) and a plain-int method (`plain`) still **bind**. Verified in the generated C header — only `AccessProbe_new/_use_color/_plain/_size_of/_align_of` are emitted. `TiInaccessibleTypeTest` asserts the surviving siblings behave. Added the `TI-inaccessible` row to `docs/features.md`.

## Observed gates (JAVA_HOME=21, -PenableClang, LLVM 22)

- `:featuregen:kplusplusSync` (cpp front-end, rebuilt krapper_parse): green — 119 Kotlin files.
- `:featuregen:nativeTest`: **191 tests, 0 failures, 0 errors** (includes `TiInaccessibleTypeTest`, 2 cases).
- `:krapper_gen:nativeTest`: green.
- `:krapper_parse:kplusplusSync` (self-host re-sync): green — 124 Kotlin files (with the retained bootstrap fixup).
- ktlint: no new violations on the touched files (`TypeBuilder.kt` clean; pre-existing `build.gradle.kts` script violations are unrelated).

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -290,6 +290,7 @@ A multi-namespace probe (`NsSingleTest`/`NsNestedTest`/`NsNestedClassTest`/
 | NS-nested-class | `Outer::Inner` | `package outer`, class `Inner`, `import outer.Inner`; `Outer` stays `root.Outer` | 🟢 | the outer class name becomes a package segment — so two `Outer::Inner` separate naturally (the review's "collide in root" prediction was wrong) |
 | NS-collision | `acolor::Tag` / `bflavor::Tag` | distinct packages + distinct C prefixes | 🟢 | no link/runtime clash; same-file Kotlin use needs an import alias (`import bflavor.Tag as FlavorTag`) |
 | NS-anon | anonymous `namespace { Hidden }` | (skipped — not bound) | 🟢 | members are now skipped (TU-internal linkage, not referenceable across a C ABI) instead of emitting an invalid empty `package ` line — `WrappedElement.map` returns null for an empty-spelling namespace |
+| TI-inaccessible | public method over a `protected`/`private` nested type — `int use(HiddenMode)` / `HiddenTag make()` | (method skipped — not bound) | 🟢 | a nested tag type declared `protected`/`private` can't be named from the free-function wrapper (the LLVM-22 repro: `ASTContext::getPredefinedSugarType(clang::Type::PredefinedSugarKind)`, param is a protected enum). `TypeBuilder` resolves such a leaf to `UNRESOLVABLE`, so the method drops through the same `notifyFailed` path as any unbindable type; sibling methods over accessible types still bind. Replaces a manual per-symbol `removeMethod` fixup. `TiInaccessibleTypeTest` (#123) |
 
 ## Inheritance & virtual dispatch
 
--- a/featuregen/src/cppMain/include/strings_feature.h
+++ b/featuregen/src/cppMain/include/strings_feature.h
@@ -871,4 +871,29 @@ struct RefHolder {
     int note;
 };
 
+// T-inaccessible (#123): a public method whose SIGNATURE references a protected/private
+// nested type can't be bound — the generated wrapper is a free function, so it can't name
+// the type to cast to it (the LLVM-22 repro: `clang::ASTContext::getPredefinedSugarType`
+// takes the protected enum `clang::Type::PredefinedSugarKind`). Skip-not-crash: the
+// generator DROPS such a method (its signature type resolves to UNRESOLVABLE), while
+// sibling methods with accessible signatures still bind. Two inaccessible kinds are
+// covered — a protected nested ENUM and a private nested STRUCT — plus a method over the
+// accessible top-level `Color` enum as a control that must still bind.
+struct AccessProbe {
+protected:
+    enum HiddenMode { HM_A = 1, HM_B = 2 };   // protected nested enum
+private:
+    struct HiddenTag { int t; };              // private nested struct
+public:
+    AccessProbe() {}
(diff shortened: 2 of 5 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. All correctness checks pass.

**AS_none handling**: correct. `isAccessibleTagType()` blocks only `AS_private` and `AS_protected`; `AS_none` (the access specifier for top-level / namespace-scope tags) falls through as accessible — exactly matching how LLVM documents `Decl::getAccess()` for non-member entities. No false drops.

**Pointer/reference/typedef coverage**: the check fires at the record and enum tag-decl leaves, which are reached via recursion from the pointer/reference branches (`buildWrappedType(ty.getPointeeType())`). A method taking `HiddenTag*` or `const HiddenMode&` is caught identically to the direct-value case. The one early-exit path for dependent pointer/reference shapes (lines 403-408) only fires for dependent types inside templates, not for concrete inaccessible tag types — no leak.

**Drop machinery reuse**: UNRESOLVABLE → existing `resolvePlainMethod`/`resolveArgument` → `notifyFailed` → `DropPhase.RESOLVE` ledger entry. No new seam added. Confirmed in the generated binding: only `AccessProbe_new/use_color/plain/size_of/align_of` appear in `featuregen.h` and `root_AccessProbe.kt`; the two inaccessible-signature methods are absent.

**Bootstrap crutch reasoning**: confirmed real. `:krapper_parse:kplusplusSync` uses the pinned stage-0 tool (0.3.2) which predates this fix; without the `removeMethod` the stale tool re-emits the broken cast and aborts. The comment is clear and actionable (drop once stage-0 is bumped past #123). featuregen's sync uses the freshly-built parser and DOES exercise the new code path — the fix is genuinely active in the test suite.

**Test quality**: `TiInaccessibleTypeTest` — 2 behavioral tests (plain sibling + accessible enum sibling), structural proof via compilation (the absent methods can't be referenced). Covers both protected enum and private struct cases. The comment explaining the structural coverage approach (compile-time absence vs runtime assertion) is clear and appropriate.

**Gates**: featuregen 191/0, TiInaccessibleTypeTest 2/2, krapper_gen nativeTest green, kplusplusSync self-host clean (124 Kotlin files). Diff is exactly the 5 intended files, no unrelated churn.

PR #106, opened 2026-06-15
krapper_gen: keep pointer indirection for value-reduced out-pointer params (#103 Family 1)

monkopedia-coder:
## What

Fixes **#103 Family 1** — an out-pointer param `T*` / `T& out`, whose pointee `T` reduces to a Kotlin scalar/enum, was bound BY VALUE, so the generated Kotlin facade passed a scalar where the C extern's `CValuesRef<...>` / `COpaquePointer` pointer is expected → **did not compile**. Surfaced by the v8 brick-3 surface (#102) and worked around at the example config level.

## Root cause

Two mis-collapsing shapes in `WrappedKotlinType.typeToKotlinType`'s pointer/reference branch:

- **`size_t*` / `size_t&`** — `size_t` is native (its wrapper C signature is a real `size_t*`), but it had **no `pointerTypeMap` entry**, so it fell through the `?:` fallback to a **bare `size_t` scalar**. Added `size_t -> kotlinx.cinterop.ULongVar` (size_t is `ULong` on LP64 — identical to `uint64_t*`, and to cinterop's own `CValuesRef<size_tVar>` since `size_tVar` aliases `ULongVar`).
- **`enum*`** — a pointer to a value-reduced enum surfaced as the generated `enum class`, so `KotlinWriter.reference()` emitted `arg.value` (an `Int`) into the wrapper's `void*` slot. The enum is **not** a `.ptr`-backed wrapper and its pointer cType is `void*`, so it now surfaces as the opaque `COpaquePointer` the caller fills/reads — exactly as a by-value `void*` arg. **Guarded on `isPointer`**, so an enum *reference* stays a by-value enum (`WrappedModifiedType.isEnum`), unchanged.

`void**` already mapped correctly to `CValuesRef<COpaquePointerVar>?` — locked as a regression test.

## Family 2 deferred

Family 2 (reference-RETURN of a value-reduced type — the `std::byte` / `_Ios_*` bitmask compound-assignment operators, whose C wrapper is **itself UB**: it returns a pointer to a by-value local) is a distinct return-path + `CppWriter` concern, meaningfully more involved/riskier. Deferred to **#105**; its example `removeMethod` workarounds are **kept**.

## New test

`OutPointerTypeTest` (6 cases) pins all four pointer shapes (`size_t*`, `size_t&`, `enum*`, `void**`) **and** the two unchanged shapes (by-value enum, enum reference) so the fix stays scoped.

## Example workarounds dropped (verified compiling)

[...]

Issue #103, opened 2026-06-15
[krapper_gen] Two codegen-bug families surfaced by v8 brick 3 (out-pointer-as-value; reference-return of a value-reduced type)


monkopedia-reviewer:
Follow-up from PR #102 (v8 brick 3). The v8 surface exposed two latent krapper_gen codegen-bug families. PR #102 worked around them at the *example* config level (18 `removeMethod` fixups) because none are on the hello-world path, but they should be fixed at the source.

## Family 1 — OUT-POINTER param modeled as a value
A `T* out` / `T& out` parameter (`enum*`, `size_t&`, `uint64_t*`, `void**`) is bound as a by-value param, so the generated Kotlin passes a scalar where the extern's `CValuesRef<...>` pointer is expected → does not compile.

Observed on:
- `v8::Maybe::To(PropertyAttribute* out)`
- `v8::String::GetExternalStringResourceBase(Encoding* out)`
- `v8::Isolate::GetCodeRange(void**, size_t*)` / `GetEmbeddedCodeRange(void**, size_t*)`
- `v8::platform::tracing::TraceBufferChunk::AddTraceEvent(size_t& index)`
- `v8::ValueSerializer::Delegate::ReallocateBufferMemory(..., size_t* actual)`

## Family 2 — REFERENCE-RETURN of a value-reduced type
An `operator|=` / `&=` / `^=` returning `T&` where `T` reduces to a Kotlin enum or `UByte` is bound as a by-value return; the extern returns the reference as a pointer, so the generated body feeds a `CPointer` into `fromValue(Int)` / a `UByte` slot. Worse, the C wrapper itself is UB: it returns a pointer to a by-value local. These are the `std::byte` / `<ios>` `_Ios_*` bitmask compound-assignment operators leaking in via transitive includes.

These are real reducer/codegen gaps, not example-specific. Tracking so the example's `removeMethod` workarounds can eventually be dropped.

Filed by monkopedia-reviewer during the #102 review.

monkopedia-coder:
[...]

PR #153, opened 2026-07-12
krapper_gen: lower a reference-return of a value-reduced enum by value (#105 Family 2)

### LINKED ISSUE #105: [krapper_gen] #103 Family 2: reference-RETURN of a value-reduced type (operator|=/&=/^= returning T&)
Split out from #103 (Family 1 — out-pointer params — is fixed; see the linked PR). This tracks the remaining, narrower Family 2.

## Symptom
An `operator|=` / `operator&=` / `operator^=` returning `T&`, where `T` reduces to a Kotlin enum or `UByte`, is bound as a by-value return. The C wrapper returns the reference as a pointer, so:
- the generated **Kotlin** feeds a `CPointer` into `fromValue(Int)` / a `UByte` slot (type mismatch, does not compile), and
- worse, the generated **C wrapper itself is UB**: `RETURN_REFERENCE` emits `return &(expr)` where `expr` is the value-reduced arg cast into a **by-value local** — i.e. it returns a pointer to a stack local.

These leak in transitively from libstdc++ `<ios>`: `std::byte& operator|=/&=/^=` and the `_Ios_Fmtflags`/`_Ios_Openmode`/`_Ios_Iostate` bitmask flag types. Pure transitive-include noise, unused by the example (the 12 `removeMethod` workarounds in `example/kotlin_project/build.gradle.kts` are KEPT).

## Why it is separate from Family 1
Family 1 lived entirely in `WrappedKotlinType.typeToKotlinType` (param type mapping). Family 2 is a RETURN-path concern that spans:
1. `CppWriter.generateReturn` `RETURN_REFERENCE` (`+Return(call.addressOf)`) combined with the arg-cast that materializes a value-reduced `T&` arg into a by-value local → `&local` UB.
2. The Kotlin return path (`KotlinWriter.generateReturn`): an enum/UByte return whose extern actually returns a pointer.

A correct fix likely needs the C wrapper to operate on (and return the address of) the real referent — not a by-value local — when the referent is a value-reduced `T&`, and the Kotlin side to read back through that pointer. This is meaningfully more involved and riskier than the Family 1 type-mapping fix, so it was deferred to keep that PR focused.

## Repro shape
`Cls& operator|=(Cls&, Cls)` where `Cls` reduces to an enum/`UByte` (e.g. `std::byte`, `std::_Ios_Fmtflags`).

Filed during the #103 Family-1 fix.

monkopedia-coder:
Fixes #105.

## Root cause
A method returning `Enum&` (the `std::byte&` / `_Ios_Fmtflags&` compound-assignment operators leaking in from `<ios>`, and the `Flag&` shape generally) was mis-lowered. `WrappedModifiedType.isEnum` is `true` for a `&`-reference to an enum, so `determineReturnStyle` already picked `ENUM_RETURN` — but the resolved return type was still the **reference**, whose `cType` is `void*` (a `&` to a non-native pointee). So the `ENUM_RETURN` C cast emitted the ill-formed `return (void*)(enumLvalue)`, the extern returned a pointer, and the Kotlin side fed that pointer into `fromValue(Int)`. Neither the wrapper nor the Kotlin compiled.

This is the narrower Family 2 split out of #103 (Family 1 — out-pointer params — was fixed in #106). Note Family 2 has a **sub-distinction** the issue's initial hypothesis didn't fully separate:
- **Enum reference return** (`Flag&`) — the actual bug. An enum has no `.ptr` storage to wrap; reading through the reference yields the enum *value*.
- **Class reference return** (`ByteFlags& operator|=`) — already correct via the existing OB-return-ref path (`return (void*)&(receiver)` C-side + wrapper reconstruction Kotlin-side, like `Vec2& operator+=`). Left untouched; covered by a regression test.

## Fix (the lowering)
In `resolvePlainMethod`, strip the reference for an enum-reference return so it resolves as a **by-value enum** (`cType` = the underlying integer). The wrapper then emits `return (<underlying>)(call)` (the `Enum&` auto-converts to `Enum`, cast to its integer) and Kotlin reads it via `fromValue` — identical to any other enum return. Non-enum (class) reference returns keep their pointer marshaling.

Generated output before → after for `Flag& FlagProbe::orInto(Flag)`:

```
// before (uncompilable):
void*        FlagProbe_or_into(unsigned int b) { ...; return (void*)FlagProbe::orInto(b_cast); }
// Kotlin: fun orInto(b: Flag): Flag? = fromValue(FlagProbe_or_into(b.value))   // CPointer into fromValue(Int)

// after:
unsigned int FlagProbe_or_into(unsigned int b) { ...; return (unsigned int)FlagProbe::orInto(b_cast); }
// Kotlin: fun orInto(b: Flag): Flag = fromValue(FlagProbe_or_into(b.value))
```

## Tests
- **featuregen `OpCompoundAssignEnumTest`** (behavioral, cpp front-end): `FlagProbe::orInto` (`Flag&`) reads the accumulated value + mutation is observable via `stateVal()`; `ByteFlags& operator|=/&=/^=` mutate and the returned wrapper chains onto the same storage. Fixture (`Flag` enum + `ByteFlags`/`FlagProbe`) added to `strings_feature.h`.
- **krapper_gen `ReferenceReturnEnumTest`** (codegen unit): resolves `ENUM_RETURN` with the underlying-integer `cType`, and asserts the C wrapper's return casts to the underlying integer, not `void*`/`&(`.

## Observed gates
- `:krapper_gen:nativeTest` — **291/0** (was 289; +2 new).
- `:featuregen:nativeTest -PenableClang` — **193/0** (cpp front-end; +2 from the new behavioral test), verified via the per-class result XML.
- `:krapper_gen:ktlintCheck` — my changed/new files are clean. The task is red only on **pre-existing** violations in `FullKotlinTests.kt` (untouched by this PR); not a regression I introduced.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -290,6 +290,7 @@ A multi-namespace probe (`NsSingleTest`/`NsNestedTest`/`NsNestedClassTest`/
 | NS-nested-class | `Outer::Inner` | `package outer`, class `Inner`, `import outer.Inner`; `Outer` stays `root.Outer` | 🟢 | the outer class name becomes a package segment — so two `Outer::Inner` separate naturally (the review's "collide in root" prediction was wrong) |
 | NS-collision | `acolor::Tag` / `bflavor::Tag` | distinct packages + distinct C prefixes | 🟢 | no link/runtime clash; same-file Kotlin use needs an import alias (`import bflavor.Tag as FlavorTag`) |
 | NS-anon | anonymous `namespace { Hidden }` | (skipped — not bound) | 🟢 | members are now skipped (TU-internal linkage, not referenceable across a C ABI) instead of emitting an invalid empty `package ` line — `WrappedElement.map` returns null for an empty-spelling namespace |
+| TI-inaccessible | public method over a `protected`/`private` nested type — `int use(HiddenMode)` / `HiddenTag make()` | (method skipped — not bound) | 🟢 | a nested tag type declared `protected`/`private` can't be named from the free-function wrapper (the LLVM-22 repro: `ASTContext::getPredefinedSugarType(clang::Type::PredefinedSugarKind)`, param is a protected enum). `TypeBuilder` resolves such a leaf to `UNRESOLVABLE`, so the method drops through the same `notifyFailed` path as any unbindable type; sibling methods over accessible types still bind. Replaces a manual per-symbol `removeMethod` fixup. `TiInaccessibleTypeTest` (#123) |
 
 ## Inheritance & virtual dispatch
 
--- a/featuregen/src/cppMain/include/strings_feature.h
+++ b/featuregen/src/cppMain/include/strings_feature.h
@@ -871,4 +871,29 @@ struct RefHolder {
     int note;
 };
 
+// T-inaccessible (#123): a public method whose SIGNATURE references a protected/private
+// nested type can't be bound — the generated wrapper is a free function, so it can't name
+// the type to cast to it (the LLVM-22 repro: `clang::ASTContext::getPredefinedSugarType`
+// takes the protected enum `clang::Type::PredefinedSugarKind`). Skip-not-crash: the
+// generator DROPS such a method (its signature type resolves to UNRESOLVABLE), while
+// sibling methods with accessible signatures still bind. Two inaccessible kinds are
+// covered — a protected nested ENUM and a private nested STRUCT — plus a method over the
+// accessible top-level `Color` enum as a control that must still bind.
+struct AccessProbe {
+protected:
+    enum HiddenMode { HM_A = 1, HM_B = 2 };   // protected nested enum
+private:
+    struct HiddenTag { int t; };              // private nested struct
+public:
+    AccessProbe() {}
+    // DROPPED: param is the protected enum (can't be named from the wrapper).
+    int useHiddenEnum(HiddenMode m) const { return (int)m; }
+    // DROPPED: return type is the private nested struct.
+    HiddenTag makeHiddenTag() const { return HiddenTag{7}; }
+    // BOUND: param is the accessible top-level `Color` enum (control for the drop).
+    int useColor(Color c) const { return (int)c; }
+    // BOUND: a plain int accessor sibling, always bindable.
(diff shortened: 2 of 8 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. Two fixes, both verified:

**#105 Family 2 (ModelResolution.kt):** The reference-strip is correctly scoped: `mapped.first.isReference && mapped.first.isEnum` fires ONLY for `Enum&` (WrappedModifiedType with modifier=="&" and baseType.isEnum), NOT for class references like `ByteFlags&`. Generated bindings confirm: `FlagProbe_or_into` returns `unsigned int` via `return (unsigned int)FlagProbe::orInto(b_cast)` — the enum reference is stripped and returned by value. `ByteFlags_operator_oreq` returns `void*` via `return (void*)&(thiz_cast->operator|=(f_cast))` — the OB-return-ref path is fully intact. Mutation is preserved: the C++ call still executes (it modifies the static `FlagProbe::state`), and the test asserts via `stateVal()`.

**#123 (TypeBuilder.kt):** `isAccessibleTagType()` correctly allows AS_none (top-level) and AS_public, blocks only AS_private/AS_protected — mirrors the existing ModelBuilder.addMember() pattern exactly. AccessProbe generated binding emits only `useColor` and `plain`; `useHiddenEnum` and `makeHiddenTag` are absent.

**Tests:** `ReferenceReturnEnumTest` asserts ENUM_RETURN style AND underlying-int cType AND rejects void*/&( in the return. `OpCompoundAssignEnumTest` asserts mutation (stateVal) and class-reference chaining. `TiInaccessibleTypeTest` structural proof via compile + surviving-sibling asserts.

**Gates verified locally (no remote CI in this repo):** :krapper_gen:nativeTest 291/0 (new ReferenceReturnEnumTest 2 cases included). :featuregen:nativeTest -PenableClang 193/0 (new OpCompoundAssignEnumTest + TiInaccessibleTypeTest). ktlint violations are pre-existing in KotlinWriter.kt and FullKotlinTests.kt (untouched); ModelResolution.kt and all new files are clean.

PR #157, opened 2026-07-12
docs: Option-b cross-request memo hazard analysis (broad-forcing design)

monkopedia-coder:
Captures the design pass on issue #10's follow-up to PR #156 (which bounded the forcing **pass-3** re-resolve, ~half the win). This expands `docs/design/broad-forcing-resolve.md` with the **Option (b)** analysis: cross-request memoization of the remaining **pass-1** O(bound)-per-request cost.

## Verdict: safe-with-caveats, and possibly not worth it
Cross-request memoization is sound ONLY under a four-condition key:
1. **value-based** — store the finished `ResolvedClass`, never re-run `resolve` on a cached element (`resolve` mutates in place);
2. **non-consumer subset ONLY** — container consumers (`CXXRecordDecl`/`DeclContext`…) resolve to a *different method set per request* by design, so they're request-dependent and must NOT be memoized (the original Hybrid sketch had this inverted);
3. **disabled under INCLUDE_MISSING** — INCLUDE's on-demand materialization mutates the shared tracker; a memo hit would starve a sibling's availability;
4. **process-reset per run** (the #11a `elementLookup` leak class).

Safe subset ≈ **3000 of ~3221** (opaque leaves); pass-1 drops from O(bound×N) to ≈O(bound + refset×N).

## The honest risk (why this is a decision, not a slam-dunk)
The only byte-identical machine-check we have is featuregen, which runs **INCLUDE_MISSING** — and the memo is **disabled** there. So featuregen is byte-identical *by construction* and **structurally cannot exercise the memo**; (b)'s correctness on the path that matters (IGNORE_MISSING broad) rests on the predicate argument + a targeted unit test, NOT regenerate-and-diff. That's the same gate-blindness that made #156 abandon filtering pass 1.

And even if implemented, the **base bind (~3700 classes, once)** is untouched and is likely the *next* wall — so pass-1 memo buys a faster forcing stage in front of an unchanged base-bind stage. The doc documents the **offline-measurement alternative** (accept O(bound×N), run broad-force periodically off the build gate) as a possibly-better cost/risk trade.

Docs-only (+257/-7). No production code. Sections b.1–b.10 + Hybrid/#156 cross-reference corrections.

### DIFF
--- a/docs/design/broad-forcing-resolve.md
+++ b/docs/design/broad-forcing-resolve.md
@@ -237,6 +237,241 @@ instantiation requests the same ~3221 classes are re-resolved from scratch N tim
 - **Verdict:** viable as a *complement*, but it still pays O(bound) per request for the
   classes that genuinely change nothing.
 
+> The bullets above were the original terse sketch. The section below is the **PR-#156
+> follow-up investigation** that expands (b) into a concrete, safety-verified design — because
+> after #156 restricted pass 3, **pass 1 re-resolving every bound class on every request is
+> the remaining dominant cost**, and cross-request memoization of pass 1 is the named lever.
+
+#### (b.1) What pass 1 actually is, and why it re-runs N times
+
+Each `requestInstantiation` (IndexedServiceImpl.kt:134) calls `loadForcingModel(path)`
+(Parsing.kt:131), which **`ModelIo.decodeFromString`s a FRESH `WrappedTU`** (a brand-new
+object graph — the forcing TU is re-decoded from JSON per request; there is no cursor→element
+memo across loads, Parsing.kt:105 comment). `resolveForcing` then builds a **fresh
+`ResolveTracker` via `withClasses`** (Resolver.kt:252, `ResolveContext.withClasses` at
+Resolver.kt:566 constructs `ResolveTracker(...)` anew and resets the `NameHandler`). So across
+N instantiation requests, pass 1 (Resolver.kt:267-277) re-resolves the *same base-bound class
+set* — re-decoded, fresh-tracker, from scratch — **N times**. That is the O(bound × N) cost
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Docs-only: confirmed diff touches only `docs/design/broad-forcing-resolve.md` (+257/-7, no code). The Option-b (b.1–b.10) hazard analysis and the Hybrid/#156 corrections are internally coherent:

- **Consumer vs non-consumer partition (b.3):** correctly identifies consumers as request-DEPENDENT (method set varies per forced container → cumulative-recovery machinery depends on re-resolution) and non-consumers as the memoizable candidate set. The reasoning is sound.
- **INCLUDE_MISSING memo disable (b.5 condition 3):** correctly notes that under INCLUDE_MISSING the on-demand tracker-mutation side effect means a memo hit would silently skip pulling siblings into the tracker — so the memo must be disabled entirely there. Also correctly notes featuregen (INCLUDE_MISSING) can't machine-check the memo path, and names that as the honest residual risk (b.10).
- **Gate-blindness caveat (b.10):** appropriately hedged — the featuregen machine-check structurally cannot exercise the IGNORE_MISSING memo path, so (b)'s safety rests on predicate argument + targeted unit test #3, not the regenerate-and-diff gate. Named and not overclaimed.
- **Hybrid/#156 correction:** the original terse Hybrid note said 'memo for consumer set only' — the correction (b.3 + end of §5) correctly flips this to 'non-consumers that also reference no consumer'. Consistent throughout.
- **Untouched base-bind wall (b.7):** honestly stated — (b) makes forcing cheap but leaves the base-bind ~3700-class resolve unchanged; the base-bind wall is the likely next bottleneck.

No contradictions or overclaims found. Analysis appropriately-hedged throughout.

PR #158, opened 2026-07-13
experimental flag system + flag-gated resolve timing

monkopedia-coder:
## What

krapper_gen is effectively a compiler, so it keeps accreting **experiments** (resolve strategies, caches, alternate codegen) and **diagnostics** (timing, dumps) we want gated behind a switch. Today those are scattered (`KRAPPER_ROUNDTRIP_MODEL` env, `-Pkpp.frontend.<module>`, one-off `--dumpParsedModel`/`--roundTripModel`). This gives experimental + diagnostic toggles **one consistent home** — a small registry, not a framework — architected to extend.

First consumer: a flag-gated timing helper wired into the forcing resolver, the instrumentation the upcoming cross-request memoization experiment needs.

## The flag system

**Declaration (single source of truth)** — `ExperimentalFlags.kt`. Each flag is declared once with a namespaced name, a type (`BoolFlag`/`StringFlag`/`IntFlag`), a default, and a one-line description, then added to `all`:

```kotlin
val DIAG_TIMING = ExperimentalFlag.BoolFlag(
    "diag.timing", default = false,
    description = "Log per-phase wall-clock timings (ms) + counts to stderr."
)
val all = listOf(DIAG_TIMING)
```

Read via a typed accessor anywhere: `ExperimentalFlags.isEnabled(DIAG_TIMING)` / `get(flag)` (returns the declared default when unset).

**Parsed once at startup** with precedence **CLI `-X` > env `KRAPPER_X` > default**:
- Env: `KRAPPER_X` = comma list of `name` (bool → on) or `name=value`, e.g. `KRAPPER_X=diag.timing,forcing.memo=lru`.
- CLI: repeatable `-X name[=value]` / `--experimental` merges over (and overrides) the env.
- Gradle: `-Pkpp.x=<value>` forwarded to krapper_gen as `KRAPPER_X` through one hook in `kplusplusSync`.

**Robust + discoverable:** unknown names **WARN** to stderr (typo visibility), never crash; active (non-default) flags logged to stderr at startup; `--list-experimental` dumps every declared flag.

**Off-by-default and inert:** with nothing set, every read returns the declared default → behavior and generated output byte-identical to today.

Not migrating the existing scattered toggles here (kept the diff tight); `docs/experimental-flags.md` lists the migration candidates.

## First consumer: `Diag.timed`

`Diag.timed(phase) { ... }` measures wall-clock via `kotlin.time.TimeSource.Monotonic` and logs `<phase>: <ms>` to stderr **only** when `diag.timing` is on; off = a plain passthrough (zero measurement). Wired into `resolveForcing`: pass 1, pass 2, pass 3, and a per-request total with bound/consumer/new/resolved counts.

## Demonstrated timing (real run, all 17 featuregen forcing requests)

```
kplusplus: active experimental flags: diag.timing=true
diag: forcing pass 1 (bind 75 bound class(es)): 797.199 ms
diag: forcing pass 2 (force 1 container(s)): 9.812 ms
diag: forcing pass 3 (recover 0 consumer(s)): 0.001 ms
diag: forcing request total: 862.777 ms (75 bound, 0 consumer(s), 2 new, 22 resolved)
... (17 requests total)
diag: forcing request total: 1397.635 ms (75 bound, 1 consumer(s), 2 new, 27 resolved)
```

(Debug-binary numbers, so absolute ms are inflated; the shape is the point — **pass 1 re-binds the same 75 classes on every one of the 17 requests**, ~750–920 ms each, which is exactly the redundancy the cross-request memo experiment targets.) The identical run with the flag OFF logs **zero** diag lines.

## Byte-identical proof (registry inert when unused)

- **On vs off (same generator, direct invocation over all 17 featuregen forcing models):** every generated `.h`/`.cc`/`.kt`/`.def` byte-identical; flag-OFF emits 0 diag lines. (The only apparent `.def` delta was the `-o` output-dir token, because the two runs used different output dirs — not a flag effect.)
- ****Clean main vs this branch (flags default):** regenerated featuregen's full cpp binding tree on both and diffed. Every generated `.kt`/`.h`/`.cc` byte-identical. The lone `.def` delta was purely the worktree-root path prefix in its `-I`/`libraryPaths` (the two builds ran in different directories); identical after normalizing that prefix. The registry is inert when unused.**

## Gates

- `:krapper_gen:nativeTest` green; new `ExperimentalFlagsTest` **9/9** (env parse, CLI-over-env precedence, unknown-flag warn, malformed-value warn, default inertness, active summary, listing).
- `:featuregen:nativeTest -Pkpp.frontend.featuregen=cpp` green: **195/0**.
- `:krapper_gen:ktlintCheck` green on the new files.

## Files

- `krapper_gen/.../ExperimentalFlags.kt` (registry), `Diag.kt` (timing helper)
- `krapper_gen/.../App.kt` (`-X`/`--experimental`/`--list-experimental` + startup init)
- `krapper_gen/.../Resolver.kt` (`Diag.timed` in `resolveForcing`)
- `compiler/gradle/.../KPlusPlusCompilerGradlePlugin.kt` (`-Pkpp.x` → `KRAPPER_X` hook)
- `krapper_gen/src/nativeTest/.../ExperimentalFlagsTest.kt`
- `docs/experimental-flags.md`

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -443,7 +443,14 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
             "kplusplusSync[$moduleName]: krapper_gen (cpp model) over ${requested.size} " +
                 "instantiation(s) from $headerPath -> $krappedDir"
         )
-        val exit = ProcessBuilder(args).inheritIO().start().waitFor()
+        // Experimental / diagnostic flag passthrough: `-Pkpp.x=diag.timing` (comma-separated
+        // for several) is forwarded verbatim to krapper_gen via its KRAPPER_X env var, so an
+        // experiment can be driven through the build without touching source. Off by default.
+        val procBuilder = ProcessBuilder(args).inheritIO()
+        (target.findProperty("kpp.x") as? String)?.takeIf { it.isNotBlank() }?.let {
+            procBuilder.environment()["KRAPPER_X"] = it
+        }
+        val exit = procBuilder.start().waitFor()
         if (exit != 0) {
             throw GradleException(
                 "kplusplusSync[$moduleName]: krapper_gen (--parsedModel) failed: exit $exit"
--- a/docs/experimental-flags.md
+++ b/docs/experimental-flags.md
@@ -1,0 +1,69 @@
+# Experimental / diagnostic flags
+
+krapper_gen is effectively a compiler, so it accretes *experiments* (new resolve strategies,
+caches, alternate codegen) and *diagnostics* (timing, dumps) that we want gated behind a switch
+rather than shipped on by default. Rather than let each grow its own ad-hoc mechanism, they share
+one home: the **experimental flag registry**.
+
+## Declaring a flag (the single source of truth)
+
+Every flag is declared once in
+[`ExperimentalFlags`](../krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ExperimentalFlags.kt)
+with a namespaced name, a type (`BoolFlag` / `StringFlag` / `IntFlag`), a default, and a one-line
+description, then added to the `all` list:
+
+```kotlin
+val DIAG_TIMING = ExperimentalFlag.BoolFlag(
+    "diag.timing",
+    default = false,
+    description = "Log per-phase wall-clock timings (ms) + counts to stderr."
+)
+
+val all: List<ExperimentalFlag<*>> = listOf(DIAG_TIMING)
+```
+
+Read it anywhere via the typed accessor:
+
+```kotlin
+if (ExperimentalFlags.isEnabled(ExperimentalFlags.DIAG_TIMING)) { ... }
+val mode = ExperimentalFlags.get(SomeStringFlag)   // returns the flag's default when unset
+```
+
+Names are dotted by area, e.g. `diag.timing`, `forcing.crossRequestMemo`.
+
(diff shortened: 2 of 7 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. Byte-identical proof verified independently (all 17 featuregen forcing specs, branch binary vs main binary, diff is ONLY the output-dir path in .def). All gates green.

**Inertness (the load-bearing property):**
- `Diag.timed` is `inline fun <T>`: off-path = `if (!timingEnabled) return block()` with no lambda allocation, no clock read, no measurement.
- `requestMark` in Resolver.kt: `if (Diag.timingEnabled) TimeSource.Monotonic.markNow() else null` — nothing allocated or read when timing is off; the mark is only consumed via `requestMark?.let { ... }`.
- The code inside each timing lambda is byte-for-byte identical to the pre-existing forEach body (verified in diff) — no control-flow change, no reordering, no changed return value.
- Byte-identical proof: ran branch binary and main binary over the full featuregen set (17 forcing specs, 6 build-seeded instantiations); `diff -r` of output is ONLY the output-dir path prefix in `featuregen.def` (-I / libraryPaths); every .h and .cc is identical.

**Gates:**
- `:krapper_gen:nativeTest`: GREEN — ExperimentalFlagsTest 9/9 (env parse, CLI-over-env precedence, unknown-flag warn, malformed-value warn, default inertness, active summary, listing)
- `:featuregen:nativeTest`: 195/0 (confirmed from test-results XML; kplusplusSync exit-134 is the pre-existing published-tool non-gate)
- `:krapper_gen:ktlintCheck`: GREEN

**Quality:**
- `ExperimentalFlags`: clean sealed hierarchy, pure `parse()` for testability, `internal` resetForTest.
- `Diag.timed` off-path is truly zero-overhead (inline + early return).
- Gradle passthrough is a single clean forwarding hook, no broader plugin churn.
- Flag registry is genuinely extensible — new flag is one declaration + `all` entry.

PR #159, opened 2026-07-13
experiment: forcing cross-request memo prototype (behind flag) + measurement

monkopedia-coder:
## What

Prototype of the Option-(b) **cross-request pass-1 memo** for `resolveForcing`, behind a new
OFF-by-default experimental flag, plus a measured report. Design reference:
`docs/design/broad-forcing-resolve.md` §b (b.1–b.10).

**Do not merge on the ON path** — this is a measurement. The prototype earns a merge only for its
OFF-inertness (proven below); the ON path is experimental and its benefit is not realizable on any
scenario we can byte-check (see honesty caveat).

## Change

- **New flag** `forcing.crossRequestMemo` (`BoolFlag`, default false) in `ExperimentalFlags`.
- **`ForcingPass1Memo`** (process-scoped) consulted in `resolveForcing` pass 1, strictly per the
  four-condition safe key (§b.5):
  1. **value-based** — caches the finished `ResolvedClass`, injected into the fresh request's
     tracker on a hit (never re-runs `resolve` against a mutated `WrappedClass`);
  2. **non-consumer + non-consumer-referencing subset only** — excludes container consumers
     (request-dependent method set) and classes referencing one (`referencesAnyContainer` +
     new `referencesAnyClass`, structural type-string scans);
  3. **IGNORE_MISSING only** — disabled under INCLUDE_MISSING (side-effecting on-demand
     materialization would make a hit unsafe);
  4. **process-reset per run** — reset in `IndexedServiceImpl` init alongside
     `DropLedger`/`GenerationContext`.
- Instrumented with hit/miss counters surfaced via `diag.timing`.

## Which policy featuregen uses (empirically determined)

The compiler plugin passes `-r INCLUDE_MISSING` by default
(`KPlusPlusCompilerGradlePlugin.kt:390`) and featuregen sets no `referencePolicy` override →
**featuregen pass-1 runs under INCLUDE_MISSING**, where the memo is disabled by condition 3. So
featuregen **cannot exercise the memo path at all** — exactly the gate-blindness the design warns
of (§b.10). The memo is active only on the **IGNORE_MISSING** path (clangwalk / broad #10).

## Scenario measured: clangwalk (IGNORE_MISSING, 3 instantiation requests)

The one completing, byte-checkable, IGNORE_MISSING multi-request scenario. Ran `krapper_gen`
directly on clangwalk's cached cpp-models, OFF vs ON.

| | value |
|---|---|
| bound classes (pass 1) | **17** |
| memoizable subset | **1** class |
| container consumers | 3–5 (the Decl hierarchy: `decls()/methods()/bases()`) |
| memo hits | req1: 0 hit / 1 miss; req2: 1 hit; req3: 2 hits |
| pass-1 time OFF | 56.6 / 54.2 / 131.6 ms |
| pass-1 time ON | 56.8 / 52.6 / 50.4 ms |
| pass-1 speedup | **none measurable** (Δ within run-to-run noise; memo saves resolving 1 of 17) |
| **output byte-identical OFF vs ON** | **YES** (same output dir → whole-tree sha256 identical) |

So on the scenario we CAN byte-check, the memo is **verifiably correct but its benefit is
negligible** — clangwalk binds 17 classes, not the ~3221 the memo targets, and only 1 is a safe
opaque leaf.

## The broad scenario does not complete (so can't be byte-checked)

`only(18)` under INCLUDE_MISSING (design Scenario B) **timed out at 240s in the BASE resolve**
(churning `llvm::MapVector` rewrites) — it never reached forcing at all. The broad blow-up is the
**base-bind**, which runs *before* pass 1 and is INCLUDE_MISSING (memo disabled there anyway). This
confirms §b.7's caveat: the memo makes *forcing* cheap but does nothing for the base-bind wall,
which is the actual gate.

## Honest verdict

This is the **measurable-XOR-verifiable bind** the design predicted (§b.10):
- The scenario where the memo is **verifiable** (clangwalk, IGNORE_MISSING, completes) has too few
  bound classes for the memo to help measurably.
- The scenario where the memo would **help** (~3221 bound) either doesn't complete (INCLUDE_MISSING
  base stall, memo disabled) or would need a purpose-built completing IGNORE_MISSING force over
  hundreds of classes — which I could not construct reliably from the available model, and which
  would still be un-byte-checkable against a non-completing baseline.

**Recommendation:** the OFF path is provably inert (byte-identical + gates green) and safe to keep
behind the flag as scaffolding, but **Option (b) does not demonstrate a measurable win on any
scenario we can also byte-verify.** Its correctness on the path that matters rests on the predicate
argument, not a machine check — the same gate-blindness that bit #156's pass-1 attempt. The
offline-measurement framing (§b.10 alternative) looks like the better cost/risk trade unless/until
a completing mid-size IGNORE_MISSING force exists to measure against.

## Gates (OFF-inertness — the merge-relevant claim)

- Output **byte-identical OFF vs ON** on clangwalk (real IGNORE_MISSING scenario, memo active).
- `:krapper_gen:nativeTest` green (includes `DeterminismTest.twoInProcessRunsAreByteIdentical`).
- `:krapper_gen:ktlintCheck` green.

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ExperimentalFlags.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/ExperimentalFlags.kt
@@ -86,9 +86,26 @@ object ExperimentalFlags {
         description = "Log per-phase wall-clock timings (ms) + counts to stderr."
     )
 
+    // EXPERIMENTAL (off by default): memoize the value-resolved form of pass-1's safely-
+    // memoizable bound classes across successive forcing requests, so the second and later
+    // --instantiate requests don't re-resolve the whole bound set from scratch. Guarded to the
+    // four safety conditions from docs/design/broad-forcing-resolve.md §(b.5): value-based
+    // (cached ResolvedClass, never a re-run against a mutated WrappedClass), restricted to the
+    // whole-run non-consumer + non-consumer-referencing subset, DISABLED under INCLUDE_MISSING
+    // (featuregen's policy — the memo buys nothing there and its on-demand materialization
+    // side effects make a memo hit unsafe), and process-reset per run. With the flag OFF this
+    // is byte-identical to today (the memo is never populated or consulted).
+    val FORCING_CROSS_REQUEST_MEMO = ExperimentalFlag.BoolFlag(
+        "forcing.crossRequestMemo",
+        default = false,
+        description = "EXPERIMENTAL: memoize pass-1's non-consumer bound-class resolutions " +
+            "across forcing requests (IGNORE_MISSING only). Off = byte-identical to today."
+    )
+
     /** Every declared flag. Add new flags here. */
     val all: List<ExperimentalFlag<*>> = listOf(
-        DIAG_TIMING
+        DIAG_TIMING,
+        FORCING_CROSS_REQUEST_MEMO
     )
 
     private val byName: Map<String, ExperimentalFlag<*>> = all.associateBy { it.name }
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/IndexedServiceImpl.kt
@@ -104,6 +104,10 @@ class IndexedServiceImpl(private val config: KrapperConfig, private val request:
         // (e.g. in writeTo) would be too late.
         DropLedger.reset()
         GenerationContext.reset(config.rootPackage, config.noRtti)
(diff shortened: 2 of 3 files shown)


monkopedia-coder:
Closing: the measurement (documented on #10, comment 4960152137) shows the forcing-loop cross-request memo is byte-safe but doesn't pay off — the broad-scale wall is the BASE-BIND (`filterAndResolve`, times out ~240s before forcing starts), not the forcing loop this prototype targets. The reusable half (flag system + timing) already landed in #158. Pivoting to profile/fix the base-bind instead. Prototype preserved on branch `experiment/forcing-crossrequest-memo` if the forcing loop ever becomes the bottleneck again.

PR #160, opened 2026-07-14
diag: base-bind profiling instrumentation (off-by-default)

monkopedia-coder:
## What

Adds a `diag.baseBindTiming` experimental flag + `BaseBindProfiler` that profiles the **base resolve** (`filterAndResolve` / `resolveAll` — the pass that runs *before* any forcing). Instrumentation only; no fix. This answers the open issue-#10 question: is the base-bind cost **uniform-breadth O(n)** or a **hotspot/superlinear O(n²)**?

## Where it is wired

- `ExperimentalFlags.kt` — the `diag.baseBindTiming` bool flag (off by default).
- `BaseBindProfiler.kt` (new) — process-scoped, gated, streaming profiler.
- `Resolver.kt` — `resolveAll` calls `begin()`/`report()`; the INCLUDE_MISSING `typeMapper` branch records on-demand materializations/re-entries.
- `Parsing.kt` — `ParsedResolver.resolve`/`resolveTemplate` time each genuine (non-memoized) resolution and count the full-TU `filterRecursive` walk + nodes visited. `resolve` refactored to extract `computeResolve` (the original `getOrPut` body); the OFF path is the exact original `classMap.getOrPut(key) { computeResolve(...) }`.
- `IndexedServiceImpl.kt` — `BaseBindProfiler.reset()` in the per-run init block (no cross-run leak).

Because the broad run can time out, the profiler **streams**: every 100 resolved types / 3 s it logs cumulative count, elapsed, instantaneous + cumulative classes/sec; keeps a running top-N slowest types; and reports the tree-walk counters.

## Byte-identical when off

Every profiler method early-returns when the flag is unset; the resolve path with the flag off is the exact original `getOrPut`. Measurement-only — it never changes what gets bound. Verified with release binaries over the `samples/minimal` fixture: generated output is **byte-identical** flag-off vs a pristine-main binary, and flag-on vs flag-off (the only delta is the `-o` output-dir path baked into the `.def`, an artifact of writing to two different dirs).

## Gates

- `:krapper_gen:nativeTest` (incl. DeterminismTest) — green.
- `:krapper_gen:compileKotlinNative` — green.
- `:featuregen:nativeTest -Pkpp.frontend.featuregen=cpp` — **fails with exit 134 (SIGABRT) in `kplusplusSync`, but this is PRE-EXISTING and environmental: it reproduces identically on pristine `main` HEAD with all of this PR's changes stashed.** Not a regression from this instrumentation.

## PROFILE RESULTS (the deliverable)

Reproduced the broad base-bind by running the release binary directly over the full Clang slice base model (`clangwalk/build/krapped-cpp/cpp-models/base_model.json`, 13 MB), `--referencePolicy INCLUDE_MISSING` + DefaultFilter (Scenario A, worst case), `-X diag.baseBindTiming`.

**Headline: the base bind COMPLETES (it no longer hangs indefinitely) — `resolveAll` returned at ~308.6 s having resolved 1597 classes.** So the "~240 s timeout" is really "~309 s of genuine work that finishes."

```
basebind: START — 2228 filtered classes
basebind: ===== REPORT =====
basebind: resolved=1597 total=308591.20ms avgRate=5.17/s avgPerType=193.23ms
basebind: treeWalks=17371 nodesVisited=1269915444 avgNodesPerWalk=73105.48
basebind: includeMaterializations=8938 reentries=0
basebind: top-25 slowest sum=91088.70ms (29.5 pct of total)
```

### Verdict: BOTH, but dominated by (a)-uniform-breadth with a specific O(n²) mechanism — plus ONE (b) pathological outlier

**1. The cost is a full-TU tree walk done ~11× per class.** `ParsedResolver.resolve`/`resolveTemplate` resolve each distinct type by calling `filterRecursive` — which walks the **entire ~73k-node TU with no short-circuit**, every time. Over the run: **17,371 walks × ~73,105 nodes = 1.27 BILLION node visits.** That is `O(distinct-types × tree-size)` — the concrete superlinear mechanism. avgPerType ≈ 193 ms is almost entirely these ~11 whole-tree scans per class.

**2. The rate is otherwise ~constant (~5 classes/s), i.e. uniform-breadth.** The rate curve is near-linear: 500 @ ~95 s, 1000 @ ~193 s, 1585 @ ~308 s. No runaway decay → the per-class cost is a roughly-fixed ~11 whole-tree walks, so total scales ~O(classes × tree-size). The top-25 slowest types account for only ~30 pct of total; the other ~70 pct is spread uniformly across the 1597 classes. This is breadth, not a single villain.

**3. One genuine outlier.** `clang::ObjCProtocolQualifiers<clang::ObjCObjectType>` alone took **43.3 s** (14 pct of the whole run) — a single template resolution cascading into a deep chain of full-TU walks. The rest of the top-25 are 1.7–2.6 s each (`llvm::ArrayRef<...>`, `llvm::MutableArrayRef<...>`, `std::map<...>`, `std::vector<...>` template specializations — each triggering its own tree walk per template arg).

`includeMaterializations=8938, reentries=0` — the INCLUDE_MISSING transitive expansion pulls in ~9k referenced types (2228 filtered → 1597 resolved + transitive), and the cycle-guard never re-enters, so there is **no redundant re-resolution of the same type** (`classMap`/`templateMap` memoize correctly). The waste is not re-resolution; it is that *each* first resolution re-scans the whole tree.

### Recommended lever (NOT implemented here)

**Fix the O(n²) tree-walk hotspot — build an index once, not a full-TU `filterRecursive` per type.** `ParsedResolver` should pre-build a `Map<type-string, WrappedClass>` and `Map<qualified, WrappedTemplate>` from a **single** `forEachRecursive` pass over the TU, then have `resolve`/`resolveTemplate` do an O(1) map lookup instead of an O(tree) `filterRecursive` scan. That converts the base bind from `O(distinct-types × tree-size)` to `O(tree-size + distinct-types)` — collapsing ~1.27 B node visits to ~73 k, and should take the ~309 s base bind to seconds. It is a pure-speedup, output-preserving change (same class/template selected — `singleOrNull()` semantics preserved by keying on the same strings), independent of and complementary to the forcing-loop levers in `docs/design/broad-forcing-resolve.md`.

Secondary: the `ObjCProtocolQualifiers` 43 s outlier is worth a separate look once the index lands (it may be a template-arg resolution fan-out that the index also mostly fixes).

### DIFF
--- a/docs/experimental-flags.md
+++ b/docs/experimental-flags.md
@@ -59,6 +59,27 @@ zero-overhead-when-off timing helper wired into `resolveForcing` (per-pass wall-
 counts, and a per-request total). This is the instrumentation the upcoming cross-request
 memoization experiment needs.
 
+## `diag.baseBindTiming` — base-resolve profiler
+
+`diag.baseBindTiming` profiles the **base resolve** (`filterAndResolve` / `resolveAll` — the pass
+that runs *before* any forcing), gating
+[`BaseBindProfiler`](../krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/BaseBindProfiler.kt).
+It exists to answer the issue-#10 question of whether the base-bind cost at broad Clang/LLVM scale
+is uniform-breadth (O(n)) or a hotspot/superlinear (O(n²)) — and because that run can time out, it
+emits **streaming** progress so a partial run still yields a profile:
+
+- every 100 resolved types (or 3 s), a `basebind: resolved=… elapsed=… inst=…/s cum=…/s walks=…
+  nodes=…` line — a decaying instantaneous rate is the O(n²) fingerprint;
+- a running **top-N slowest types** (names the pathological types, if any);
+- **tree-walk counters** — every `ParsedResolver.resolve`/`resolveTemplate` does a full-TU
+  `filterRecursive` walk, so `walks`/`nodes` expose the O(distinct-types × tree-size) cost directly;
+- INCLUDE_MISSING on-demand materialization + re-entry counts.
+
+`BaseBindProfiler.report()` (after `resolveAll` returns) dumps the final distribution. **Inert +
+byte-identical when off** (every method early-returns; the profiler is measurement-only and never
+changes what gets bound). Run it e.g. with
+`krapper_gen … --referencePolicy INCLUDE_MISSING -X diag.baseBindTiming`.
+
 ## Candidates to migrate later (NOT done here — kept out of scope)
 
 These pre-existing ad-hoc toggles could later move into the registry for consistency:
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/BaseBindProfiler.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/BaseBindProfiler.kt
@@ -1,0 +1,196 @@
+/*
+ * Copyright 2022 Jason Monk
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.monkopedia.krapper.generator
+
+import com.monkopedia.krapper.generator.Utils.printerrln
+import kotlin.time.TimeSource
+
+/**
+ * Process-scoped, off-by-default profiler for the BASE resolve (`filterAndResolve` / `resolveAll`
+ * — the pass that runs BEFORE any forcing). Gated by the `diag.baseBindTiming` experimental flag
+ * ([ExperimentalFlags.DIAG_BASE_BIND_TIMING]); when the flag is OFF, every method here is an inert
+ * early-return, so a normal run is byte-identical to one with no profiler wired in at all
+ * (measurement only — the profiler never changes what gets bound).
+ *
+ * It answers the issue-#10 question the base bind poses: is the ~240 s base-resolve cost
+ *   (a) UNIFORM-BREADTH — ~N classes each costing ~constant time, a steady classes/sec rate; the
+ *       lever is bounding the surface; or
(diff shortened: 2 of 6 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 standard. Inertness verified.

**Byte-identical-when-OFF proof (re-verified independently):** Built release binaries from both `main` and `diag/base-bind-profiling`. Ran both against `samples/minimal` and `cppfixture` models. Generated `.kt` sources, `.cc`, and `.h` are byte-identical across: (a) main vs branch-flag-OFF, (b) branch-flag-OFF vs branch-flag-ON. Only the `-I<outdir>` path in `.def` differs — expected, output-dir-baked, not a generation difference.

**Profiler hooks are pure observation:**
- `Resolver.kt` — `begin()`/`report()` are called unconditionally but both have `if (!enabled) return` as their first line → zero-overhead when off.
- `Parsing.kt / resolve()` — splits into a clean two-path: flag-OFF takes the exact original `classMap.getOrPut(key) { computeResolve(...) }` with a single `enabled` check overhead; flag-ON adds per-miss timing.
- `computeResolve()` and `resolveTemplate()` — the `if (BaseBindProfiler.enabled)` guard inside the `filterRecursive` lambda does add one property-getter call per node visit on memo miss even when flag is OFF. Not a result change — purely a minor per-node overhead confined to memo-miss code paths. Does not alter what gets resolved, returned, or stored.
- `Resolver.kt / typeMapper` — `recordInclude()` inserted at the existing branch points, each guarded by early-return in the callee. No ordering change.

**No sequence or resolution mutation:** the profiler only reads (counts, measures) — it does not add/remove elements, force eager evaluation of lazy sequences, or mutate any resolver state.

**`BaseBindProfiler.reset()` per-run:** correctly called in the IndexedServiceImpl init block after flags are stable. Early-return when OFF is correct (nothing to reset).

**Flag registry:** clean one-liner declaration in `ExperimentalFlags.kt`, added to `all` list, documented in `docs/experimental-flags.md`. Follows the exact same shape as `DIAG_TIMING` from #158.

**`printerrln` `%`-format footgun:** correctly worked around — the code uses "pct" and the comment on line 178-179 explicitly documents why no `%` literal appears. Pre-existing bug in `Utils.printerrln`, out of scope.

**featuregen exit-134:** reproduced identically on pristine `main` HEAD — pre-existing environmental issue, not a regression from this PR.

**Gates:**
- `:krapper_gen:ktlintCheck` — GREEN
- `:krapper_gen:nativeTest` (incl. DeterminismTest) — GREEN
- `:featuregen:nativeTest -Pkpp.frontend.featuregen=cpp` — exit 134, pre-existing, confirmed on main

Approved.

PR #161, opened 2026-07-14
fix(plugin): don't hard-crash IDE import when :krapper_parse is LLVM-gated out

monkopedia-coder:
## Root cause
A cpp-front-end module (`kpp.frontend.<module>=cpp`, e.g. `samples/v8`) wires its `:kplusplusSync` onto `:krapper_parse`'s link task. But `:krapper_parse` is **LLVM-gated**: the root `settings.gradle.kts` only `include()`s it under `-PenableClang`/`-PllvmConfig`.

In the **composite/included-build layout** (v8 pulls kplusplus in via `includeBuild("../../")`), `IncludedBuild.task(":krapper_parse:...")` returns a **lazy reference that does not validate the project exists**. So an IntelliJ Gradle sync — which passes no `-PenableClang` — got a dangling task ref into the included `:kplusplus` build where `:krapper_parse` was never included, and Gradle hard-failed at **graph resolution / model-fetch time**:

```
Could not determine the dependencies of task ':kplusplusSync'.
> Project with path ':krapper_parse' not found in build ':kplusplus'.
```

That aborts the **entire IDE import**, not just the one task. The pre-existing `isDirectory` guard in `resolveKrapperParse` doesn't fire here: the `krapper_parse/` directory exists on disk in the root build regardless of whether the project was `include()`d, so the guard passes and the dangling ref is still returned.

## The fix (guard Option B)
Gate the `:krapper_parse` `dependsOn` on the **same `-PenableClang`/`-PllvmConfig` signal that gates `:krapper_parse` into the build** — a new `clangEnabled()` helper that mirrors `settings.gradle.kts` exactly (`-P` properties propagate across the composite, so the consumer build sees the flag precisely when `:krapper_parse` was actually `include()`d). When clang is off, skip the `dependsOn` so **configuration succeeds** (the IDE loads the model). The LLVM requirement is **deferred to execution**, where `runKrapperParseSync` fails fast with the existing actionable "build with `-PenableClang`" message if the `krapper_parse` binary is genuinely absent.

Chose Option B over Option A (null-out the dangling ref) because the property gate uses the *exact* condition that governs inclusion — the two can never disagree — and it sidesteps fragile "does this lazy task ref dangle?" detection.

## Before / after (config-only, from `samples/v8`, no `-PenableClang`)
Before (on `main`): `./gradlew :kplusplusSync --dry-run` → **BUILD FAILED**, `Project with path ':krapper_parse' not found in build ':kplusplus'`.

After: `:kplusplusSync --dry-run`, `tasks`, and `help` all **BUILD SUCCESSFUL** — `:krapper_parse` correctly absent from the graph. This is the IDE-import-equivalent.

## Non-regression (WITH the flag — unchanged)
- v8 `:kplusplusSync --dry-run -PenableClang`: `:kplusplus:krapper_parse:linkReleaseExecutableKlinker` is in the graph — real dependency still wired.
- sibling featuregen cpp path `:featuregen:kplusplusSync --dry-run -PenableClang -Pkpp.frontend.featuregen=cpp`: `:krapper_parse:linkReleaseExecutableKlinker` in the graph — unchanged.

## Deferred requirement intact
Running `:kplusplusSync` without LLVM and without a prebuilt binary now fails at **execution** (not config) with the actionable message: `kplusplusSync[v8]: -Pkpp.frontend.v8=cpp but the krapper_parse binary was not found at ... — build with -PenableClang (or -PllvmConfig=<path>) ...`. Deferred, not lost.

## Validation
- `:kplusplus-compiler-gradle:compileKotlin` green.
- `:krapper_gen:nativeTest` green (sibling cpp modules unaffected).

Files touched: `compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt`

### DIFF
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -74,10 +74,21 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
             // ANY module flipped to cpp now self-wires the front-end build generically.
             // resolveKrapperParse covers BOTH the sibling layout (featuregen/cppfixture) and
             // the included-build layout (the standalone v8 example, where krapper_parse lives in
-            // the kplusplus build pulled in via includeBuild) — its link task is null when
-            // clang isn't enabled, in which case the doLast fails fast with the -PenableClang
-            // guidance.
-            if (target.findProperty("kpp.frontend.${target.name}") == "cpp") {
+            // the kplusplus build pulled in via includeBuild).
+            //
+            // The dependsOn is gated on `clangEnabled` — the SAME -PenableClang/-PllvmConfig
+            // signal that gates :krapper_parse into the build (settings.gradle.kts). This is
+            // load-bearing for the included-build layout: `IncludedBuild.task(":krapper_parse:…")`
+            // returns a LAZY reference that does NOT validate the project exists, so when the
+            // module is on the cpp front-end but clang is NOT enabled (e.g. an IntelliJ Gradle
+            // sync, which passes no -PenableClang), resolveKrapperParse hands back a dangling
+            // task ref into the included :kplusplus build — where :krapper_parse was never
+            // include()d — and Gradle hard-fails at graph resolution ("Project with path
+            // ':krapper_parse' not found in build ':kplusplus'"), aborting the ENTIRE IDE import.
+            // Skipping the dependsOn when clang is off keeps CONFIGURATION clean; the LLVM
+            // requirement is deferred to EXECUTION, where the doLast fails fast with the
+            // -PenableClang guidance (runKrapperParseSync / the missing-binary check below).
+            if (target.findProperty("kpp.frontend.${target.name}") == "cpp" && clangEnabled(target)) {
                 resolveKrapperParse(target).first?.let { ct -> it.dependsOn(ct) }
             }
             val krappedDir = krappedDirFor(target)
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Reviewed the fix to `KPlusPlusCompilerGradlePlugin.kt` — both axes sound.

**Correctness:**

`clangEnabled()` exactly mirrors `settings.gradle.kts:74`: `(enableClang != null && enableClang != "false") || llvmConfigProp != null`. Both signals (`enableClang` AND `llvmConfig`) are checked with identical truthiness, including the `!= "false"` special-casing. The two gates cannot disagree.

- **No-flag path (the bug):** `./gradlew :kplusplusSync --dry-run` from `samples/v8` on `main` → `BUILD FAILED: Project with path ':krapper_parse' not found`. On this branch → `BUILD SUCCESSFUL`. Verified.
- **With-flag path (non-regression):** `--dry-run -PenableClang` → `:krapper_parse:linkReleaseExecutableKlinker` is in the graph. The real dependency is still wired. Verified.
- **Featuregen sibling:** `:featuregen:kplusplusSync --dry-run -PenableClang -Pkpp.frontend.featuregen=cpp` → `:krapper_parse:linkReleaseExecutableKlinker` in graph. Unchanged. Verified.
- **Deferred, not lost:** When clang is off, the `dependsOn` is skipped so CONFIGURATION succeeds; the existing `runKrapperParseSync` / missing-binary `doLast` guard still fires at EXECUTION with the actionable `-PenableClang` message. The requirement is deferred, not silently dropped.
- **`:kplusplus-compiler-gradle:compileKotlin`** — green.
- **`:krapper_gen:nativeTest`** — green (39s, all passing).

**Quality:**

Comment block is dense but justified — the Gradle lazy-ref subtlety is genuinely non-obvious and the explanation saves the next reader from repeating the same investigation. The helper's KDoc names the two conditions and the reason the guard is load-bearing. Consistent with the comment density and style of surrounding code. The helper is small and pays for itself at multiple call sites (the inline `if` would need to be copy-pasted or the condition re-explained twice). Clean.

No CI checks on the branch (repo has no remote CI workflows per project setup); local verification covers the fix.

PR #163, opened 2026-07-15
fix(codegen): only emit the LLVM dyn_cast down-cast when well-formed for the base (broad-surface cluster, #10)

monkopedia-coder:
## The cluster (issue #10 broad surface)

PR #162 made the broad DefaultFilter + INCLUDE_MISSING force over the full Clang slice reach codegen for the first time; compiling the generated C++ yielded **1531 errors**, the largest bucket being **597 "cannot initialize a parameter … with an rvalue"**. This is that cluster.

## Root cause — ONE systematic codegen pattern

The generator emits a checked down-cast helper per (base, derived) pair. When the derived declares an LLVM `classof`, it uses the RTTI-free LLVM path:

```cpp
void* clang_AttributeCommonInfo_dyncast_clang_InheritableAttr(void* p) {
    return llvm::dyn_cast_or_null<clang::InheritableAttr>(reinterpret_cast<clang::AttributeCommonInfo*>(p));
}
```

`llvm::dyn_cast<Derived>` lowers to `Derived::classof(const Root*)` — for `InheritableAttr` the Root is `clang::Attr`. But the `base` here is `clang::AttributeCommonInfo`, which is a base **of** `Attr` (a mixin above the RTTI root), so `AttributeCommonInfo*` does **not** convert to `Attr*`. The compiler reports exactly:

```
error: cannot initialize a parameter of type 'const Attr *' with an rvalue of type 'const clang::AttributeCommonInfo *'
   64 |   static inline bool doit(const From &Val) { return To::classof(&Val); }
```

Every one of the 597 followed this shape — the `base` was a mixin/utility ancestor OUTSIDE the derived's LLVM-RTTI hierarchy:

| source (base) fed to `classof(const Root*)` | count |
|---|---|
| `clang::AttributeCommonInfo` (base of Root `Attr`) | 462 |
| `llvm::FoldingSetBase::Node` (intrusive list mixin) | 68 |
| `clang::Redeclarable<T>` / `clang::Mergeable<T>` (CRTP bases of `Decl`) | ~45 |
| `clang::KeywordHelpers`, `APIntStorage`, `RefCountedBase<T>`, `ObjCProtocolQualifiers<T>`, … | rest |

`downCastTargetsFor` reached these because it inverts `castTargetsFor`, whose BFS walks **every** public base — including the mixins — and it took the LLVM path whenever the *derived* had a `classof`, without checking the *base* was a legal operand for that `classof`.

## The fix (`codegen/CastTargets.kt`)

Take the LLVM `classof` down-cast path only when `llvm::dyn_cast<Derived>(Base*)` is actually well-formed. `dynCastFromBaseIsWellFormed(base, derived)` admits exactly the two operand shapes LLVM's cast machinery accepts:

1. **`base` is (or derives from) a `classof` Root of `derived`** — read straight off the bound `classof` signature's first parameter. This keeps `dyn_cast<InheritableAttr>(Attr*)` working even though `Attr` itself declares no `classof` (it IS the Root).
2. **`base` declares its own `classof`** — an LLVM-RTTI node with its own cross-hierarchy bridge, e.g. `Decl`↔`DeclContext` via `castFromDeclContext`. This keeps `dyn_cast<TagDecl>(DeclContext*)` working even though `DeclContext` is a parallel base, not on the `Decl` Root chain.

A mixin ancestor satisfies neither and is dropped (no safe checked down-cast from an off-hierarchy ancestor exists). The generic-`dynamic_cast` path is untouched.

## Broad before/after (harness: `docs/campaigns/broad-error-measure.sh`)

| bucket | before | after |
|---|---:|---:|
| **TOTAL** | **1531** | **935** |
| **rvalue-param-init** | **597** | **1** |
| no-matching-function | 556 | 556 |
| no-matching-constructor | 73 | 73 |
| deleted-ctor | 55 | 55 |

−596 total; the cluster goes 597 → **1**. Every other bucket is untouched (the fix is surgical). The 1 residual is a *distinct, unrelated* cause — a most-vexing-parse (`clang::TemplateArgumentListInfo item();` parsed as a function declaration), not a base-pointer init — so the 597 base-pointer cluster is effectively cleared.

Confirmed on the fixed `.cc`: `Attr_dyncast_*` = 462 kept (valid), `DeclContext_dyncast_*` = 31 kept (valid bridge), `AttributeCommonInfo_dyncast_*` = 0 and `FoldingSetBase_Node_dyncast_*` = 0 (broken dropped). No cast-helper appears in the error output anymore.

## Behavior preservation

**Verified-correct, no regression on existing bindings:**

- **clangwalk 18-class clang-slice binding (IGNORE_MISSING, the shipped stage1 binding — the one that DOES carry LLVM-`classof` types):** regenerated with a clean-main baseline kexe vs this branch → **`.cc` and `.h` byte-identical**. The valid `Attr`/`DeclContext`/`Decl`-rooted casts it emits are unchanged; the fix only removes casts that never compiled.
- **featuregen:** its model carries **zero `classof` methods** (`grep -c '"classof"'` = 0 across base + all forcing models), so `hasLlvmClassof(derived)` is always false and this code path is **provably inert** — featuregen's `dynamic_cast` down-casts are untouched.

An intermediate cut of the fix over-tightened the gate (dropped valid `DeclContext→TagDecl` / degraded `Decl→NamedDecl` to `dynamic_cast`); the byte-identical clangwalk diff is what caught it, and the final two-way `dynCastFromBaseIsWellFormed` rule restores those.

## Gates

- `:krapper_gen:nativeTest` — green.
- Known non-gate: the featuregen `-Pkpp.frontend.featuregen=cpp` path drives the *published* tool, not this branch's kexe, so it can't exercise the fix; inertness is proven by the byte-identical clangwalk diff + the zero-classof featuregen argument instead.

## Repeatable harness

`docs/campaigns/broad-error-measure.sh` runs the broad force → clang++ compile → bucket table (reproduces TOTAL 935 / rvalue 1). Reuse for the next #10 cluster (the 556 no-matching-function bucket is now the largest).

Progresses #10.

### DIFF
--- a/docs/campaigns/broad-error-measure.sh
+++ b/docs/campaigns/broad-error-measure.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+# Broad-surface codegen error measurement harness (issue #10).
+#
+# Forces the FULL transitive Clang slice under DefaultFilter + INCLUDE_MISSING (the broadest
+# case krapper_gen can be asked to bind), then compiles the generated C++ with clang++ and
+# buckets the errors. This is the "measure" half of the #10 measure-fix-remeasure loop that
+# PR #162 first made reachable (base-bind collapsed 250s -> 4.3s).
+#
+# Prerequisites (run once, on an LLVM box):
+#   export JAVA_HOME=/usr/lib/jvm/java-21-openjdk
+#   export GRADLE_OPTS="-Dorg.gradle.jvmargs=-Xmx12g -XX:MaxMetaspaceSize=512m"
+#   ./gradlew :krapper_gen:linkReleaseExecutableNative -PenableClang
+#   ./gradlew :clangwalk:kplusplusSync -PenableClang   # produces the clang-slice base_model
+#
+# Usage:  docs/campaigns/broad-error-measure.sh
+# Output: prints the total error count + a per-family bucket table.
+set -euo pipefail
+
+REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+KEXE="$REPO/krapper_gen/build/bin/native/releaseExecutable/krapper_gen.kexe"
+CM="$REPO/clangwalk/build/krapped-cpp/cpp-models"
+HDR="$REPO/krapper_parse/include/clang_slice.h"
+OUT="${1:-/tmp/kpp10-broad-measure}"
+
+[ -x "$KEXE" ] || { echo "missing $KEXE — build it first (see header)"; exit 1; }
+[ -f "$CM/base_model.json" ] || { echo "missing $CM/base_model.json — run :clangwalk:kplusplusSync"; exit 1; }
+
+rm -rf "$OUT"; mkdir -p "$OUT"
+
+# DefaultFilter (no --only) + INCLUDE_MISSING = the broad transitive surface. The generator
+# runs its OWN final compile and exits 134 on the remaining errors, but the .cc/.h are fully
+# written before that, which is all we need. So the non-zero exit here is EXPECTED.
+"$KEXE" \
+  --header "$HDR" \
+  --parsedModel "$CM/base_model.json" \
+  --referencePolicy INCLUDE_MISSING \
+  --compiler clang++ --std c++17 -o "$OUT" \
+  --instantiate "std::vector<clang::Decl*>" \
+  --instantiate "std::vector<clang::CXXMethodDecl*>" \
+  --instantiate "std::vector<clang::CXXBaseSpecifier*>" \
+  --forcingModel "std::vector<clang::Decl*>=$CM/KrapperForce_std_vector_clang_DeclPtr.json" \
+  --forcingModel "std::vector<clang::CXXMethodDecl*>=$CM/KrapperForce_std_vector_clang_CXXMethodDeclPtr.json" \
+  --forcingModel "std::vector<clang::CXXBaseSpecifier*>=$CM/KrapperForce_std_vector_clang_CXXBaseSpecifierPtr.json" \
+  > "$OUT/gen.log" 2>&1 || true
+
+[ -f "$OUT/clang_slice.h.cc" ] || { echo "generation produced no .cc — see $OUT/gen.log"; exit 1; }
+
+# Standalone compile with all errors surfaced (-ferror-limit=0).
+clang++ -std=c++17 -fsyntax-only -ferror-limit=0 \
+  -I"$OUT" -I"$REPO/krapper_parse/include" -I/usr/include \
+  "$OUT/clang_slice.h.cc" 2> "$OUT/errors.txt" || true
+
+E="$OUT/errors.txt"
+count() { grep -c "$1" "$E" || true; }
+echo "=== broad-surface C++ error buckets ($OUT/errors.txt) ==="
+printf '%-42s %s\n' "TOTAL"                     "$(count 'error:')"
+printf '%-42s %s\n' "rvalue-param-init"         "$(count "cannot initialize a parameter of type .* with an rvalue")"
+printf '%-42s %s\n' "no-matching-function"      "$(count 'no matching function for call')"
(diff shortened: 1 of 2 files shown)


monkopedia-reviewer (approved):
Reviewed to merge standard. Approving.

## Correctness — two-way gate analysis

**No over-drop (false negatives):** The byte-identical clangwalk check passes — both `.cc` and `.h` (and all three KrapperForce headers) are byte-identical between `main` and the branch, freshly regenerated from both kexes against the same model. This is the ground-truth proof: the 18-class clang-slice binding that carries real LLVM `classof` types (Decl→NamedDecl, DeclContext→TagDecl, Attr→InheritableAttr chain) is unchanged.

Traced all documented cases through `dynCastFromBaseIsWellFormed`:
- `dyn_cast<InheritableAttr>(Attr*)`: Attr has no classof of its own → arm 1 hits `base.type.toString() in roots {"clang::Attr"}` → **true** ✓
- `dyn_cast<TagDecl>(DeclContext*)`: DeclContext has own classof → arm 2 `hasLlvmClassof(base)` → **true** ✓  
- `dyn_cast<InheritableAttr>(AttributeCommonInfo*)`: no classof, not in roots, not a transitive base OF Attr (it's a base OF the Root, above it in hierarchy) → **false** → dropped ✓

**No under-drop (false positives):** The 597 rvalue-param-init cluster is cleared. The 1 residual is a distinct most-vexing-parse (cited in PR, not a base-pointer init). Every other error bucket is numerically unchanged.

**Dropped-cast fallback is sound:** when `dynCastFromBaseIsWellFormed` returns false for a derived with classof, `useLlvmDynCast = false`. Two subsequent guards then apply: (a) `noRtti && !useLlvmDynCast → continue` (drops the helper entirely for -fno-rtti libraries, correct — no safe cast exists), (b) `!useLlvmDynCast && !basePolymorphic → continue` (drops for non-polymorphic base without LLVM RTTI, correct — no runtime check mechanism exists). If the base IS polymorphic and not noRtti, we emit `dynamic_cast<D*>(base*)` — valid C++ for any polymorphic base regardless of hierarchy relationship (returns null on failure). The mixin bases that were causing the 597 errors (FoldingSetBase::Node, Redeclarable<T>, AttributeCommonInfo) are non-polymorphic, so arm (b) drops them cleanly.

**`classof`-root detection:** `classofRootTypeStrings` reads the bound `classof` parameter's typeString (`"const clang::Attr *"`) and strips const/pointer/whitespace via `stripToTypeName()` → `"clang::Attr"`. Compared to `ResolvedClass.type.toString()` which is also the bare C++ type string. The strip logic is correct for all real LLVM classof param shapes (class pointer params); the multi-word-type edge case (e.g. `const unsigned char *`) would mangle but never arises in practice for LLVM-RTTI roots (all are named class pointers). Null safety: the `?.` chain in `mapNotNullTo` filters absent-arg or unresolved cases.

**Edge: derived has classof but arg is unbound under IGNORE_MISSING:** if classof's arg fails to resolve, the method is dropped in resolution and `hasLlvmClassof` returns false → fallback to dynamic_cast or drop. The byte-identical clangwalk proof confirms this doesn't cause regression on the shipped surface.

## Gates

- `:krapper_gen:nativeTest` — **green**
- `:krapper_gen:ktlintCheck` — **green**  
- No remote CI workflow on PR branches (only `release.yml` on tag)
- featuregen cpp path: pre-existing exit-134 (published tool, not this kexe), confirmed inert (zero classof in featuregen model)

## Quality

Code is idiomatic Kotlin, well-commented, consistent with the surrounding file. The `dynCastFromBaseIsWellFormed` function is appropriately documented with the two-arm rationale and concrete examples. `stripToTypeName` is a private extension on String — correct scope. `classofMethods` extraction from `hasLlvmClassof` is a sound DRY refactor (single call site is the internal helpers, not a callsite explosion). The harness script is clean, safe, self-documented, and reusable for the next #10 cluster.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #164, opened 2026-07-15
fix(codegen): global-scope ::new for placement-new into caller buffer (broad-surface no-matching-function cluster, #10)

monkopedia-coder:
## The error pattern

The broad transitive Clang surface (`DefaultFilter + INCLUDE_MISSING`) was producing **556 "no matching function for call"** errors — of which **541 were `operator new`**:

```
/tmp/.../clang_slice.h.cc:45812:12: error: no matching function for call to 'operator new'
 45812 |     return new (location) clang::WarnUnusedResultAttr(*Ctx_cast, *CommonInfo_cast, Message_cast);
       |            ^   ~~~~~~~~~~
/usr/include/clang/AST/Attr.h:75:9: note: candidate function not viable: cannot convert argument of
     incomplete type 'void *' to 'ASTContext &' for 2nd argument
   75 |   void *operator new(size_t Bytes, ASTContext &C, ...
```

Same shape via `DeclBase.h` for `clang::Decl` subclasses (its allocator `operator new(size_t, const ASTContext&, GlobalDeclID, size_t)`).

Breakdown of the 556 by callee: **541 `operator new`**, 8 `_Construct`, 4 `__fill_a1`, 2 `swap`, 1 `kpp_to_elem_ptr`.

## Root cause

The constructor wrapper emits placement-new into a caller-provided buffer:

```cpp
return new (location) clang::WarnUnusedResultAttr(...);
```

`clang::Attr`/`clang::Decl` subclasses declare a **class-scoped `operator new`** (the ASTContext allocator new). A class-scoped `operator new` **hides** the standard placement `::operator new(size_t, void*)`, so name lookup for the unqualified `new (location) T(...)` finds only the allocator overload — which doesn't match `(size_t, void*)` — and fails. This is *one* systematic codegen pattern, not several.

## The fix

Emit **global-scope** placement new — `::new (location) T(...)` — at the two placement-into-buffer sites (constructor `DIRECT` allocation + `ARG_CAST` return-in-place). Both flow through the `New` symbol, so the fix is a one-line change in `New.build()`: when a `location` is present, prefix `::`. The leading `::` bypasses class-scope lookup and resolves to the standard `::operator new(size_t, void*)`. For any type that does *not* declare its own `operator new`, `::new (buf)` is **identical** to unqualified placement-new, so this is universally safe for construct-into-provided-buffer.

Files: `krapper_gen/.../builders/BasicSymbols.kt` (the fix) + `krapper_gen/src/nativeTest/kotlin/CppCodeTests.kt` (31 placement-new expectations updated to `::new`).

## Broad before/after (harness `docs/campaigns/broad-error-measure.sh`)

| bucket | before | after |
|---|---|---|
| **TOTAL** | **935** | **394** |
| **no-matching-function** | **556** | **15** |
| rvalue-param-init | 1 | 1 |
| no-matching-constructor | 73 | 73 |
| deleted-ctor | 55 | 55 |

**−541 total.** Surgical: every other bucket is unchanged. The entire 541-error `operator new` sub-cluster is eliminated; the remaining 15 no-matching-function are a pre-existing tail (8 `_Construct`, 4 `__fill_a1`, 2 `swap`, 1 `kpp_to_elem_ptr` — std internal helpers) left as the next target.

## Behavior preservation

Regenerated the clangwalk clang-slice binding (`--frontend=cpp`) on clean `main` vs this branch:
- `clangwalk.h`, all `src/*.kt`, `.def`: **byte-identical**.
- `clangwalk.cc`: the ONLY diff is `new (X)` → `::new (X)` on all 164 placement sites; each `<` line maps to its `>` line by adding exactly `::` (verified mechanically — no other change). krapper_gen's own final compile of the regenerated `.cc` succeeds, so the existing binding still builds cleanly.

This is the "verified correctness improvement" case: additive `::` qualification, semantically identical on the existing surface (clang `QualType` etc. don't hide placement new). Gates: `:krapper_gen:nativeTest` green, `:krapper_gen:ktlintCheck` green, `:featuregen:nativeTest -Pkpp.frontend.featuregen=cpp` green.

Progresses #10

### DIFF
--- a/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/builders/BasicSymbols.kt
+++ b/krapper_gen/src/nativeMain/kotlin/com/monkopedia/krapper/generator/builders/BasicSymbols.kt
@@ -104,11 +104,20 @@ class New(private val s: Symbol, private val location: Symbol? = null) :
         get() = listOf(s)
 
     override fun build(builder: CodeStringBuilder) {
-        builder.append("new ")
         if (location != null) {
-            builder.append('(')
+            // Global-scope placement new: construct into the caller-provided buffer
+            // via ::operator new(size_t, void*). The leading `::` is load-bearing —
+            // classes that declare a class-scoped `operator new` (e.g. clang's
+            // Attr/Decl allocator news taking an ASTContext) HIDE the standard
+            // placement new, so an unqualified `new (buf) T(...)` would resolve to
+            // the allocator overload and fail to match. `::new` bypasses class-scope
+            // lookup and is identical to unqualified placement-new for every type
+            // that does not declare its own operator new.
+            builder.append("::new (")
             location.build(builder)
             builder.append(") ")
+        } else {
+            builder.append("new ")
         }
         s.build(builder)
     }
--- a/krapper_gen/src/nativeTest/kotlin/CppCodeTests.kt
+++ b/krapper_gen/src/nativeTest/kotlin/CppCodeTests.kt
@@ -50,7 +50,7 @@ class CppCodeTests {
     """.trimIndent()
     private val stdVectorStringNew = """
         void* std_vector_std_string_new(void* location) {
-            return new (location) std::vector<std::string>();
+            return ::new (location) std::vector<std::string>();
         }
     """.trimIndent()
     private val stdVectorStringDispose = """
@@ -68,7 +68,7 @@ class CppCodeTests {
     """.trimIndent()
     private val testlibOtherclassNew = """
         void* TestLib_OtherClass_new(void* location) {
-            return new (location) TestLib::OtherClass();
+            return ::new (location) TestLib::OtherClass();
         }
     """.trimIndent()
     private val testlibOtherclassDispose = """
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
## Review verdict: approve

**Semantic correctness of `::new`:** The reasoning is sound and complete. Global-scope `::new (buf) T(...)` bypasses class-scope name lookup and resolves directly to the standard `::operator new(size_t, void*)`. For any type without a class-scoped `operator new`, `::new (buf)` is semantically identical to unqualified placement-new — same resolution, same result. For types like `clang::Attr` / `clang::Decl` that DO declare a class-scoped `operator new` taking an `ASTContext&`, the class-scoped overload hides the standard placement new from unqualified lookup; `::new` bypasses that hiding and correctly selects `::operator new(size_t, void*)`. Global placement new is implicitly declared by the C++ standard (no `<new>` include needed for the built-in) and is always available. There is no case where these MemScope/caller-buffer constructions would legitimately want a class-scoped placement `operator new` — we always want to construct into a provided `void*` buffer, which is exactly the standard placement new's contract. The fix is universally correct and safe.

**Gated on `location != null` — ordinary heap `new` untouched:** The `New.build()` change is cleanly gated on `if (location != null)`. When `location` is absent the `else` branch emits plain `new `. Verified in the regenerated output: 540 `new char[]` array allocations in `clang_slice.h.cc` are byte-identical between main and branch. Ordinary heap `new` is untouched.

**Clangwalk diff is EXACTLY the additive `::`:** Regenerated `clang_slice.h.cc` with the main kexe vs the branch kexe against identical cpp-models. The diff has exactly 4686 line-pair changes; every `<` line contains `new (` and every `>` line contains `::new (` — verified mechanically (grep-checked for any non-conforming line: zero). The `.h` is byte-identical. The `.def` differs only in absolute output-directory paths (expected). No other change in the generated output.

**Generated .cc compiles:** Both main's and branch's generated `clang_slice.h.cc` have the same pre-existing compilation errors (Align::MaybeAlign codegen, clang-c naming conflict). These are tracked pre-existing issues unrelated to this PR; the `::new` change introduces no new errors.

**Broad-error-measure confirmed:** TOTAL 394, no-matching-function 15 — exactly matches the PR's table (−541 from the operator new sub-cluster).

**31 CppCodeTests.kt updates:** All 31 are `new (...)` → `::new (...)` on placement sites. No unqualified placement-new string literal remains. Spot-checked constructor sites (testlibTestclassNew 1–4, stdVectorStringNew), operator-return-in-place sites (op+, op-, op==, op++, op[], etc.) — every update matches the codegen change. CppCodeTests: 83/0, total nativeTest: 300/0. ktlintCheck: green.

**featuregen:** exit-134 on the published-tool env (pre-existing known env issue per review instructions, not a regression — the coder reports 195/0 green with the in-tree tool).

**No remote CI configured** (only a release.yml on:release — no PR check workflow). Local gates are the full verification bar for this repo.

Issue #124, opened 2026-07-08
cppfrontend cpp-parse relies on system-default Clang header path (fragile cross-machine)


monkopedia-coder:
Surfaced by #122 step 1.1. When cppfrontend parses its own `include/clang_slice.h` via the =cpp path, it found `<clang/AST/...>` etc. only because this box has system LLVM 22 with its headers on the default include search path (resource-dir /usr/lib/clang/22). `:cppfrontend` declares no `headerDirectory`, so on a machine where the LLVM headers are NOT on the default path (e.g. apt.llvm.org under /usr/lib/llvm-22/include — cf. the CI CPATH workaround in the portability probe), the parse would fail to find them.

**Proposed:** thread the LLVM include dir (`llvm-config --includedir`) into the cpp-parse invocation as an explicit `-I` / headerDirectory, so cppfrontend's self-parse (and any consumer parsing against a non-default-path LLVM) is host-portable. Related to the distribution work in #122 (per-major builds on varied hosts).

monkopedia-coder:
Triage: labeling `blocked`. The include-path fix here is explicitly folded into the first-shippable-release roadmap as brick **R4 of #128** ("fold in #124 — explicit cpp-parse include path"). Dispatching it independently would risk duplicating/conflicting with that brick's work, so tracking it as blocked-on-#128 rather than agent-workable. If you'd rather land it standalone ahead of #128, relabel it `agent-workable` and it'll dispatch next run.

monkopedia-coder:
Triage update: the blocker here is gone. The include-path fix landed in PR #170 ("resolve the consumer's LLVM include path for cpp-parse (#124, #128 R4)", merged 2026-07-19), and #128 — which this was tracked as blocked-on — is now closed. Removing the `blocked` label. This looks resolved by #170; recommend closing as fixed. (Triage classifies rather than closes, so leaving the final close to the owner.)

monkopedia-coder:
Closing — resolved by #170 (merged), which threads the consumer's LLVM include path into the cpp-parse via `llvm-config --includedir` auto-discovery + the `kplusplus { llvmConfig }` DSL override (additive-only, keeps the in-tree path byte-identical). The system-default-Clang-header-path fragility this issue described is fixed. Its parent #128 also closed (first-shippable-release goal met).

PR #170, opened 2026-07-19
feat(plugin): resolve the consumer's LLVM include path for cpp-parse (#124, #128 R4)

monkopedia-coder:
## The gap (#124 / #128 R4)

`krapper_parse`'s `--parity-emit` builds its Clang parse args as `-std=X`, `-resource-dir=$(clang++ -print-resource-dir)`, plus the `-I` flags the plugin passes from `kplusplus { headerDirectory(...) }`. So:

- **Clang's own builtin headers** (`stddef.h` etc.) — found via `clang++ -print-resource-dir` on `PATH`. **Robust** (already handled).
- **A library's own headers** under a **non-default-path** LLVM install (e.g. apt.llvm.org's `/usr/lib/llvm-22/include`, where `<clang/AST/...>` lives) — found **only if explicitly on the include path**. In-tree this box happens to have system LLVM-22 on the default path, so no `-I` was ever needed; `:krapper_parse` (ex-`:cppfrontend`) declares no `headerDirectory`. A from-published consumer on a versioned install has neither the `settings.gradle.kts` toolchain probe nor that default path → the parse silently drops the include.

So the gap is category (b), the consumer's/self target headers — **not** the builtin headers.

### "blocked" verdict

The `blocked` label was **pure coordination**, not a technical blocker: triage folded #124 into #128 R4 to avoid a duplicate/conflicting dispatch. R4 is exactly this work, so it is unblocked now. No Jason-level decision required; it was "unbuilt", not externally blocked.

## Design

- **Auto-discover** the LLVM include dir via `llvm-config --includedir` on `PATH` (mirrors the root `settings.gradle.kts` probe) and thread it into `krapper_parse` as a `-I`, reusing the existing `includeFlags` partition (no parallel path).
- **Override chain**: new `kplusplus { llvmConfig = "..." }` DSL knob → the existing `-PllvmConfig=<path>` project property → `llvm-config` on `PATH`.
- **Actionable failure**: a configured `llvm-config` that isn't executable, or that reports a non-directory `--includedir`, fails the build immediately with a message telling the consumer exactly what to set — instead of a cryptic `file not found` deep in the parse.
- **Additive-only guard**: when `--includedir` reports a default system path (`/usr/include`, `/usr/local/include`), nothing is added — Clang already searches those, and injecting them as an explicit `-I` would only reorder search precedence. This keeps the **in-tree system-LLVM cpp path byte-for-byte unchanged** (featuregen/cppfixture declare no `headerDirectory` and rely on the default path).

## What I verified

- `:kplusplus-compiler-gradle:compileKotlin` green; module has no ktlint task / no unit tests (NO-SOURCE).
- **The #124 gap reproduced + fixed through the actual published 0.3.3 `krapper_parse` binary.** A consumer header `#include <mylib/widget.h>` where `widget.h` lives only under a non-default dir:
  - **without `-I`** (repros #124): `fatal error: 'mylib/widget.h' file not found`, and the emitted base model is truncated (captures only the outer `Gadget`, drops `mylib::Widget`).
  - **with `-I<non-default>`** (what the plugin now threads): clean parse, model captures **both** `Gadget` and `Widget`.
- **LLVM-not-on-default-path case**: a fake versioned `llvm-config` reporting `<dir>/include` → the resolver threads `-I<dir>/include`. The real system `llvm-config --includedir` here is `/usr/include` → resolver returns **empty** (guarded) → in-tree path unaltered.
- In-tree additivity confirmed structurally: featuregen/cppfixture declare no `headerDirectory`, and the guard drops the `/usr/include` result, so their `includeFlags` are identical to before.

## Docs

`docs/getting-started.md` — the real external-consumer flow (published `0.3.3` plugin from `mavenCentral()`/Plugin Portal, LLVM-22 prerequisite, the `llvmConfig`/`-PllvmConfig` override, minimal `kplusplus { }` config, build + run), plus a README quickstart link and the `llvmConfig` knob in the DSL snippet.

## Files touched

- `compiler/gradle/.../KPlusPlusCompilerGradlePlugin.kt` — `resolveLlvmIncludeDirs` + `probeExec`/`resolveExecutable` helpers; thread the discovered dir into `runKrapperParseSync`'s `includeFlags`; up-to-date input for `llvmConfig`.
- `compiler/gradle/.../KPlusPlusExtension.kt` — `var llvmConfig`.
- `docs/getting-started.md` (new), `README.md`.

Refs #124, #128.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ kplusplus {
 
     cppStandard = "c++17"             // default c++14
 
+    // Point the front-end at a non-default-path LLVM-22 install, if your `llvm-config`
+    // isn't on PATH (auto-discovered otherwise; also settable via -PllvmConfig).
+    // llvmConfig = "/usr/lib/llvm-22/bin/llvm-config"
+
     // Force template specializations you use (C++ instantiates templates at
     // compile time, so each concrete type you call needs to be requested).
     instantiate("std::vector<int>")
@@ -38,6 +42,12 @@ kplusplus {
 ```
 
 Build, and the generated Kotlin bindings are on the classpath — call your C++ API directly.
+
+New to K++, or consuming the **published** plugin from an external project? Start with the
+step-by-step [docs/getting-started.md](docs/getting-started.md) — the full external-consumer flow
+(declare the plugin from `mavenCentral()`/the Plugin Portal, the LLVM-22 prerequisite + how to
+point K++ at a non-default-path install via `llvmConfig` / `-PllvmConfig`, minimal config, build +
+run).
 
 Runnable samples live under [`samples/`](samples). Start with
 [`samples/minimal`](samples/minimal) — a tiny hand-written C++ geometry library bound and run
--- a/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
+++ b/compiler/gradle/src/main/kotlin/com/monkopedia/kplusplus/compiler/gradle/KPlusPlusCompilerGradlePlugin.kt
@@ -507,14 +507,6 @@ class KPlusPlusCompilerGradlePlugin : KotlinCompilerPluginSupportPlugin {
     }
 
     /**
-     * Pick a C++ compiler for krapper_gen's wrapper-library step. krapper_gen
-     * generates C++ code that references some libstdc++ internals (matching
-     * what v8's own headers expose); newer host gcc / clang reject those.
-     * The konan-bundled toolchain (gcc 8.3 / glibcxx 7) accepts them. Try
-     * the konan toolchain first if present; otherwise fall back to clang++.
-     * Override via `kplusplus { compiler = "..." }`.
-     */
-    /**
      * Resolve the consumer's LLVM/Clang include dir(s) to thread into the cpp-parse as `-I`
      * (#124, #128 R4). The cpp front-end's bundled Clang finds Clang's OWN builtin headers via
      * its resource-dir (`clang++ -print-resource-dir`, resolved inside krapper_parse) and system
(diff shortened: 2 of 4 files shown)


monkopedia-reviewer (changes requested):
One clear quality issue before merge — correctness and logic are sound, so this should be a quick fix.

**Orphaned KDoc for `resolveDefaultCompiler` (compiler/gradle/.../KPlusPlusCompilerGradlePlugin.kt, lines 509–516)**

The new `resolveLlvmIncludeDirs`, `probeExec`, and `resolveExecutable` functions were inserted immediately after `resolveDefaultCompiler`'s existing KDoc block, pushing `resolveDefaultCompiler()` to line 605. The KDoc now floats 95 lines above its function, documenting nothing — and `resolveDefaultCompiler` itself is left with no doc. Fix: move the KDoc block to sit immediately above `private fun resolveDefaultCompiler()` at line 605.

---

Everything else is sound:

- **Additive-only guard verified**: `llvm-config --includedir` on this box returns `/usr/include`, which is in `DEFAULT_SYSTEM_INCLUDE_DIRS` → `resolveLlvmIncludeDirs` returns `emptyList()` → `llvmIncludeDirs` is empty → `includeFlags` is identical to pre-PR. Featuregen and cppfixture declare no `headerDirectory` → their `includeFlags` were already empty, and remain empty.
- **#124 fix is sound**: override chain (DSL → `-P` property → PATH) is correctly implemented; configured-but-unusable `llvm-config` throws a `GradleException` with an actionable message rather than a silent skip; `resolveExecutable` mirrors `settings.gradle.kts` faithfully.
- **Up-to-date input**: `llvmConfig` declared as `inputs.property` → Gradle caching is correct.
- **`resolveLlvmIncludeDirs`/`probeExec`/`resolveExecutable`**: logic correct, idiomatic, consistent with surrounding code.
- **Docs**: `docs/getting-started.md` + README accurately describe the consumer flow, LLVM prerequisite, and override chain. Well-written.
- **`compileKotlin` gate**: green (via `samples/v8` composite build).

monkopedia-coder:
Relocated the orphaned `resolveDefaultCompiler` KDoc to immediately above its function per review nit. The three new private functions inserted in this PR had displaced the block ~90 lines upward, leaving `resolveDefaultCompiler()` undocumented. Pure doc-comment relocation — no logic change. `:kplusplus-compiler-gradle:compileKotlin` green on the updated branch.

monkopedia-reviewer (approved):
Addressing commit 09ffead verified: KDoc-only relocation, no logic change. The 8-line resolveDefaultCompiler KDoc block was removed from its stray position (~line 506) and re-added immediately above private fun resolveDefaultCompiler() — diff shows exactly one removed block and one added block of identical text, single file only. :kplusplus-compiler-gradle:compileKotlin green. All prior substantive findings (additive-only guard, #124 fix correctness, override chain, docs accuracy) remain as approved. Merging.

PR #172, opened 2026-07-19
fix(codegen): spell fully-defaulted template specializations with their canonical args (broad-surface CTAD cluster, #10)

monkopedia-coder:
## Cluster chosen & why
The broad-surface tail's largest **systematic** cluster: **use of a class template without its arguments (CTAD not allowed here)** — 25 errors that manifest in three grep buckets that are the SAME underlying failure:
- `requires template arguments; argument deduction not allowed here` (25)
- `cannot form pointer to deduced class template specialization type` (25)

Split by type: **13 `llvm::BumpPtrAllocatorImpl` + 12 `std::chrono::time_point`** — one shared root cause. (The other big buckets — `deleted-ctor` 15 default + 16 copy + 24 explicit-deleted, `invalid-binary-operands` 35 — are each several distinct bespoke roots, not one systematic pattern.)

## Error pattern
```
error: use of class template 'std::chrono::time_point' requires template arguments; argument deduction not allowed here
    std::chrono::time_point* ret_value_cast = reinterpret_cast<std::chrono::time_point*>(ret_value);
    ::new (ret_value_cast) std::chrono::time_point(thiz_cast->getLastAccessedTime());
```
The type is spelled as the **bare template name** — no `<...>`.

## Root cause
`TimePoint<>` (alias for `time_point<system_clock, D>`, `D` defaulted) and `BumpPtrAllocator = BumpPtrAllocatorImpl<>` (all params defaulted) are written with an **empty** argument list, so `numTemplateArgs(type)` reads the sugared TemplateSpecializationType's ZERO *written* args. `TypeBuilder`'s template branch is guarded on that count (`> 0`), so it never fires; the type falls through to the **record leaf**, which spells `qualifiedName(record)` = the naked template name. C++ rejects a naked template-name wherever the binding names the type by value (a field getter's `reinterpret_cast<T*>`, a return placement-`new T(...)`).

## Fix
`krapper_parse/.../TypeBuilder.kt`, record-leaf branch: when the canonical record IS a `ClassTemplateSpecializationDecl` carrying args (`numTemplateArgs(canonical) > 0`), recover the FULL resolved argument list and emit a real `WrappedTemplateType` instead of the bare ref. Only fully-defaulted specializations reach this leaf (any with ≥1 written arg took the sugared branch), so the `vector<Item*>` written-arg preference is untouched.

**Guard:** a specialization carrying a bare `void` TYPE arg (`std::less<void>` & the other transparent comparators) is NOT expanded — its members then take dependent params that resolve to `void` BY VALUE (`operator()(const void, const void)`, ill-formed), so we keep the prior inert bare-name leaf. `time_point`/`BumpPtrAllocatorImpl` carry no `void` arg and expand.

## Broad before/after (`docs/campaigns/broad-error-measure.sh`, clean-main-fresh → branch)
| bucket | before | after |
|---|---|---|
| **TOTAL** | **295** | **245 (-50)** |
| requires-template-args / deduction-not-allowed | 25 | 0 |
| cannot-form-pointer-to-deduced | 25 | 0 |

Zero new error signatures; **no other bucket regressed** (verified by `comm`/`diff` of per-signature counts vs a freshly-parsed clean-main baseline — the only deltas are the three removals above).

## Behavior-preservation
The clangwalk self-hosted slice (`--only` allowlist + IGNORE_MISSING) regenerates **BYTE-IDENTICAL**: same-env clean-main-exe vs branch-exe over freshly re-parsed `base_model.json`, `.cc` + `.h` + `src/*.kt` + all 3 forcing headers unchanged (the expanded types live only in the broad transitive surface, never in the scoped slice). `:krapper_gen:nativeTest`, `:krapper_gen:ktlintCheck`, `:krapper_parse` source ktlint, and `:featuregen:nativeTest -Pkpp.frontend.featuregen=cpp --rerun-tasks` all green.

## Remaining next targets
The tail is now increasingly bespoke: `deleted-ctor` (synthesized default `_new` for classes whose implicit default ctor is deleted — diverse per-type reasons, clang exposes no single `defaultedDefaultConstructorIsDeleted` bit in this version; the copy/explicit-deleted halves are move-only-by-value passes) and `invalid-binary-operands` (35 — STL `==` on element types lacking `operator==`, each a distinct type).

Progresses #10

### DIFF
--- a/krapper_parse/src/nativeMain/kotlin/TypeBuilder.kt
+++ b/krapper_parse/src/nativeMain/kotlin/TypeBuilder.kt
@@ -281,6 +281,39 @@ fun buildWrappedType(type: QualType): WrappedType {
             ?.takeIf { it.isNotEmpty() } ?: return UNRESOLVABLE
         // #123: an inaccessible nested record can't be named from the wrapper — drop it.
         if (!record.asDecl().isAccessibleTagType()) return UNRESOLVABLE
+        // FULLY-DEFAULTED template specialization (#10): a use written with an EMPTY
+        // argument list (`BumpPtrAllocatorImpl<>`, or `TimePoint<>` = the alias
+        // `time_point<system_clock, D>` whose sole non-defaulted arg is itself defaulted)
+        // carries ZERO *written* args, so the sugared-TST branch above (guarded on the
+        // written count) never fires and the type reaches this record leaf. Spelling it as
+        // the bare template NAME (`llvm::BumpPtrAllocatorImpl`) is not a valid type — C++
+        // rejects the naked template-name ("requires template arguments; argument deduction
+        // not allowed here") wherever the binding names it by value (a field getter's
+        // `reinterpret_cast<T*>`, a return placement-`new T(...)`). The CANONICAL record IS
+        // a ClassTemplateSpecializationDecl carrying the FULL, resolved argument list, so
+        // recover it here and emit a real `WrappedTemplateType` (`BumpPtrAllocatorImpl<...>`).
+        // Only fully-defaulted specializations reach this point — any with ≥1 written arg
+        // took the sugared branch — so the `vector<Item*>` written-arg preference is intact.
+        val canonArgCount = numTemplateArgs(canonical)
+        if (canonArgCount > 0) {
+            val args = (0u until canonArgCount.toUInt()).map { i ->
+                val arg = templateArgAsType(canonical, i)
+                if (arg.typePtr() == null) UNRESOLVABLE else buildWrappedType(arg)
+            }
+            // A `void` TYPE argument marks a TRANSPARENT / incomplete-friendly specialization
+            // (`std::less<void>` & the other `<void>` transparent comparators): the member
+            // signatures then take dependent params that resolve to a bare `void` BY VALUE
+            // (`operator()(const void, const void)`), which is ill-formed C++ ("'void' must be
+            // the first and only parameter"). Such a type never legitimately binds by value, so
+            // KEEP the prior bare-name leaf here — on which the resolver never materialized a
+            // usable binding anyway (the naked name is inert, not named by value) — rather than
+            // expanding it into a specialization whose newly-resolved members emit `void`
+            // value params. `time_point`/`BumpPtrAllocatorImpl` carry no `void` arg and expand.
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (approved):
Reviewed at tier-1 standard. All gates verified independently.

**Byte-identical clangwalk (load-bearing safety check):** Confirmed byte-identical. Method: built main's `krapper_gen` kexe fresh in a `/tmp/review-172-main` worktree, ran both kexes against a FRESHLY regenerated `base_model.json` (produced by the branch's `krapper_parse --parity-emit` for the `--only` allowlist). `.cc` + `.h` + `src/*.kt` diff clean; only `.def` differs by output path (expected). The CTAD types (`BumpPtrAllocatorImpl`, `time_point`) do not appear in the `--only` allowlist and thus never enter the scoped slice — byte-identical is the correct and expected result.

**Fix correctly scoped:** The fix fires ONLY in the record-leaf branch, which is reached only by types with zero WRITTEN template args. Any type with ≥1 written arg exits earlier via the sugared-TST branch (lines ~173–211), untouched. The void-arg guard is correct and matches the PR's reasoning (`std::less<void>`-style transparent comparators emit `void` BY VALUE params — ill-formed C++). Code mirrors the existing sugared-branch pattern (same `numTemplateArgs` / `templateArgAsType` / `buildWrappedType` calls, same `0u until count.toUInt()` iteration idiom from line 207).

**Broad measure (independently reproduced):**
Built main's `krapper_parse` binary, ran `--parity-emit` to get a true pre-fix model, then compiled with clang++:

| Bucket | BEFORE (main) | AFTER (branch) |
|---|---|---|
| **TOTAL** | **295** | **245 (−50)** |
| requires-template-args | 25 | **0** |
| deduction-not-allowed | 25 | **0** |
| pointer-to-deduced | 25 | **0** |
| rvalue-param-init | 1 | 1 |
| no-matching-constructor | 24 | 24 |
| deleted-ctor | 55 | 55 |

Zero regression in any other bucket.

**Gates:** `:krapper_gen:nativeTest` green, `:krapper_gen:ktlintCheck` green. No CI configured on PRs (release.yml only). No new test expectations needed (the CTAD types are absent from the scoped slice; broad surface is measured directly).

